Issue #339, opened 2015-01-07
The helios-agent can't talk to docker in the vagrant setup


noaresare:
Running 'vagrant up' on my OSX 10.10 machine on a checked out master branch of helios results
in a vagrant box I can ssh into (using 'vagrant ssh' in the helios directory) that has the components set up with master, agent, zookeeper and others but the running helios-agent can't communicate with the docker daemon due to permissions on the docker unix socket (/var/run/docker.sock) which is only writeable and readable by the root user and docker group. The helios user running helios-agent is not a member of the docker group.

noaresare:
It turns out there is a bit of a chicken and egg problem with adding the helios user to the appropriate group as the user is created by the helios-agent deb package postinst script and the daemon is started only a few lines later. 

My suggestion would be to duplicate the helios user creation command into the Vagrantfile and rely on the fact that the postinst checks for existance of an account named helios before attempting to add it, then add the helios user to the docker group after having crated the package (but before installing and starting helios-agent). Thoughts?

davidxia:
Can we close since this is merged or are there other things to do?

noaresare:
Sure, let's close it.

PR #340, opened 2015-01-07
Make sure helios-agent in the vagrant box can connect to docker

noaresare:
My suggested fix for https://github.com/spotify/helios/issues/339

Duplicate adding the helios user from the helios-agent deb package
postinst file so that the helios user can be added to the docker
group before the helios-agent is started.


rculbertson:
:+1:

PR #343, opened 2015-01-09
README update, to cater for new users

noaresare:
* Expanded a bit on the high level project description
* Added a section giving a high level architecture overview
* Added some more details on dependencies
* Renamed debian package -> deb package

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,8 @@ The `helios` command line tool connects to your helios master via HTTP. The
 Helios master is connected to a Zookeeper cluster that is used both as
 persistent storage and as a communications channel to the agents. The
 helios agent is a java process that typically lives on the same host as
-the Docker daemon, connecting to it via a Unix socket.
+the Docker daemon, connecting to it via a Unix socket or optionally TCP
+socket.
 
 Helios is designed for high availability, with execution state being confined
 to a potentially highly available Zookeeper cluster. This means that several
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,6 +37,22 @@ $ helios undeploy -a nginx:v1
...
 ```
 
+How it all fits together
+------------------------
+
+The `helios` command line tool connects to your helios master via HTTP. The
+Helios master is connected to a Zookeeper cluster that is used both as
+persistent storage and as a communications channel to the agents. The
+helios agent is a java process that typically lives on the same host as
+the Docker daemon, connecting to it via a Unix socket.
```
davidxia: Unix sockets are supported but I think it's still HTTP?

README.md
```diff
@@ -97,12 +117,21 @@ If you're looking for how to download, build, install and run Helios, keep readi
 Prerequisities
 --------------
 
+The binary release of Helios is built for Ubuntu 14.04.1 LTS, but Helios should
```
davidxia: Do we actually specifically build for Ubuntu 14.04 or is it just a generic deb package? Is it built from jdeb plugin specified in pom.xml? I'm not too familiar with this part.

noaresare: Well, deb packaging is not a proper standard, so the fact that you sometimes can install deb packages on a distribution it was not built for is more of a happy coincidence than something you can expect. Because of this I think it's a good idea to be explicit about what target environment we're building for.

README.md
noaresare: Please see updated wording.

README.md
davidxia: sounds good to me

davidxia:
:+1:

davidxia:
@noaresare Thanks for this!

PR #344, opened 2015-01-09
Java 8 build fix ( javadoc errors with the new angry javadoc tool)

noaresare:
The long URL in the TemporaryJobs.java can no longer be split into two lines, and we enforce line lengths, so I'm removing it. If anyone wants to figure out the proper way to do that instead, be my guest :)

### DIFF
--- a/helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
@@ -242,6 +242,9 @@ public class TemporaryJobs implements TestRule {
    * </ol>
    *
    * @return an instance of TemporaryJobs
+   * @see <a href=
+   * "https://github.com/spotify/helios/blob/master/docs/testing_framework.md#configuration-by-file"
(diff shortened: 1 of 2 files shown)


helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
```diff
@@ -242,8 +242,6 @@ public TemporaryJobBuilder job() {
    * </ol>
    *
    * @return an instance of TemporaryJobs
-   * @see <a href="https://github.com/spotify/helios/blob/master/docs/testing_framework.md#
-   * configuration-by-file">Helios Testing Framework - Configuration By File</a>
    */
```
rculbertson: Could we instead change the `@see` tag to this? Formatting looks a little funny, but it passes the javadoc check.

    * @see <a href=
    * "https://github.com/spotify/helios/blob/master/docs/testing_framework.md#configuration-by-file"
    * >Helios Testing Framework - Configuration By File</a>

rculbertson:
:+1:

PR #345, opened 2015-01-21
Add support for binding volumes to all containers

rohansingh:
Add an agent-level argument for binding a volume into all containers. This
is handy if you have a directory on all agents that you want accessible in
every container.

Our use case at Spotify is to mount the host's /etc/ssl/certs into all
containers.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/agent/BindVolumeContainerDecorator.java
+++ b/helios-services/src/main/java/com/spotify/helios/agent/BindVolumeContainerDecorator.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright (c) 2014 Spotify AB.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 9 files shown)


helios-services/src/main/java/com/spotify/helios/agent/BindVolumeContainerDecorator.java
```diff
@@ -0,0 +1,87 @@
...
+      b.addAll(hostConfig.binds());
+    }
+
+    b.addAll(this.binds);
+
+    hostConfig.binds(b);
+  }
+
+  @Override
+  public void decorateContainerConfig(Job job, ImageInfo imageInfo,
```
davidxia: Is this supposed to be empty?

rohansingh: Yes, since this class implements `ContainerDecorator` but doesn't need to do anything here.

davidxia:
:+1:

PR #346, opened 2015-01-24
Reduce log noise from docker health check

rculbertson:
The healthcheck is printing out a success message about 5 times per minute
which makes the logs tough to read. Let's print a message only if we detect
a problem.


davidxia:
:+1:

PR #347, opened 2015-01-26
Build docker image for master and agent during package phase

rculbertson:


davidxia:
:+1: Does this add a little more time to our builds? Took 11:23min for me to `mvn clean package`

rculbertson:
I tested yesterday and it added maybe 1.5 minutes? We could build the containers as part of a pre-integration test phase. That way they are only build if you're about to run tests which require them. If it would be useful to be able to create the containers as part of the package phase, we could create a profile which does that.

davidxia:
yea that sounds like a good idea

rculbertson:
I added a profile for building the containers.
    mvn -P build-containers package

We can also make it part of the pre-integration test phase once we have some test to run.

PR #349, opened 2015-01-28
Add a STOPPING state for tasks that are in the grace period.

mbruggmann:
Before this patch, a task in the grace period would appear to be RUNNING, causing our continuous deployment scripts to undeploy certain jobs twice.


rculbertson:
:+1:

PR #350, opened 2015-01-28
add a link to Spotify's homebrew tap

mattnworb:
In case anyone gets confused looking for helios in the default homebrew formulas.


davidxia:
:+1:

PR #353, opened 2015-02-03
remove mentions of Hermes from the testing doc

mattnworb:
Initially I was updating the document to make it clear what the
`temporaryJob` variable was but then I noticed the references to hermes,
which might be confusing to non-Spotify readers, so I replaced it with
HttpClient.


rohansingh:
Oops. Nice catch.

PR #354, opened 2015-02-03
HeliosClient should try other hosts if UnknownHostException is thrown

rculbertson:
HeliosClient currently retries if a ConnectException is thrown. This happens
if nothing is listening on the specified port. But we don't retry on
UnknownHostException which happens if we can't resolve the hostname to an IP.
We should retry on that as well since it's a nicer user experience.


rohansingh:
:+1:

PR #357, opened 2015-02-04
Get rid of mentions of & references to Hermes

rohansingh:
Hermes is a proprietary Spotify protocol that is only used internally. It
doesn't make sense to mention it here.


davidxia:
:+1:

PR #360, opened 2015-02-06
Use the 3.5.0-alpha ZooKeeper client

### LINKED ISSUE #359: Helios agent can't start when a ZooKeeper node can't be resolved
The Helios agent can't start if a ZooKeeper node that is specified by hostname can't be resolved to an IP. This can happen if the hostname is wrong or if DNS is down.

The root cause is a bug in the ZooKeeper client itself:
https://issues.apache.org/jira/browse/ZOOKEEPER-1576

rohansingh:
Fixes #359.

Normally I'd be wary of using a version marked "3.5.0-alpha". But given the pace of ZooKeeper releases and also judging from messages from the maintainers on the mailing list, I believe:

1. This version should be stable, especially the client bits.
2. It could literally be months/years until we see the next official "stable" release.

Given that we've already been bitten by #359 in production at Spotify, I'd rather just go ahead with this. Especially since all the tests pass and we're not upgrading any ZooKeeper servers, just the client.

### DIFF
--- a/helios-services/pom.xml
+++ b/helios-services/pom.xml
@@ -135,7 +135,7 @@
     <dependency>
       <groupId>org.apache.curator</groupId>
       <artifactId>curator-recipes</artifactId>
-      <version>2.5.0</version>
+      <version>2.7.1</version>
       <exclusions>
         <exclusion>
           <groupId>org.slf4j</groupId>
--- a/helios-services/pom.xml
+++ b/helios-services/pom.xml
@@ -133,6 +133,21 @@
       <version>17.0</version>
     </dependency>
     <dependency>
+      <groupId>org.apache.zookeeper</groupId>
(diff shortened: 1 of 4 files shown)


davidxia:
Wow good find. I took a look at the at your fix and it seems sensible. I didn't look at zookeeper mailing lists and change log though. I'll defer to you on that. It is a bit scary it's an alpha though...

rohansingh:
Just to summarize our in-person discussion for the record: yeah, it's "alpha", but all known issues are pretty well-documented in the ZooKeeper JIRA — and it will probably be several releases before we get through alpha, beta, RC, etc. Here's the proposed schedule:

http://markmail.org/message/ymxliy2rrwjc2pmo

None of the 3.5.0 issues affect our usage of it as a client library.

davidxia:
In that case :+1

PR #364, opened 2015-02-06
[WIP] History Event Listeners

Kobolog:
WIP.

Enable external HTTP endpoint subscriptions in Helios. Such endpoints would be POST-ed with a JSON made from each Job state transition event happening in the cluster, for example RUNNING -> STOPPED.


Kobolog:
@evgeny-goldin

rohansingh:
What's the deal with this? Should it still be an open pull request?

Kobolog:
re @evgeny-goldin, do you think we should still keep this open considering that I've implemented Kafka event delivery thingie?

evgeny-goldin:
No, we shouldn't keep this PR.

PR #366, opened 2015-02-06
Further parallelize the CircleCI build

rohansingh:
Looks like we actually have 3 build containers available. Take advantage of
them by splitting the system tests across two containers.


davidxia:
:+1:

PR #369, opened 2015-02-07
Show better errors for bad HELIOS_MASTER env vars

davidxia:
Instead of just showing NullPointerException,
tell user what's actually wrong.

### DIFF
--- a/helios-tools/src/main/java/com/spotify/helios/cli/CliConfig.java
+++ b/helios-tools/src/main/java/com/spotify/helios/cli/CliConfig.java
@@ -41,6 +41,7 @@ import java.util.List;
 import java.util.Map;
 
 import static com.google.common.base.Preconditions.checkNotNull;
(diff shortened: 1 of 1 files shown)


helios-tools/src/main/java/com/spotify/helios/cli/CliConfig.java
```diff
@@ -157,6 +157,11 @@ public static CliConfig fromEnvVar(Map<String, Object> config) throws URISyntaxE
 
     // TODO (dxia) Remove DEPRECATED_SITE_SCHEME after a period of time
     final String scheme = uri.getScheme();
+    if (scheme == null) {
```
rohansingh: Maybe use `isNullOrEmpty` here? Best practice is to treat null/empty strings the same.

davidxia: good point

rohansingh:
:+1:

PR #374, opened 2015-02-10
Log job state when job deploy message is specified

rculbertson:
We were not logging job state (e.g. PULLING_IMAGE) when jobDeployedMessageFormat
was specified. This meant when image pulls took a long time, there was no indication
what was happening.


davidxia:
:+1:

PR #375, opened 2015-02-12
Add HTTP timeout to client

rculbertson:
If a host was unreachable, it would take the client 2 minutes to timeout. This would
put it past the retry deadline, so it wouldn't try any other hosts. This sets the HTTP
timeout to 10s, and the retry timeout to 60s.


rohansingh:
:+1:

PR #376, opened 2015-02-13
[TempJobs] Made deploy timeout configurable

rculbertson:
Also changed the default from 5 min to 10 min because
we've been seeing image pulls taking ~8 min. This caused
test failures even though they would have passed otherwise.

### DIFF
--- a/helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJob.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJob.java
@@ -22,7 +22,6 @@
 package com.spotify.helios.testing;
 
 import com.google.common.base.Optional;
-import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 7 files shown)


helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJob.java
```diff
@@ -193,21 +196,12 @@ void deploy() {
         }
       }
 
-      // Wait for jobs to have container ids
```
davidxia: Ah so this logic was moved to https://github.com/spotify/helios/pull/376/files#diff-d935303eb7db0e249425458704c02455R246?

rculbertson: Yea, awaitHasContainerId was a copy and paste of awaitUp, but it didn't need to be. It was awkward having two methods, because you could apply a timeout to both, so you could wait twice as long as the timeout value.

davidxia:
:+1:

PR #378, opened 2015-02-13
master: allow for hash-less job create

danielnorberg:
* Validate job after generating the hash, if not present.
* Add regression test.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/master/resources/JobsResource.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/resources/JobsResource.java
@@ -135,12 +135,12 @@ public class JobsResource {
   @ExceptionMetered
   public CreateJobResponse post(@Valid final Job job,
                                 @RequestUser final String username) {
-    final Collection<String> errors = JOB_VALIDATOR.validate(job);
-    final Job actualJob = job.toBuilder()
+    final Job.Builder clone = job.toBuilder()
         .setCreatingUser(username)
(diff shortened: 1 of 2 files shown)


rohansingh:
Hmm, that didn't go so well.

danielnorberg:
So here's my understanding of what's going wrong.

1. The `HeliosClient` `POST /jobs` a `Job` with a full `id` but no `creatingUser` set.
2. The `HeliosMaster` `POST /jobs` handler sets the `creatingUser` field and rebuilds the job, causing a new hash to be generated.
3. Lose. The new hash does not match the hash that was sent by the client.

There's three ways forward that I can think of right now.

1. Make the CLI and `HeliosClient` users set the `creatingUser` field before posting. This is a breaking change.
2. Exclude the `creatingUser` field from the hash generation. In practice this is what the CLI does today, so even though it's a breaking change it would only impact users of `HeliosClient` that are currently explicitly setting `creatingUser` on the `Job`.
3. Remove hash equality validation from `JobValidator` and leave it up to users to enforce hash equality if they want. My impression is that users today do not really leverage the hash to enforce `Job` consistency between different Helios clusters, so this might be a minimally impacting change.

davidxia:
I'm wondering how it would it break users if we went with choice 1?

rohansingh:
Added a commit that implements the second option that @danielnorberg suggested above — explicitly excluding `creatingUser` from the hash generation.

rohansingh:
Continued in #678.

PR #379, opened 2015-02-14
No longer skip integration tests on Jenkins

rculbertson:
We now have integration tests which use TempJobs and can run on jenkins.


davidxia:
:+1:

PR #380, opened 2015-02-15
Use Java 8 on server side components

rculbertson:
Use Java 8 on server side components, leave client side on Java 7

### DIFF
--- a/helios-client/pom.xml
+++ b/helios-client/pom.xml
@@ -68,6 +68,15 @@
   <build>
     <plugins>
       <plugin>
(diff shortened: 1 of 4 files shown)


davidxia:
Re circleCI failure: I think we need to somehow specify multiple java versions in our `circle.yml`. See https://circleci.com/docs/configuration#dependencies and https://circleci.com/docs/environment#java. It didn't seem obvious so I sent CircleCI a message.

rohansingh:
:-1: in favor of #384

PR #382, opened 2015-02-16

### LINKED ISSUE #362: Remove need to explicitly deregister helios agents?
Remove need to explicitly deregister helios agents when a host with the same hostname as a down agent tries to register. This happens when someone reinstalls their agent machine.

Lots of users get tripped up by this. This behavior doesn't seem to be very useful.

davidxia:
If a host goes DOWN and loses its ZooKeeper host ID file due to
reinstallation or reprovisioning, it cannot come back up without
the user explicitly deregistering it first. Users are confused by
this behavior.

This change will make the host update the mtime of its host ID
node in ZooKeeper every minute.

If a hostname registration collides with an existing record in
the master, the agent will check the mtime of the ID node.
If the ID node hasn't been touched in 10 minutes, the new agent
will deregister the old one and register itself instead of exiting.

Fixes #362

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/agent/AgentZooKeeperRegistrar.java
+++ b/helios-services/src/main/java/com/spotify/helios/agent/AgentZooKeeperRegistrar.java
@@ -23,7 +23,10 @@ package com.spotify.helios.agent;
 
 import com.google.common.util.concurrent.Service;
 
+import com.spotify.helios.master.HostNotFoundException;
+import com.spotify.helios.master.HostStillInUseException;
 import com.spotify.helios.servicescommon.ZooKeeperRegistrarEventListener;
+import com.spotify.helios.servicescommon.ZooKeeperRegistrarUtil;
 import com.spotify.helios.servicescommon.coordination.Paths;
 import com.spotify.helios.servicescommon.coordination.ZooKeeperClient;
 
@@ -34,10 +37,12 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.util.concurrent.TimeUnit;
 
 import static com.google.common.base.Charsets.UTF_8;
 import static com.google.common.base.Optional.fromNullable;
 import static java.lang.String.format;
+import static java.lang.System.currentTimeMillis;
 import static org.apache.curator.framework.recipes.nodes.PersistentEphemeralNode.Mode.EPHEMERAL;
 
 /**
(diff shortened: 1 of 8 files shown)


davidxia:
TODO write tests

@rculbertson What do you think of this approach?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -123,22 +102,9 @@ public ZooKeeperMasterModel(final ZooKeeperClientProvider provider) {
    */
   @Override
   public void registerHost(final String host, final String id) {
-    log.info("registering host: {}", host);
```
davidxia: Basically, I moved `registerHost()` and `deregisterHost()` and any helper methods to a new class `ZooKeeperRegistrarUtil` since I needed to reuse that logic in `AgentZooKeeperRegistrar`.

helios-services/src/main/java/com/spotify/helios/servicescommon/ZooKeeperRegistrarUtil.java
```diff
@@ -0,0 +1,331 @@
+/*
```
davidxia: At first it looked like `ZooKeeperRegistrar` was a good class to put this, but I think that class does something very different and mostly used as an instance whereas registerHost() and deregisterHost() work nicely as static methods.

helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/NodeUpdater.java
```diff
@@ -26,4 +26,6 @@
 public interface NodeUpdater {
 
   boolean update(byte[] bytes) throws IOException;
+
+  boolean touch() throws IOException;
```
davidxia: Added an extra method here for touching files.

rohansingh:
I like the approach. Makes sense.

helios-services/src/main/java/com/spotify/helios/agent/AgentZooKeeperRegistrar.java
```diff
@@ -79,34 +87,34 @@ public void shutDown() throws Exception {
...
     } else {
       final byte[] bytes = client.getData(idPath);
       final String existingId = bytes == null ? "" : new String(bytes, UTF_8);
       if (!id.equals(existingId)) {
-        final String message = format("Another agent already registered as '%s' " +
-                                      "(local=%s remote=%s).", name, id, existingId);
-        log.error(message);
-        agentService.stopAsync();
-        return;
+        final long mtime = client.stat(idPath).getMtime();
```
davidxia: Here's the good stuff.

PR #385, opened 2015-02-16
Do not ignore 2 tests

davidxia:


davidxia:
Any reason we didn't run these 2 tests before?

rohansingh:
Not that I'm aware of. Maybe just an oversight.

:+1:

PR #370, opened 2015-02-07
Add support for tags in service registration ports

carlanton:
Context: #367 

This is one way to do it, don't now if there are better alternatives. In helios-consul we map a registered Helios port to a Consul service, so therefore I think it makes sense to add tags to the ports. I used the mysterious ServicePortParameters class for the tags, and then expose them to helios-consul through ServiceRegistration.Endpoint.

A Helios job with "port tags" looks like this in this implementation:
```
{
  "ports": {
    "my-port": {
      "internalPort": 8080,
      "protocol": "tcp"
    },
    "another-port": {
      "internalPort": 8081,
      "protocol": "tcp"
    }
  },
  "registration": {
    "redis/master": {
      "ports": {
        "my-port": {
          "tags": [
            "master",
            "web-scale"
          ]
        }
      }
    },
    "redis/slave": {
      "ports": {
        "my-port": {
          "tags": [
            "slave",
            "web-scale",
            "meh"
          ]
        }
      }
    }
  }
}
```

What do you think?

PR #386, opened 2015-02-16
Dxia/endpoint tags tests

davidxia:


rohansingh:
Can we close this since @carlanton has added the tests to #370?

PR #387, opened 2015-02-17
TempJobs unit tests no longer use actual env vars

rculbertson:
Previously having certain env vars set (e.g. HELIOS_HOST_FITLER) would cause
the tests to fail. The tests now inject the environment variables they want
used during the test so the tests aren't affected by the environment.


davidxia:
:+1:

helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
```diff
@@ -430,41 +433,46 @@ static String getProfileFromConfig(final Config preConfig) {
...
+  static Builder builder(final String profile, final Map<String, String> env) {
+    return new Builder(profile, TemporaryJobs.loadConfig(), env);
+  }
 
-    Builder() {
-      this(TemporaryJobs.loadConfig());
-    }
+  public static class Builder {
 
-    // I feel like I'm building the y-combinator here because Java insists on the calls to this()
```
davidxia: @drewcsillag Hm, what did you mean by this comment?

PR #388, opened 2015-02-17
Kafka Integration

Kobolog:
This PR is about Kafka support for Helios. Essentially, it augments QueueingHistoryWriter to not only push historical data to ZK, but also to try to send it to Kafka.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/agent/QueueingHistoryWriter.java
+++ b/helios-services/src/main/java/com/spotify/helios/agent/QueueingHistoryWriter.java
@@ -21,16 +21,20 @@
 
 package com.spotify.helios.agent;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
+import com.google.common.base.Optional;
 import com.google.common.base.Supplier;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+
 import com.google.common.util.concurrent.AbstractIdleService;
 import com.google.common.util.concurrent.MoreExecutors;
 
 import com.fasterxml.jackson.core.type.TypeReference;
+
 import com.spotify.helios.common.descriptors.JobId;
 import com.spotify.helios.common.descriptors.TaskStatus;
 import com.spotify.helios.common.descriptors.TaskStatusEvent;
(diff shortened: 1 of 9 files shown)


Kobolog:
This is how Helios posts stuff to Slack, if you're curious why the hell we need Kafka here.

Kobolog:
@rculbertson @rohansingh @davidxia @danielnorberg

drewcsillag:
Cool stuff! :D

rculbertson:
:)

helios-services/src/main/java/com/spotify/helios/agent/QueueingHistoryWriter.java
```diff
@@ -130,6 +158,11 @@ protected void startUp() throws Exception {
   protected void shutDown() throws Exception {
     zkWriterExecutor.shutdownNow();
     zkWriterExecutor.awaitTermination(1, TimeUnit.MINUTES);
+
+    if (kafkaProducer.isPresent()) {
+      // Otherwise it enters an infinite loop for some reason.
```
rohansingh: Would be good to understand why?

helios-services/src/main/java/com/spotify/helios/agent/QueueingHistoryWriter.java
```diff
@@ -272,12 +305,13 @@ public void run() {
         return;
       }
 
+      final JobId jobId = item.getStatus().getJob().getId();
+      final String historyPath = Paths.historyJobHostEventsTimestamp(
+              jobId, hostname, item.getTimestamp());
+      log.debug("writing queued item to zookeeper {} {}", item.getStatus().getJob().getId(),
```
rohansingh: Maybe move this log line back into the `try`.

helios-services/src/main/java/com/spotify/helios/agent/QueueingHistoryWriter.java
```diff
@@ -299,6 +333,17 @@ public void run() {
         putBack(item);
         break;
       }
+
+      try {
+        if (kafkaProducer.isPresent()) {
+          Future<RecordMetadata> future = kafkaProducer.get().send(
```
rohansingh: `final`

helios-services/src/main/java/com/spotify/helios/agent/TaskStatusEventSerializer.java
```diff
@@ -0,0 +1,47 @@
...
+ * Provides serialization support for KafkaProducer used for History subsystem
+ * in {@link com.spotify.helios.agent.QueueingHistoryWriter}.
+ */
+
+public class TaskStatusEventSerializer implements Serializer<TaskStatusEvent> {
+    @Override
+    public void configure(Map<String, ?> configs, boolean isKey) { }
+
+    @Override
+    public byte[] serialize(String topic, TaskStatusEvent value) {
```
rohansingh: Mark params `final` too.

pom.xml
```diff
@@ -77,6 +77,11 @@
       <name>David Xia</name>
       <email>dxia@spotify.com</email>
     </developer>
+    <developer>
+      <id>kobolog</id>
```
rohansingh: :+1: :smiley:

helios-services/src/main/java/com/spotify/helios/agent/QueueingHistoryWriter.java
Kobolog: Docs say that:
    The producer manages a single background thread that does I/O as well as a TCP connection to each of the brokers it needs to communicate with. Failure to close the producer after use will leak these resources.
So I guess maybe it's because the background thread is never stopped or something like that.

helios-services/src/main/java/com/spotify/helios/agent/QueueingHistoryWriter.java
Kobolog: Why? It doesn't throw.

rohansingh: Same reason you have the `log.debug` for sending an event to Kafka in that block. Just to put the log statement next to where that is actually happening, right?

rohansingh:
:+1: if you squash your commits.

PR #389, opened 2015-02-19
Add toString to ServiceRegistration for better logging in registration p...

rculbertson:
...lugins

### DIFF
--- a/helios-service-registration/src/main/java/com/spotify/helios/serviceregistration/ServiceRegistration.java
+++ b/helios-service-registration/src/main/java/com/spotify/helios/serviceregistration/ServiceRegistration.java
@@ -21,6 +21,8 @@
 
 package com.spotify.helios.serviceregistration;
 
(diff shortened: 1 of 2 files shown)


helios-service-registration/src/main/java/com/spotify/helios/serviceregistration/ServiceRegistration.java
```diff
@@ -24,6 +24,7 @@
 import java.util.ArrayList;
 import java.util.List;
 
+import com.google.common.base.Objects;
```
rohansingh: Really nitpicking here but I don't think this `import` should be in the same stanza as the static imports.

rculbertson: Good catch. I moved the imports around manually, and forgot to clean them up. Will push another commit.

rohansingh:
:+1:

PR #390, opened 2015-02-20
[Kafka] Make it possible to actually use Kafka.

Kobolog:
There's no way to actually specify any Kafka nodes in the previous PR since it appears
that there's no way to specify a configuration file for Helios Agent. So, this PR adds
command line argument to provide Kafka bootstrap broker list.


rohansingh:
:+1:

PR #392, opened 2015-02-24
deb: Start helios-agent master when installed

rohansingh:
Follow the convention of every other Debian-packaged daemon and start when
installed, without needing to muck with conf files.


rculbertson:
:+1:

PR #395, opened 2015-02-24
Put jacoco reports in target/jacoco/report

rculbertson:
This is so we can easily make artifacts out of reports without pulling in
other files like the jacoco ant jar which is downloaded into target/jacoco.


rohansingh:
:+1:

PR #397, opened 2015-02-24
deb: Add helios to docker group if it exists

rohansingh:
If the "docker" group exists when installing the Helios agent, add the
"helios" user to it. This allows helios-agent to talk to a vanilla Docker
installation.

Also fix some output redir issues to make the postinst prettier.


rculbertson:
:+1:

PR #398, opened 2015-02-25
WIP healthcheck in job definition

davidxia:


helios-client/src/main/java/com/spotify/helios/common/descriptors/HealthCheck.java
```diff
@@ -0,0 +1,257 @@
...
+
+  public static final String HTTP = "http";
+  public static final String TCP = "tcp";
+  public static final String CMD = "cmd";
+  public static final List<String> VALID_HEALTH_CHECK_TYPES = Lists.newArrayList(HTTP, TCP, CMD);
+
+  private final String type;
+  private final String path;
+  private final String port;
+  private final String cmd;
```
rohansingh: I wonder, do we want to explicitly define `path`, `port`, and `cmd` as first-class members of HealthCheck? I'd lean toward having an `args` dictionary and have each type of health check be responsible for parsing its own `args`.

That way we keep from exploding into a giant list of possible fields over time, most of which don't apply to most health checks.

rculbertson: Another option would be to make each HealthCheck it's own object. This has a few benefits.
1. More type safe than a map of Objects, which can be messy when getting values out of the map because you need to cast everything. Also not sure if a map of Objects will cause issues when deserializing.
2. Users can write very concise code when creating healthchecks in TempJobs, e.g. HeathCheck.newHttpCheck("/healthcheck", 8080);
3. More efficient wrt serialization than a single object, because you will serialize only the fields needed for that health check type.

helios-client/src/test/java/com/spotify/helios/common/JobValidatorTest.java
```diff
@@ -255,6 +261,16 @@ public void testInvalidVolumesFail() {
     assertEquals(newHashSet("Volume source is not absolute: bar"),
                  validator.validate(j.toBuilder().addVolume("/foo", "bar").build()));
   }
+
+  @Test
+  public void testInvalidHealthCheckFail() {
+    final Job j = Job.newBuilder().setName( "foo" ).setVersion( "1" ).
+      setImage( "foobar" ).setHealthCheck( HEALTH_CHECK ).build();
```
Kobolog: I think there should be no spaces inside function call operators.

rohansingh:
I'd like to close this in favor of #401 if that's OK?

PR #399, opened 2015-02-25
Fix bug where agent stops fetching data from zookeeper

rculbertson:
This happened because the agent machine was unable to resolve the ZK host
name into an IP address, which caused the ZK client to throw an
IllegalArgumentException. The agent did not handle this exception properly,
and PersistentPathChildrenCache always thought it was in a synced state.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/PersistentPathChildrenCache.java
+++ b/helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/PersistentPathChildrenCache.java
@@ -169,7 +169,9 @@ public class PersistentPathChildrenCache<T> extends AbstractIdleService {
         try {
           update();
           return;
-        } catch (KeeperException e) {
+        } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


rohansingh:
Personally feel that `IllegalArgumentException` is a ridiculous exception to throw when DNS resolution fails. Grr, ZooKeeper.

:+1:

rculbertson:
Yea, it's not helpful. It generates a list of IP address to connect to, but because it couldn't resolve any of them, the list is empty. At that point it should throw a helpful exception, but instead passes that to some class to make the connections. That class can't do anything with an empty list, so throws the IllegalArgumentException. Oh well.

davidxia:
Wow, nice catch. How did you debug this? bc I couldn't figure this out.

rculbertson:
You had pointed out those networking exceptions in the log before the agent became unresponsive, so I tried to recreate those conditions. I tried bringing down zookeeper while the agent was connected, and then tried having the agent connect to a zookeeper host which doesn't exist to get.

PR #400, opened 2015-02-25
Add healthcheck to job definition

rohansingh:
Note that this does not implement health checks. You can define them in your job, but they will have no effect.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/ExecHealthCheck.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/ExecHealthCheck.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright (c) 2014 Spotify AB.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.helios.common.descriptors;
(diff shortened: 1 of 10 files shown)


rohansingh:
The major change in this over #398 is that each type of health check is in its own class.

Note that I also renamed the `docker exec`-based healthcheck to `ExecHealthCheck` and `{"type": "exec"}`. I think this name better reflects what the check does.

helios-client/src/main/java/com/spotify/helios/common/descriptors/ExecHealthCheck.java
```diff
@@ -0,0 +1,107 @@
...
+import org.jetbrains.annotations.Nullable;
+
+import static com.google.common.base.Strings.isNullOrEmpty;
+
+public class ExecHealthCheck extends HealthCheck {
+
+  private final String cmd;
+
+  public ExecHealthCheck(@JsonProperty("cmd") @Nullable final String cmd) {
+    super(TCP);
```
rculbertson: Should this be exec?

rohansingh: Nope. The `ExecHealthCheck` takes a `cmd` to exec. I guess maybe it could be `command`?

rculbertson: Sorry, I meant the argument to super(). Should it be `EXEC` instead of `TCP`?

rohansingh: Haha. Yup. Sorry.

helios-client/src/main/java/com/spotify/helios/common/descriptors/ExecHealthCheck.java
```diff
@@ -0,0 +1,107 @@
...
+    return Objects.toStringHelper(this)
+        .add("cmd", cmd)
+        .toString();
+  }
+
+  static Builder newBuilder() {
+    return new Builder();
+  }
+
+  public static class Builder {
```
rculbertson: Several of our descriptors have `of` factory methods like [this](https://github.com/spotify/helios/blob/947d06c952422de0143455cfba57163e65debe3c/helios-client/src/main/java/com/spotify/helios/common/descriptors/PortMapping.java#L93). Might be nice to add `of` to these object as well since it's a little more concise than invoking the builder directly.

rohansingh: Added `of` methods for this and the other health check types.

rohansingh:
Closed in favor of #401.

PR #401, opened 2015-02-26
Implement healthchecks

rohansingh:


davidxia:
:+1: This is great. Perhaps someone else should review as well. A lot of code in here.

helios-services/src/main/java/com/spotify/helios/agent/HealthChecker.java
```diff
@@ -0,0 +1,30 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.helios.agent;
+
+import com.spotify.docker.client.DockerClient;
```
rculbertson: Unused import

helios-client/src/main/java/com/spotify/helios/common/descriptors/HttpHealthCheck.java
```diff
@@ -0,0 +1,146 @@
...
+
+import static com.google.common.base.Strings.isNullOrEmpty;
+
+public class HttpHealthCheck extends HealthCheck {
+
+  private final String path;
+  private final String port;
+
+  public HttpHealthCheck(@JsonProperty("path") @Nullable final String path,
+                         @JsonProperty("port") @Nullable final String port) {
```
rculbertson: Any reason to make these fields Nullable? Looks like the builder ensures that they are not null

helios-client/src/main/java/com/spotify/helios/common/descriptors/TcpHealthCheck.java
```diff
@@ -0,0 +1,112 @@
...
+
+import org.jetbrains.annotations.Nullable;
+
+import static com.google.common.base.Strings.isNullOrEmpty;
+
+public class TcpHealthCheck extends HealthCheck {
+
+  private final String port;
+
+  public TcpHealthCheck(@JsonProperty("port") @Nullable final String port) {
```
rculbertson: Does this need to be nullable?

helios-client/src/main/java/com/spotify/helios/common/descriptors/ExecHealthCheck.java
```diff
@@ -0,0 +1,111 @@
...
+
+import org.jetbrains.annotations.Nullable;
+
+import static com.google.common.base.Strings.isNullOrEmpty;
+
+public class ExecHealthCheck extends HealthCheck {
+
+  private final String command;
+
+  public ExecHealthCheck(@JsonProperty("command") @Nullable final String command) {
```
rculbertson: Does this need to be nullable?

helios-client/src/main/java/com/spotify/helios/common/descriptors/HttpHealthCheck.java
rohansingh: Yeah, I think these all need to be nullable since they could be specified as null in JSON.

helios-system-tests/src/main/java/com/spotify/helios/system/HealthCheckTest.java
```diff
@@ -0,0 +1,220 @@
...
+    deployJob(jobId, testHost());
+    awaitTaskState(jobId, testHost(), RUNNING);
+
+    // wait a few seconds to see if the service gets registered
+    Thread.sleep(3000);
+    // shouldn't be registered, since we haven't poked it yet
+    verify(registrar, never()).register(any(ServiceRegistration.class));
+
+    // poke container to get it to start listening on the healthcheck port
+    Polling.await(LONG_WAIT_SECONDS, SECONDS, new Callable<Object>() {
```
davidxia: I'm just trying to understand this test. So we poll the "poke" port until we can make a TCP connection to it. Then we check that `registrar.register()` has been called within `LONG_WAIT_SECONDS` because that means the health check was executed and succeeded?

rohansingh: Actually it's more like this:

1. The container starts listening with `nc` on the poke port, but *not* the health port.
2. We check that it hasn't been registered. It shouldn't be registered since it's not listening on the health port anh should fail the health check.

Then: 

1. We poke the container on the poke port. This is done in a polling loop since we want to keep trying to poke until the poke succeeds. It might fail if `nc` is really slow to start for some reason, so we try until it succeeds.
2. The poke causes the first `nc` to exit, and the second `nc` to start listening on the health port.

We then check to see that the service has been registered now, since it's listening on the health port and should pass the health check.

helios-client/src/main/java/com/spotify/helios/common/descriptors/HttpHealthCheck.java
rohansingh: Note that by the time we're running the health check, we should be pretty confident things aren't null since the `JobValidator` should have caught any null path or port. However, when the object is first constructed, the values could indeed be null.

rculbertson: I think Nullable is used to indicate that a null value is valid and expected, which I don't think it is in this case? There are other parts of the code which assume the getters will return a non-null value. Right now those generate findbugs warnings since they are Nullable. So I think we'd either need to handle nulls in those cases, or make this not nullable.

helios-services/src/main/java/com/spotify/helios/agent/TaskRunnerFactory.java
```diff
@@ -21,9 +21,11 @@
 
 package com.spotify.helios.agent;
 
+import com.google.common.base.Optional;
 import com.google.common.collect.Lists;
 
 import com.spotify.docker.client.DockerClient;
+import com.spotify.helios.common.descriptors.HealthCheck;
```
rculbertson: unused import

helios-client/src/main/java/com/spotify/helios/common/descriptors/ExecHealthCheck.java
```diff
@@ -0,0 +1,110 @@
...
+ * under the License.
+ */
+
+package com.spotify.helios.common.descriptors;
+
+import com.google.common.base.Objects;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import org.jetbrains.annotations.Nullable;
```
rculbertson: unused import

rculbertson:
:+1:

davidxia:
@evgeny-goldin FYI, we took a different approach.

PR #402, opened 2015-02-26
Nitpicky edits, squash if these are good to include

davidxia:


davidxia:
@rohansingh some more nitpicky edits for you :)

rohansingh:
Included in #401.

PR #403, opened 2015-02-26
Get rid of all Travis CI stuff

rohansingh:
We don't use it and it doesn't work for us anyway. Get rid of the various
files and configuration we have scattered around.


davidxia:
:+1:

PR #404, opened 2015-02-26
Make checkstyle fail build for unused imports

rculbertson:
We try to keep unused imports out of our code anyway, so should we have checkstyle just enforce this for us?


rohansingh:
Excellent idea. :+1:

PR #405, opened 2015-02-27
Bump jacoco jar to earlier version

rculbertson:
The most recent one is not compatible with the v1.8 of the Jenkins
jacoco plugin. So use this one until this fix is released.
https://issues.jenkins-ci.org/browse/JENKINS-27088


davidxia:
:+1:

PR #406, opened 2015-02-27
Ignore unknown fields when deserializing Job object

rculbertson:
Without this, masters will reject calls from newer clients which have
additional fields. This allows us to deploy new helios jobs on older
clusters when running helios integration tests using TempJobs.


rohansingh:
:+1:

PR #407, opened 2015-03-02
[CLI] Do not warn on minor CLI version mismatch.

Kobolog:
Do not nag people about mismatched minor versions between CLI and the server, because these versions do not match most of the time.


davidxia:
:+1:

PR #408, opened 2015-03-04
Add HEALTHCHECKING state for tasks

davidxia:
This state is after STARTED but before RUNNING
if health check is defined in the job.


Kobolog:
LGTM.

rohansingh:
:+1:

PR #411, opened 2015-03-04
TCP health check connections need to be closed

rculbertson:


davidxia:
Not sure why CircleCI failed. Rerunning. LGTM though.

PR #412, opened 2015-03-04
Print health check for job inspect cmd

davidxia:


helios-tools/src/main/java/com/spotify/helios/cli/command/JobInspectCommand.java
```diff
@@ -103,12 +122,13 @@ protected int runWithJobId(final Namespace options, final HeliosClient client,
     } else {
       out.printf("Id: %s%n", job.getId());
       out.printf("Image: %s%n", job.getImage());
-      out.printf("Token: %s%n", job.getToken());
       out.printf("Command: %s%n", quote(job.getCommand()));
       printMap(out, "Env:   ", QUOTE, job.getEnv());
+      out.printf("Health check:%n    %s%n", formatHealthCheck(job.getHealthCheck()));
```
rculbertson: Should the first %n be there? Looks like the other items start printing the value on the same line?

davidxia: Yea it can all be on the same line.

rculbertson:
:+1:

PR #414, opened 2015-03-05
Don't kill container if health checks fail.

rculbertson:
One reason we killed the container was so the job wouldn't be stuck in STARTING
state forever, which could be confusing for user. We now have explicit
HEALTHCHECKING state, so it should be clear why the job isn't running.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/agent/TaskRunner.java
+++ b/helios-services/src/main/java/com/spotify/helios/agent/TaskRunner.java
@@ -47,7 +47,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import static com.google.common.base.Preconditions.checkNotNull;
-import static java.util.concurrent.TimeUnit.MINUTES;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
 /**
@@ -57,7 +56,6 @@ class TaskRunner extends InterruptingExecutionThreadService {
 
   private static final Logger log = LoggerFactory.getLogger(TaskRunner.class);
   private static final int SECONDS_TO_WAIT_BEFORE_KILL = 120;
-  private static final long HEALTHCHECK_MAX_RETRY_MILLIS = MINUTES.toMillis(5);
 
   private final long delayMillis;
   private final SettableFuture<Integer> result = SettableFuture.create();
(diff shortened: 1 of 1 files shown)


davidxia:
:+1:

helios-services/src/main/java/com/spotify/helios/agent/TaskRunner.java
```diff
@@ -151,20 +149,11 @@ private int run0() throws InterruptedException, DockerException {
...
                  retryMillis, config, containerId);
-
-        totalMillis += retryMillis;
-        if (totalMillis >= HEALTHCHECK_MAX_RETRY_MILLIS) {
-          docker.killContainer(containerId);
-          throw new RuntimeException("container failed repeated health checks");
-        } else {
-          Thread.sleep(retryMillis);
-        }
+        Thread.sleep(retryMillis);
```
davidxia: Should we just keep health checking forever? The state will stay as HEALTHCHECKING. Should it be marked as FAILED?

rculbertson: If we health check forever, something at a higher level (i.e. helios-helper) will need to decide when to give up, and what action to take (probably undeploy). I think that's ok, and pretty straight forward to reason about.

But I think things get trickier if we put that logic in the agent.

* If health checks fail, should we let the container keep running? If we do, and it eventually becomes healthy, should we detect that and mark it as RUNNING? 
* Should we stop the container? If so, should helios restart it as if it exited with an error code?
* Should we reuse FAILED (which currently means your container exited with an error code) or introduce HEALTH_CHECK_FAILED?
* If we health check forever, callers can decide when to give up. If the agent gives up should we make the timeout configurable?

I'm inclined to let it health check forever, and revisit if it's proves insufficient, but maybe I'm overthinking this?

@rohansingh @matslina?

matslina:
I'm wary of introducing new states unless we absolutely have to. It's messy enough as it is. Letting it health check forever seems fine to me. Job definition can always be extended with an optional restartIfHealthCheckFailsForMoreThanThisManySeconds or something, should we ever need that behaviour.

helios-services/src/main/java/com/spotify/helios/agent/TaskRunner.java
matslina: And commenting here would of course have been more appropriate...

davidxia: Yea we can leave it for now and let helios-helper handle poll for RUNNING. If that doesn't happen within X time, we undeploy it and fail the build. We can add that logic here https://ghe.spotify.net/helios/helios-helper/blob/master/helios_helper/helios_deploy_job.py#L155

davidxia:
:+1:

rohansingh:
:+1:

PR #415, opened 2015-03-05
Add helios-solo (formerly named helios-standalone)

rohansingh:
This includes two things:

* Building the helios-solo Docker image as part of the helios-services build.

* Building a helios-solo Debian package, which contains helios-up and other
  scripts for launching and using helios-solo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,33 +130,20 @@ Install & Run
 -------------
 
 ### Quick start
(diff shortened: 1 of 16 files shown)


danielnorberg:
:+1: :bear: :elephant:

danielnorberg:
s/:bear:/:beer:

rculbertson:
:+1:

PR #416, opened 2015-03-05
Make ReapingTest working with docker 1.5

rculbertson:
This test has been failing for me locally for a long time. The test expects that killing a container gives it an error code -1, but for me it always has exit code 137. Do other see that? (When I test on ubuntu then I get exit code -1).

### DIFF
--- a/helios-system-tests/src/main/java/com/spotify/helios/system/ReapingTest.java
+++ b/helios-system-tests/src/main/java/com/spotify/helios/system/ReapingTest.java
@@ -32,11 +32,13 @@ import com.spotify.helios.client.HeliosClient;
 import org.junit.Test;
 
 import java.security.SecureRandom;
+import java.util.List;
 
 import static com.spotify.helios.common.descriptors.HostStatus.Status.UP;
 import static java.lang.Integer.toHexString;
+import static java.util.Arrays.asList;
 import static java.util.concurrent.TimeUnit.SECONDS;
-import static org.hamcrest.core.Is.is;
+import static org.hamcrest.Matchers.isIn;
 import static org.junit.Assert.assertThat;
 
 public class ReapingTest extends SystemTestBase {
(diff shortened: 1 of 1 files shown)


helios-system-tests/src/main/java/com/spotify/helios/system/ReapingTest.java
```diff
@@ -65,22 +67,21 @@ public void test() throws Exception {
...
-    }
+    // With LXC, killing a container results in exit code 0.
+    // On ubuntu it's -1, and on boot2docker it's 137.
+    final List<Integer> expectedExitCodes = docker.info().executionDriver().startsWith("lxc-") ?
+                                            asList(0) : asList(-1, 137);
 
     // Wait for the agent to kill the container
     final ContainerExit exit1 = docker.waitContainer(intruder1);
-    assertThat(exit1.statusCode(), is(expectedExitCode));
+    assertThat(exit1.statusCode(), isIn(expectedExitCodes));
```
davidxia: @rculbertson I get -1 on my boot2docker version: v1.3.2, Git commit: e41a9ae

rculbertson: I'm running 1.5.0. executionDriver is native-0.2 for me too.

rohansingh:
137 is usually an out-of-memory exception. Perhaps your boot2docker VM is just starved?

helios-system-tests/src/main/java/com/spotify/helios/system/ReapingTest.java
davidxia: I just upgraded from 1.3.2 to 1.5.0 and got 137.

rohansingh: That's interesting. I guess I'm down to merge this, but academically it would be nice to figure out *why* it's different.

rculbertson: It looks like they changed this intentionally.
https://github.com/docker/libcontainer/pull/328

And according to this, returning 137 is the correct thing to do.
http://tldp.org/LDP/abs/html/exitcodes.html

I'll change the commit message to explain this.

PR #419, opened 2015-03-06
Use onescience/alpine instead of uggedal/alpine

rohansingh:
Switch to a better-maintained Alpine image. We already use this for most of
our tests. It also comes with curl preinstalled, which makes the syslog
redirection test faster and more reliable.


rculbertson:
:+1:

tedyoung:
FYI, the gliderlabs/alpine looks to be even more widely used than the onescience image.

PR #423, opened 2015-03-09
Fix destination filename for the helios-solo zip

rohansingh:
Instead of having the filename be a long string containing the Maven
project name and version, just go with "helios-solo.zip". This matches what
we do for "helios-debs.tar.gz".


rculbertson:
:+1:

PR #425, opened 2015-03-09
README: Clarify helios-solo installation steps

rohansingh:
* Explicitly mention that Docker needs to be installed locally.

* Add the helios-apt/go script as a step for Debian/Ubuntu.

* Split out the installation instructions for OS X more clearly.


davidxia:
:+1:

PR #426, opened 2015-03-09
Better helios-solo description in debian/control

rohansingh:
Just to make things nicer for folks who are looking at or installing our
Debian packages.


matslina:
:+1:

PR #427, opened 2015-03-09
solo: Add a versions list to `helios-use`

rohansingh:
Running `helios-use` with no arguments will now print out a list of
available versions.


davidxia:
:+1:

PR #428, opened 2015-03-10
Run integration tests in CircleCI

rohansingh:
Run the integration tests in CircleCI, using helios-solo as the Helios
cluster. This gives us better test coverage and also works out helios-solo.

Also switch to using golang binary archive instead of building from source.
This speeds up the helios-solo image build a bit.


rculbertson:
:+1:  Very cool

PR #429, opened 2015-03-10
solo: Split out spotify/helios-solo-base image

rohansingh:
The new spotify/helios-solo-base image contains the various dependencies
for helios-solo. This lets us build spotify/helios-solo much more quickly
since the dependencies only have to be built once.


davidxia:
:+1:

PR #430, opened 2015-03-10
CircleCI: Bump up parallelism

rohansingh:
We have containers to spare, so use them. Split the system tests across
three containers instead of two.


davidxia:
:+1:

PR #431, opened 2015-03-11
Add integration test for helios-solo

rohansingh:
This test brings up helios-solo in Helios using TemporaryJobs, and then
deploys some jobs to that helios-solo (also using TemporaryJobs).

We test that the following helios-solo functions work:

* Starting helios-solo
* Creating and deploying jobs
* Service registration & discovery

This also contains a bugfix for TemporaryJobs.configureEnv, wherein the
"hostFilter" specified in a profile wasn't being respected if a Helios
endpoint wasn't also explicitly specified.


davidxia:
:+1:

PR #432, opened 2015-03-11
solo: Get rid of helios-enter

rohansingh:
The `helios-enter` command was a handy way to get a running shell in the
solo container from before the days of `docker exec`. No need to still keep
it around.


davidxia:
:+1:

PR #433, opened 2015-03-11
Add jq as dependency of helios-solo

rohansingh:
It's used by `helios-use` to parse the tag list from the Docker registry.


davidxia:
:+1:

PR #438, opened 2015-03-12
solo: Inject right environment in integration test

rohansingh:
Instead of looking at `DOCKER_HOST` and stuff from the environment of the
test runner, we need to be getting it from the environment of the host on
which the helios-solo job is run.

This fixes test failures when running `HeliosSoloIT` against a remote
Helios cluster (as opposed to a local helios-solo instance).

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/HostInfo.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/HostInfo.java
@@ -33,6 +33,8 @@ import com.fasterxml.jackson.annotation.JsonProperty;
  * {
  *   "architecture" : "amd64",
  *   "cpus" : 24,
+ *   "dockerHost" : "unix:///var/run/docker.sock",
+ *   "dockerCertPath" : null,
  *   "dockerVersion" : {
  *     "apiVersion" : "1.12",
  *     "arch" : "amd64",
@@ -68,6 +70,8 @@ public class HostInfo extends Descriptor {
   private final long swapTotalBytes;
   private final long swapFreeBytes;
   private final DockerVersion dockerVersion;
+  private final String dockerHost;
+  private final String dockerCertPath;
 
   /**
    * @param hostname The hostname of the agent.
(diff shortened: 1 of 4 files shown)


davidxia:
LGTM. I'm also checking i can `mvn package` locally

rohansingh:
Check out the latest update. I realized that `getEnvironment()` doesn't actually return what I think it does. It returns the `--env` flags that are set when starting the Helios agent, not the environment variables for the Helios agent itself.

So to get the Docker host and cert path from the agent, I'm now having it explicitly report those.

rohansingh:
So much yak shaving.

helios-integration-tests/src/test/java/com/spotify/helios/HeliosSoloIT.java
```diff
@@ -49,44 +51,53 @@
...
+        // using boot2docker, so use the unix socket endpoint
+        dockerHost = "unix:///var/run/docker.sock";
+        certPath = null;
+      }
+    }
 
     // build the helios-solo job
     final TemporaryJobBuilder solo = temporaryJobs.job()
-        .image(soloImage())
+        .image("registry.spotify.net/spotify/helios-solo:0.8.253")
```
davidxia: Why this change? Will the hardcoded version be a problem?

rohansingh: Oops.

helios-client/src/main/java/com/spotify/helios/common/descriptors/HostInfo.java
```diff
@@ -94,7 +98,9 @@ public HostInfo(@JsonProperty("hostname") final String hostname,
                   @JsonProperty("memoryFreeBytes") final long memoryFreeBytes,
                   @JsonProperty("swapTotalBytes") final long swapTotalBytes,
                   @JsonProperty("swapFreeBytes") final long swapFreeBytes,
-                  @JsonProperty("dockerVersion") final DockerVersion dockerVersion) {
+                  @JsonProperty("dockerVersion") final DockerVersion dockerVersion,
```
davidxia: Maybe update the JSON in the javadoc above?

rohansingh: Done.

PR #443, opened 2015-03-13
solo: Update deb to provide helios-standalone

rohansingh:
In its prior incarnation, helios-solo was known as helios-standalone. This
"Provides" line makes it so that helios-solo is installed in case anyone
still tries to `apt-get install helios-standalone`.


davidxia:
:+1:

PR #444, opened 2015-03-13
Refactor docs a bit

rohansingh:
* Rearrange README so that information that people are looking for in a
  hurry (i.e., how to install and run) are near the top, and the more
  philosophical musings are further down.

* Add instructions for installing Docker to the helios-solo docs.

* Fix some typos and out-of-date stuff.


davidxia:
:+1:

PR #445, opened 2015-03-13
More fixes for the HeliosSoloIT

rohansingh:
* If the DOCKER_HOST is 127.0.0.1, fall back to Unix sockets instead.

* Use the latest version of docker-client, which supports checking the OS
  name directly rather than kernel version (for Boot2Docker detection).

Note that as part of the docker-client upgrade, we need to declare an
explicit dependency on commons-lang:commons-lang:2.6 to resolve a conflict
between docker-client and another package.


davidxia:
:+1:

PR #446, opened 2015-03-13
HealthCheckTest: Use an actual HTTP server, not nc

rohansingh:
The hacky netcat-as-HTTP-server seems to randomly exit sometimes, causing
the test to fail.


davidxia:
:+1:

PR #447, opened 2015-03-13
debian: Depend on headless versions of Java

rohansingh:
No need to force people to install UI kits and fonts and a bunch of other
crap just to run Helios.


davidxia:
:+1:

PR #448, opened 2015-03-13
Restart container if it dies during healthchecking

### LINKED ISSUE #440: Containers that die during health checking will not be restarted
If a container has a health check configured and dies prior to passing the health check, it will never be restarted.

rohansingh:
If a container dies during healthchecking — or at any other time — we
should restart it. Due to a bug, we were not restarting containers that
died while in the HEALTHCHECKING state.

Now, as part of the health check, we ensure the container is still running.
If not, we throw an exception and go to FAILED, which eventually causes the
container to be restarted. Fixes #440.

This also includes a fix for flakiness in `HealthCheckTest`, wherein the
test container would exit immediately after passing the health check. This
could cause the test to fail. Instead, we now make sure the test container
stays up and listening.


davidxia:
:+1:

PR #449, opened 2015-03-14
Get rid of encrypted_secring

rohansingh:
This was used for releases from Travis CI, which we don't do anymore. No
need to keep it hanging around.


davidxia:
:+1:

PR #452, opened 2015-03-18
Use mainline etcd and skydns for helios-solo

rohansingh:
Instead of building with our own weird patches, just use mainline stuff.


davidxia:
:+1:

drewcsillag:
A little late to the party, but does newer mainline etcd still treat entries that start with underscore as hidden?

rohansingh:
@drewcsillag So that's why we needed your patch. Oops.

rohansingh:
@drewcsillag So actually here, we are reading back records that start with underscore from SkyDNS as part of an integration test of helios-solo:
https://github.com/spotify/helios/blob/master/helios-integration-tests/src/test/java/com/spotify/helios/HeliosSoloIT.java#L167

Is that what you would expect to not work? As far as I can tell the hidden-key feature still hasn't been removed from etcd, but this test somehow passes.

drewcsillag:
What I recall is that if you looked up the thing with the underscore, if the discovery record was written as a subtree item of the underscore-prefixed entry, you wouldn't see it, but if it was written directly at the underscore name, it was ok.  I don't recall/know what solo is doing when it puts in it's records into etcd.  Hopefully this makes sense to someone not in my head.

rohansingh:
Great, thanks for the explanation. Found some docs on this page that corroborate that:
https://coreos.com/docs/distributed-configuration/etcd-api/

Basically underscore nodes don't show up in directory lists. I think this is fine for our uses as long as a *hostname* doesn't start with underscore, which would be illegal anyway. Thanks Drew!

PR #454, opened 2015-03-19
Revert "Use mainline etcd and skydns for helios-solo"

rohansingh:
Reverts spotify/helios#452 based on what @drewcsillag pointed out:
https://github.com/spotify/helios/pull/452#issuecomment-83663219


rohansingh:
Actually we shouldn't need this. See #452 for discussion.

PR #455, opened 2015-03-20
solo: Don't use domain from /etc/resolv.conf

rohansingh:
Don't use the domain from /etc/resolv.conf when starting the service
registrar in helios-solo.

This fixes an issue where having a domain present in /etc/resolv.conf
caused helios-solo to fail to start, since helios-skydns isn't capable of
creating a registrar for a specific domain.


davidxia:
we don't need spotify domains?

rohansingh:
Not in helios-solo. The registrar for helios-solo should always just register stuff in the local domain.

On our machines, looking up Spotify services from helios-solo jobs would still work since we have a search domain of "spotify.net". But the helios-solo jobs themselves shouldn't be registered in "spotify.net".

davidxia:
oh i see. it's for registration :+1:

PR #460, opened 2015-03-26
[wip] Expose health checks to service registration plugins

carlanton:
Hi!

In Consul it is possible to define service health checks:
https://www.consul.io/docs/agent/checks.html Since they are pretty similar to
the new health checking functionality in Helios I think it makes sense to
expose the health checks defined in Helios jobs to service registrar plugin.

The health checking plan looks like this:
* During deploy and app boot: Helios runs the health check
* After app boot: Consul runs the health check
* Profit!

It feels a bit strange to re-implement subsets of the job model in the
ServiceRegistration class, but I guess it also makes sense to have them that way.

### DIFF
--- a/helios-service-registration/src/main/java/com/spotify/helios/serviceregistration/ServiceRegistration.java
+++ b/helios-service-registration/src/main/java/com/spotify/helios/serviceregistration/ServiceRegistration.java
@@ -50,12 +50,11 @@ public class ServiceRegistration {
   public static class Builder {
 
     private List<Endpoint> endpoints = new ArrayList<>();
-
     @Deprecated
     public Builder endpoint(final String name,
                             final String protocol,
                             final int port) {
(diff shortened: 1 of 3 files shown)


carlanton:
I will write some tests for this...

rohansingh:
Looks like the tests are failing due to `DockerTimeoutException`. We see this happen sometimes, I think it has to do with performance issues with either Docker daemon or the Docker Hub when we try to pull public images as part of integration tests.

I'll try to retrigger again in a little bit.

carlanton:
I added some tests but CircleCI died unexpectedly :-)

davidxia:
@carlanton I've retriggered CircleCI (I think you should be able to as well). We'll see if it passes this time.

rohansingh:
:+1: from me. @davidxia?

PR #461, opened 2015-03-27
Fix race condition when fetching list of jobs

rculbertson:
We would first get a list of all the job nodes, then retrieve details
about each indivdual job. If a job was deleted before we read the details,
NoNodeException would be thrown, and the request would fail. We now
ignore that exception.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
@@ -394,9 +394,14 @@ public class ZooKeeperMasterModel implements MasterModel {
       for (final String id : ids) {
         final JobId jobId = JobId.fromString(id);
         final String path = Paths.configJob(jobId);
-        final byte[] data = client.getData(path);
-        final Job descriptor = parse(data, Job.class);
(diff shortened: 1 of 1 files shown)


helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -394,9 +394,13 @@ private Job getJob(final ZooKeeperClient client, final JobId id) {
...
         final String path = Paths.configJob(jobId);
-        final byte[] data = client.getData(path);
-        final Job descriptor = parse(data, Job.class);
-        descriptors.put(descriptor.getId(), descriptor);
+        try {
+          final byte[] data = client.getData(path);
+          final Job descriptor = parse(data, Job.class);
+          descriptors.put(descriptor.getId(), descriptor);
+        } catch (NoNodeException e) {
+          // Ignore, the job was deleted before we had a chance to read it.
```
rohansingh: Don't even want to log this?

rculbertson: Yea logging it is a good idea. Just pushed that.

davidxia:
:+1: seems more likely this will happen the more jobs there are.

PR #467, opened 2015-04-21
solo: Better method of detecting Spotify network

rohansingh:
Still sucks that we have to do this, but this way is more reliable than
the old way.


davidxia:
:+1:

PR #470, opened 2015-04-22
Update docs for helios-solo

### LINKED ISSUE #463: Need to start docker service before running helios
It would probably be nice to add the requirement to start docker service before doing helios-up to the docs:
https://github.com/spotify/helios/blob/master/docs/helios_solo.md#linux

Otherwise it will fail and you will get a kind of confusing error message when running the suggested ```docker info``` command:

```
FATA[0000] Get http:///var/run/docker.sock/v1.17/info: dial unix /var/run/docker.sock: no such file or directory. Are you trying to connect to a TLS-enabled daemon without TLS?
```

davidxia:
Emphasize that docker needs to be started for helios-solo to work.

fixes #463

### DIFF
--- a/docs/helios_solo.md
+++ b/docs/helios_solo.md
@@ -11,8 +11,9 @@ Install & Run
 
 ### OS X
 
-Ensure you have installed Docker using [Boot2Docker](http://boot2docker.io/) or
-another similar solution. Then:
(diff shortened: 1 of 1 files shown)


rohansingh:
:+1:

PR #476, opened 2015-04-29
Do not health check already RUNNING containers

davidxia:
Do not health check already RUNNING containers when
the Helios agent process is restarted, eg on deploys of Helios itself.


rohansingh:
:+1:

PR #478, opened 2015-05-04
Hack to make "helios status" output valid JSON when no jobs match

gimaker:


gimaker:
@rculbertson @davidxia @rohansingh

davidxia:
:+1:

PR #482, opened 2015-05-18
Implement docker exec healthchecks.

davidxia:
* Only run exec healthcheck IT on compatible docker versions.
* Make the exec healthcheck command a list.
* Simplify system test by storing job config in a file.
* Don't allow multi-command exec healthchecks in CLI.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/agent/HealthCheckerFactory.java
+++ b/helios-services/src/main/java/com/spotify/helios/agent/HealthCheckerFactory.java
@@ -21,18 +21,27 @@
 
 package com.spotify.helios.agent;
 
+import com.google.common.base.Splitter;
+import com.google.common.base.Strings;
+import com.google.common.collect.Iterables;
+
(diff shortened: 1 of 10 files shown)


helios-services/src/main/java/com/spotify/helios/agent/HealthCheckerFactory.java
```diff
@@ -55,19 +64,76 @@ public static HealthChecker create(final TaskConfig taskConfig, final DockerClie
...
+      } catch (DockerException e) {
+        return false;
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        return false;
+      }
+    }
+
+    private static boolean compatibleDockerVersion(final DockerClient docker) {
+      final String executionDriver;
```
rohansingh: One thing i just realized, I believe we already have the Docker version in the agent state somewhere. That's what we get from `helios hosts`. So we should use that instead of making a new Docker request.

PR #483, opened 2015-05-19
Fix flapping test by lowering DEFAULT_FLAPPING_RESTART_COUNT

davidxia:


davidxia:
@rculbertson @gimaker

davidxia:
This is a slightly ugly fix for this test that always times out. It times out because the container doesn't exit 10 times in 200 seconds. Instead of increasing the timeout which will increase our test duration, I just changed the field to be public and non-final and alter it before the test starts.

We could have a set method for that field in the FlapController, but then we'd need to figure out a way to pass some value through a dozen layers of factories and builders and other crap.

gimaker:
:-1: I'd rather have longer tests than mutable public static fields, ew

Question: should it exit 10 times in 200 seconds? If yes, then why does the test fail? By virtue of being too timing dependence?

rohansingh:
@gimaker Yeah, I think it's the timing that's a problem. On cloudy CI boxes especially, we can fail to get into a flapping state pretty much ever.

If mutable public static field is so bad, maybe we should just make this and related properties configurable for real via agent arguments.

gimaker:
So... that means that it takes helios over 20 seconds to (re)start the container on average? That sounds fishy. Is that expected?

davidxia:
Longer timeout here https://github.com/spotify/helios/pull/484

davidxia:
closed by https://github.com/spotify/helios/pull/484

PR #490, opened 2015-06-03
[solo] fix params passed to helios-cleanup

rculbertson:
When helios-cleanup was calling undeploy with -f, which isn't a valid option. It was
calling remove with --force, which has been deprecated in favor of --yes. We now use
--yes in both places.


rouzwawi:
:+1:  thanks for the quick fix

davidxia:
:+1:

PR #496, opened 2015-06-09
Update helios inspect for exec health check

davidxia:


davidxia:
@gimaker

gimaker:
where is the unit test david?

PR #499, opened 2015-06-23
TemporaryJobs: Change the test report output dir

rohansingh:
Instead of helios-test-reports/ use helios-reports/test/

We will also have other subdirs like helios-reports/deploy/ so this lets us
collect everything in one place.


rohansingh:
@rculbertson @gimaker

gimaker:
:+1: assuming it correctly creates the (nested) directory, does it?

rohansingh:
Yes, it does.

PR #501, opened 2015-06-24
TemporaryJobs: Make event logging more robust

rohansingh:
1. If we can't determine the current test or class name (due to having
   threaded tests), just log to stdout.

2. If we can't create a JSON log output file, log it but don't crash.


rohansingh:
@gimaker

gimaker:
:+1:

PR #504, opened 2015-06-28
Implement labels for agents

rohansingh:
Agents can now have arbitrary labels. You can now list only agents that
match the label that you care about:

    helios hosts --label some_label

In the future, we can use labels for other batch operations against hosts.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/HostStatus.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/HostStatus.java
@@ -21,9 +21,7 @@
 
 package com.spotify.helios.common.descriptors;
 
-import com.google.common.base.Function;
 import com.google.common.base.Joiner;
-import com.google.common.collect.Iterables;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 
@@ -31,7 +29,6 @@ import org.jetbrains.annotations.Nullable;
 
 import java.util.Collections;
 import java.util.Map;
-import java.util.Map.Entry;
 
 import static com.google.common.base.Optional.fromNullable;
 import static com.google.common.base.Preconditions.checkNotNull;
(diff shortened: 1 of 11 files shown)


helios-client/src/main/java/com/spotify/helios/common/descriptors/HostStatus.java
```diff
@@ -91,13 +93,15 @@
    * @param hostInfo The host information.
    * @param agentInfo The agent information.
    * @param environment The environment provided to the agent on it's command line.
+   * @param labels The labels assigned to the agent.
```
davidxia: Can we also update the javadoc above?

helios-tools/src/main/java/com/spotify/helios/cli/command/HostListCommand.java
```diff
@@ -108,6 +116,12 @@ int run(final Namespace options, final HeliosClient client, final PrintStream ou
     }
 
     final List<String> sortedHosts = natural().sortedCopy(hosts);
+    final List<String> selectedLabels = options.getList(labelArg.getDest());
+
+    if (selectedLabels != null && !selectedLabels.isEmpty() && json) {
+      System.err.println("Warning: filtering by label is not supported for JSON output. Not doing"
```
davidxia: I guess we don't want to do this because then the JSON output will be slow?

rohansingh: It's more because we need the host status to filter, and right now we don't get the host status if all you ask for is JSON output. We could add this functionality, but if you're doing JSON output you can always filter it yourself with `jq` or similar anyway.

helios-client/src/main/java/com/spotify/helios/common/descriptors/HostStatus.java
rohansingh: Done.

davidxia:
:+1:

Issue #473, opened 2015-04-24
Improve TemporaryJob opaque err msg and don't let user set name


davidxia:
Petter tried to start an elastic search tempjob

```java
    log.info("starting elastic search");
    elasticSearch = temporaryJobs.job()
        .image("elasticsearch:latest")
        .name("elasticsearch")
        .port("tcp", 9300)
//        .registration("elasticsearch", "tcp", "tcp")
        .deploy();
```
It fails with

```
java.lang.AssertionError: Failed to create job elasticsearch:null - CreateJobResponse{status=INVALID_JOB_DEFINITION, errors=[Job hash was not specified in job id [elasticsearch:null:fd23f2e0da5c96f1bef0fa6a32f005c4ed90edcd].], id=elasticsearch:null:9620f7ed8ca9ef4785474b1ddf46057652a39db6}
    at org.junit.Assert.fail(Assert.java:88)
    at com.spotify.helios.testing.TemporaryJob.deploy(TemporaryJob.java:174)
    at com.spotify.helios.testing.DefaultDeployer.deploy(DefaultDeployer.java:135)
    at com.spotify.helios.testing.DefaultDeployer.deploy(DefaultDeployer.java:95)
    at com.spotify.helios.testing.TemporaryJobBuilder.deploy(TemporaryJobBuilder.java:287)
    at com.spotify.helios.testing.TemporaryJobBuilder.deploy(TemporaryJobBuilder.java:248)
    at com.spotify.sys.acceptance.World.startElasticSearch(World.java:123)
    at com.spotify.sys.acceptance.AcceptanceIT.startUp(AcceptanceIT.java:13)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
    at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
    at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
    at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:24)
    at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
[...]

PR #510, opened 2015-07-05
Fix NPE in TempJobs

davidxia:
The job hash can be null or empty when user sets the TempJob name.
Check for null or empty job hashes when trying to get a description.


rohansingh:
Is this PR deprecated by #473?

davidxia:
@rohansingh I guess so? Closing then

PR #513, opened 2015-07-05
Say what Docker version Helios has been tested on

### LINKED ISSUE #342: Users can find out easily which versions of Helios are compat with which versions of Docker
Sometimes users don't know which versions of Helios are compatible with which versions of Docker. Is this information hard to find out? Maybe `helios --version` can say that?

davidxia:
Include this in CLI's `helios --version`.

Fixes #342


rohansingh:
:+1: (assuming test failure is spurious)

davidxia:
Yea the HealthCheckTest failed. I'm assume that's unrelated.

PR #516, opened 2015-07-06
Fail softly on task status retrieval failure

davidxia:
Sometimes bad data gets written to zookeeper and the master
cannot deserialize the data. Fail softly by skipping bad
task statuses instead of failing the entire attempt.


davidxia:
@rculbertson

rculbertson:
:+1:

rohansingh:
Maybe also log stuff? I would also be interested to know how or when this happens. Bad data shouldn't just get written.

davidxia:
@rohansingh We `log.debug` here https://github.com/spotify/helios/pull/516/files#diff-8d500c531ffb1f7836e9d5b60585a002R850. Should we increase the log level?

rculbertson:
Turns out the bad data was being written to disk by an agent, and it would put it back into ZK every time the agent restarted (or reconnected to ZK for some other reason). Still seems like a good PR though...better if 'helios hosts' doesn't break for an entire site because of one agent.

rohansingh:
:+1:

PR #517, opened 2015-07-06
Add flag to not register master with zookeeper

davidxia:
For debugging purposes when you want to connect to zookeeper
but not register the master process.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/servicescommon/ServiceParser.java
+++ b/helios-services/src/main/java/com/spotify/helios/servicescommon/ServiceParser.java
@@ -65,6 +65,7 @@ public class ServiceParser {
   private final Argument zooKeeperConnectiontimeoutArg;
   private final Argument zooKeeperNamespace;
   private final Argument zooKeeperClusterId;
+  private final Argument noZooKeeperRegistrationArg;
   private final Argument noMetricsArg;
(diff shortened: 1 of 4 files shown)


rculbertson:
:+1:

rohansingh:
:+1:

gimaker:
:+1:

rohansingh:
Looks like the build failure is partially due to the issue addressed in #522.

helios-services/src/main/java/com/spotify/helios/servicescommon/ServiceParser.java
```diff
@@ -230,6 +236,10 @@ public String getZooKeeperClusterId() {
     return options.getString(zooKeeperClusterId.getDest());
   }
 
+  public Boolean getNoZooKeeperRegistration() {
+    return fromNullable(options.getBoolean(noZooKeeperRegistrationArg.getDest())).or(true);
```
rohansingh: I think this is why the tests are failing. Doesn't the true make it so the master never registers by default?

Part of the master registration is creating the appropriate root nodes in ZooKeeper. Not creating those nodes would explain why the SoloIT fails.

davidxia: doh

PR #518, opened 2015-07-09
TemporaryJobs: Change the test report output dir

davidxia:
Instead of .helios-reports/ use target/helios-reports
because people complained. :(

### DIFF
--- a/helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
@@ -87,7 +87,7 @@ public class TemporaryJobs implements TestRule {
   private static final Prober DEFAULT_PROBER = new DefaultProber();
   private static final String DEFAULT_LOCAL_HOST_FILTER = ".+";
   private static final String DEFAULT_PREFIX_DIRECTORY = "/tmp/helios-temp-jobs";
-  private static final String DEFAULT_TEST_REPORT_DIRECTORY = "./helios-reports/test";
+  private static final String DEFAULT_TEST_REPORT_DIRECTORY = "target/helios-reports/test";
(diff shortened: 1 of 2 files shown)


davidxia:
@rculbertson

davidxia:
@gilles

helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
```diff
@@ -87,7 +87,7 @@
   private static final Prober DEFAULT_PROBER = new DefaultProber();
   private static final String DEFAULT_LOCAL_HOST_FILTER = ".+";
   private static final String DEFAULT_PREFIX_DIRECTORY = "/tmp/helios-temp-jobs";
-  private static final String DEFAULT_TEST_REPORT_DIRECTORY = "./helios-reports/test";
+  private static final String DEFAULT_TEST_REPORT_DIRECTORY = "target/helios-reports/test";
```
gilles: This assumes you're running with maven or any build system that relies on the maven directory structure

davidxia: Perhaps it's better to just have users remember to configure it to output to target or add to .gitignore?

gilles: Maybe this could be null which means disabled by default?

rohansingh:
:+1:

helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
gimaker: if you don't use maven, and consequently don't have the target directory, then the only thing that will happen is that we create the target directory and put stuff in there. is that really so bad?

gimaker:
:+1:

PR #519, opened 2015-07-10
Explicitly use :latest images for system tests

rohansingh:
By using X:latest images instead of just X, we avoid pulling all tags for
the given image. This should hopefully improve test speed and reliability,
since a lot of tests fail due to timeouts while pulling public images.


rohansingh:
@davidxia @gimaker @rculbertson

davidxia:
@rohansingh Also here? https://github.com/spotify/helios/blob/master/helios-system-tests/src/main/java/com/spotify/helios/system/PredefinedPortImageDeploymentTest.java#L54

gimaker:
omg yes :+1:

PR #522, opened 2015-07-10
solo: Block on starting ZooKeeper

rohansingh:
Ensure that ZooKeeper is started before starting the agent or master. This
will help mitigate issues where the agent or master starts up too soon and
can't connect to ZK.


rohansingh:
@gimaker

davidxia:
:+1:

PR #523, opened 2015-07-13
Add some additional output for test failures

rohansingh:
This will help debug common test failures.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
@@ -300,6 +300,9 @@ public class ZooKeeperMasterModel implements MasterModel {
         }
         throw new JobExistsException(id.toString());
       }
+    } catch (NoNodeException e) {
+      throw new HeliosRuntimeException("adding job " + job + " failed due to missing ZK path: " +
+                                       e.getPath(), e);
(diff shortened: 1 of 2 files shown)


rohansingh:
@davidxia

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -300,6 +300,9 @@ public void addJob(final Job job) throws JobExistsException {
         }
         throw new JobExistsException(id.toString());
       }
+    } catch (NoNodeException e) {
```
davidxia: It's the `client.transaction()` that throws the `NoNodeException` right? How can this happen with `ensurePath()`?

rohansingh: If the /config or /config/jobs paths haven't been created for some reason then the `create` ops in the transaction above will throw.

gimaker:
:+1:

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
davidxia: Should we add an `ensurePath()` for those paths as well? Does this happen when the master is slow to startup?

PR #525, opened 2015-07-13
Kill container based on name if ID not available

rohansingh:
If a job is undeploy right after it's deployed, we may have asked Docker to
create a container but still not gotten the ID back. In the past, we would
then leave this orphaned container around.

Now, if we don't have a container ID, we attempt to kill any container that
matches the predetermined container name to avoid this race condition.


rohansingh:
@davidxia @gimaker 

This should fix all those damn build errors due to `DeploymentTest`.

gimaker:
:+1: (but this is an area I'm certainly a novice in, so you should probably wait for another thumbs-up)

PR #527, opened 2015-07-14
ZooKeeperMasterModel: Fix log level for a message

rohansingh:
The log message for getDeploymentGroupHosts was incorrectly at ERROR
instead of DEBUG.


davidxia:
:+1:

PR #528, opened 2015-07-14
Don't log warning when stopping container by name

rohansingh:
If a TaskRunner is told to stop a container before it's gotten back the
container ID from starting the container, it will instead try to stop the
container by name. However, this might throw a ContainerNotFoundException
if the underlying container actually hasn't been started yet.

In this case, don't log a warning like we normally would.


davidxia:
:+1:

PR #530, opened 2015-07-14
Fix issues with how TaskRunner stops containers

rohansingh:
There were a couple bugs with how TaskRunner keeps track of and stops
containers. These would also cause JobExpirationTest to fail.


rohansingh:
@davidxia This should fix at least one of the failures we see in Jenkins.

davidxia:
:+1:

PR #531, opened 2015-07-14
More flexible host selectors for deployment groups

gimaker:
* Support inequality conditions. E.g. 'role=my_role canary!=yes'
 * Use a more storage format for host selectors, enabling implementing
   more advanced host selectors in the future.

Note that the host list command has not be updated (yet!), and currently
does not support the != operator.


gimaker:
@rohansingh @rculbertson @davidxia

rohansingh:
Rebase on master to get the test fix from #532.

rohansingh:
OK, I just rebased. Had to force push though.

davidxia:
:+1: After a quick read through. Maybe use more immutable lists? :)

gimaker:
@davidxia: there are now more immutable lists

PR #532, opened 2015-07-14
MultipleJobsTest: Wait for agent to come up

rohansingh:
MultipleJobsTest was trying to deploy to an agent before the agent was even
up or registered. This would occasionally result in test failures.


gimaker:
:+1:

PR #533, opened 2015-07-15
HeliosSoloIT: Use dynamic ports

rohansingh:
We have no particular reason to need static ports for this test. Using
dynamic ports reduces the likelihood of a port conflict.


davidxia:
:+1:

PR #534, opened 2015-07-15
Only include new Deployment fields if non-null

rohansingh:
This will let new clients deploy to older masters if they don't need the
new fields. Basically, backwards-compatibility.


rohansingh:
@davidxia

PR #535, opened 2015-07-15
Ignore unknown JSON in deserialization

davidxia:
This allows us to add fields to JSON read by the masters
without our integration tests breaking.


davidxia:
@rohansingh @gimaker Ignore all the unknowns?

rohansingh:
:+1:

PR #536, opened 2015-07-15
Moving rollingupdate host logic into HostMatcher

rohansingh:
Move the logic for determining which hosts match a deployment group into
RollingUpdateService.HostMatcher and add tests for it. This includes a fix
so that deployment groups with no selector match no hosts (rather than all
hosts).


rohansingh:
@gimaker

gimaker:
:+1:

PR #537, opened 2015-07-15
Rename job to jobId in DeploymentGroup

davidxia:
To be consistent with other classes like Deployment

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/DeploymentGroup.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/DeploymentGroup.java
@@ -30,10 +30,8 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import org.jetbrains.annotations.Nullable;
 
 import java.util.List;
-import java.util.Map;
 
 import static java.util.Collections.emptyList;
-import static java.util.Collections.emptyMap;
 
 /**
  * Represents a Helios deployment group.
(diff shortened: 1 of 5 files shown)


davidxia:
@rohansingh @gimaker

rohansingh:
:+1:

helios-client/src/main/java/com/spotify/helios/common/descriptors/DeploymentGroup.java
```diff
@@ -55,30 +53,29 @@
...
-   * @param job The job for the deployment group.
+   * @param jobId The job ID for the deployment group.
    * @param hostSelectors The selectors that determine which agents are part of the deployment
    *                       group.
    */
   public DeploymentGroup(
       @JsonProperty("name") final String name,
       @JsonProperty("hostSelectors") final List<HostSelector> hostSelectors,
-      @JsonProperty("job") @Nullable final JobId job,
+      @JsonProperty("job") @Nullable final JobId jobId,
```
gimaker: should the json name also be jobId, then?

rohansingh: If we change it what will happen to existing stuff? We just clear out all the deployment groups we've created in testing?

gimaker: yes. if we don't, them those deployment groups will be defunct -- we might not even be able to remove them at that point.

davidxia: This is just internal variable naming. The json key will still be "job" like in Deployment.java

PR #538, opened 2015-07-15
HealthCheckTest: Fix intermittent failures

### LINKED ISSUE #529: HealthCheckTest is flaky
`HealthCheckTest` occasionally fails due to an exited container or other issues. Here's an example:
https://circleci.com/gh/spotify/helios/2141

rohansingh:
The HealthCheckTest runs netcat in some containers via TemporaryJobs and
then probes those netcat instances to make sure they're reachable. When
netcat is probed, it usually exits immediately. Generally we wouldn't care
since the test ends right after the probe.

However, if TemporaryJobs noticed that the job exited while the test was
still running, it would throw an exception due to the job being unhealthy.
The test would then fail.

So now, we run netcat with `-k` so it sticks around as a persistent server
and the container doesn't exit prematurely.

Fixes #529.


davidxia:
:+1:

gimaker:
:+1:

PR #541, opened 2015-07-16
Update maven-replacer-plugin

davidxia:
Newer version is marked thread safe so parallel builds don't show
a warning.


rohansingh:
:+1:

PR #543, opened 2015-07-16
status-deployment-group: Print host selectors

gimaker:
The deployment-group status endpoint now returns the DeploymentGroup
in its entirety, including host selectors.


gimaker:
@davidxia @rohansingh

davidxia:
:+1:

PR #546, opened 2015-07-16
Fix deployment group status when no rolling-update

davidxia:


davidxia:
@gimaker @rohansingh

gimaker:
:+1:

PR #547, opened 2015-07-16
Rename CLI command status-deployment-group -> deployment-group-status

gimaker:


gimaker:
@rohansingh @davidxia

davidxia:
no strong preferences here. If we're still not super committed to these command names, we can warn early adopters that these names could change for now. :+1:

rohansingh:
+:100:

PR #549, opened 2015-07-17
Make QueueingHistoryWriter work for generic events

rohansingh:
The QueueingHistoryWriter was used to write task status events to ZooKeeper
and Kafka. Unfortunately it was very coupled to TaskStatusEvent.

Now we've moved it to servicescommon and made it generic and abstract so it
can be used for writing all types of events. The class that writes events
for tasks is now named TaskHistoryWriter and extends QueueingHistoryWriter.


rohansingh:
@davidxia @gimaker This is the first step to being able to record rolling update events. Still need to implement a `RollingUpdateHistoryWriter` that extends `QueuingHistoryWriter`.

helios-services/src/main/java/com/spotify/helios/servicescommon/QueueingHistoryWriter.java
```diff
@@ -0,0 +1,455 @@
...
+ *    should not block on ZK being in any particular state, and ideally not while a file write is
+ *    occurring, as the file may get large if ZK has been away for a long time.
+ *
+ * 2. We set limits on the maximum number of events stored at any particular ZK path, and also the
+ *    overall total number of events.
+ *
+ * To use this class, implement a QueueingHistoryWriter for a specific type of event and call the
+ * add(TEvent) method to add an event.
+ */
+public abstract class QueueingHistoryWriter<TEvent>
```
gimaker: Composition > inheritance?

PR #550, opened 2015-07-17
CircleCI: Rerun failing tests

rohansingh:
If an indivdual test fails, rerun it up to 2 times. If it passes on a
subsequent run, it will get reported as "flaky" but won't fail the build.


davidxia:
:+1:

PR #551, opened 2015-07-17
Rename CLI command abort-rolling-update to stop-deployment-group

gimaker:
This more accurately reflects what the command actually does - as it
doesn't just abort a rolling-update update, but also stops an ACTIVE
deployment-group, for example.


gimaker:
@davidxia @rohansingh

rohansingh:
:+1:

PR #552, opened 2015-07-17
remove-deployment-group: Fix output

gimaker:
Don't say it failed when it didn't. Also add some tests for the CLI, and
remove the corresponding system tests.


gimaker:
@rohansingh @davidxia

rohansingh:
:+1:

PR #554, opened 2015-07-20
CircleCI - Support free plan by default

gtonic:
Default parallelism is set to *5*, don't know if this is by intention. *4* is the maximum of the free plan and would make it easier to contribute.

### DIFF
--- a/circle.sh
+++ b/circle.sh
@@ -3,7 +3,7 @@
 case "$1" in
   pre_machine)
     # ensure correct level of parallelism
-    expected_nodes=5
+    expected_nodes=4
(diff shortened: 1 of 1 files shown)


rohansingh:
@gtonic We actually need all those nodes to run the tests. If you only run with 4 nodes, then one of the integration tests suites doesn't run.

We do build PR's from forks though. For example: https://circleci.com/gh/spotify/helios/1701

I think in this case you've setup CircleCI explicitly for your fork, so it's using your account with the free plan. If you didn't have that and instead just open a PR, then the build will run with the Spotify plan that supports the parallelism that we need.

PR #556, opened 2015-07-20
Handle error conditions in rollingUpdateAwaitRunning

rculbertson:
1) If a job was undeployed with 'helios undeploy' during a rolling update,
there was a race condition where we would loop on rollingUpdateAwaitRunning
forever. We now detect this case by checking the job is still listed under
/config/hosts/<host> and fail if it has been removed.

2) rollingUpdateAwaitRunning now honors timeout if it gets stuck while waiting
for status of job to appear.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
@@ -627,26 +627,31 @@ public class ZooKeeperMasterModel implements MasterModel {
     final Map<JobId, TaskStatus> taskStatuses = getTaskStatuses(client, host);
 
     if (!taskStatuses.containsKey(deploymentGroup.getJobId())) {
-      // job hasn't shown up yet, probably still being written
+      // Handle cases where agent has not written job status to zookeeper.
+
+      // If job is not listed under /config/hosts node, it may have been deployed successfully and
+      // then manually undeployed. The job will not get redeployed, so treat this as a failure.
+      final Deployment deployment = getDeployment(host, deploymentGroup.getJobId());
+      if (deployment == null) {
+        return RollingUpdateTaskResult.error(
+            "Job unexpectedly undeployed. Perhaps it was manually undeployed?");
+      }
+
+      // Check if we've exceeded the timeout for the rollout operation.
+      if (isRolloutTimedOut(deploymentGroup, client)) {
+        return RollingUpdateTaskResult.error("timed out while retrieving job status");
+      }
+
+      // We haven't detected any errors, so assume the agent will write the status soon.
       return RollingUpdateTaskResult.TASK_IN_PROGRESS;
     } else if (!taskStatuses.get(deploymentGroup.getJobId()).getState()
         .equals(TaskStatus.State.RUNNING)) {
       // job isn't running yet
 
-      try {
-        final String statusPath = Paths.statusDeploymentGroup(deploymentGroup.getName());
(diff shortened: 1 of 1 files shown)


helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -627,26 +627,30 @@ private RollingUpdateTaskResult rollingUpdateAwaitRunning(final DeploymentGroup
...
 
     if (!taskStatuses.containsKey(deploymentGroup.getJobId())) {
-      // job hasn't shown up yet, probably still being written
+      // Handle cases where agent has not written job status to zookeeper.
+
+      // If job is not listed under /config/hosts node, it was probably deployed successfully and
+      // then manually undeployed. The job will not get redeployed, so treat this as a failure.
+      final Deployment deployment = getDeployment(host, deploymentGroup.getJobId());
+      if (deployment == null) {
+        return RollingUpdateTaskResult.error("job was manually undeployed");
```
gimaker: There are fringe cases where the "manual" part of this statement is not quite true. E.g. if someone created an overlapping deployment group and used rolling-update with the `--migrate` flag.

Maybe something along the lines of:

"Job unexpectedly undeployed on host %s. Perhaps it was undeployed manually?"

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -656,7 +660,7 @@ private RollingUpdateTaskResult rollingUpdateAwaitRunning(final DeploymentGroup
       // won't be able to undeploy the job on the next update.
       final Deployment deployment = getDeployment(host, deploymentGroup.getJobId());
       if (deployment == null) {
-        return RollingUpdateTaskResult.error("deployment for this job is very broken in ZK");
+        return RollingUpdateTaskResult.error("deployment for this job not found in zookeeper");
```
davidxia: Maybe a bit more info about how this might've happened like the message here https://github.com/spotify/helios/pull/556/files#r35044609?

gimaker:
tests would be nice... but... :+1:

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
rculbertson: Good point. Changed the message.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
rculbertson: @staffan do you remember how we got this message to occur earlier? Not sure how we would end up here.

gimaker: same way as the other one -- undeploy -a --yes <job-id>

rculbertson: Ah right. If you undeploy and the updater kicks in before the agent removed the status. Will updated the message.

davidxia:
:+1: Yea spending some time to create tests for these edge cases will be good to do in the future.

gimaker:
:+1:

rculbertson:
Agreed tests would be good for this. This isn't easy right now because the class is structured in a way that makes it difficult to inject unexpected data. It would be good to refactor this so more methods are made public, and data is easily injected, so we can test at a more granular level.

PR #558, opened 2015-07-21
Change some rolling update logs to DEBUG, not INFO

rohansingh:
Logging every rollingupdate step and update of deployment group hosts just
floods the info.log.


rculbertson:
:+1:

PR #559, opened 2015-07-21
Perform rolling-update even if deployment group hasn't changed

rculbertson:
When rolling-update is called, the server logic was to perform the operation
only if the job ID or rollout options had changed. This would cause a problem
if a rollout failed, and the operation needed to be performed again. The code
would detect that the desired state had not changed and do nothing. We now
perform the rollout every time to ensure the actual state matches the desired
state.


rculbertson:
@gimaker @davidxia

rohansingh:
:+1:

rohansingh:
:shipit:

rculbertson:
Added a test.

davidxia:
:+1:

PR #560, opened 2015-07-21
DeploymentGroupHistoryWriter: Write actual bytes

rohansingh:
We were erroneously serializing all DeploymentGroupEvent objects to an
empty byte array. Instead, serialize correctly into JSON.


rohansingh:
@davidxia

PR #562, opened 2015-07-21
Swallow BadVersionException in rolling update

rohansingh:
All masters race to process rolling update steps, so some are going to get
a BadVersionException when attempting to save changes to ZooKeeper. This is
by design and not exceptional, so don't throw an exception on it.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
@@ -593,8 +593,13 @@ public class ZooKeeperMasterModel implements MasterModel {
           zkOperations.toArray(new ZooKeeperOperation[zkOperations.size()])
       ));
     } catch (final KeeperException e) {
-      throw new HeliosRuntimeException(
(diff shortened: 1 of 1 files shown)


davidxia:
:+1:

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -593,8 +593,13 @@ public void rollingUpdateStep(final DeploymentGroup deploymentGroup,
           zkOperations.toArray(new ZooKeeperOperation[zkOperations.size()])
       ));
     } catch (final KeeperException e) {
-      throw new HeliosRuntimeException(
-          "rolling-update on deployment-group " + deploymentGroup.getName() + " failed", e);
+      if ((e instanceof KeeperException.BadVersionException) && e.getPath().equals(statusPath)) {
```
davidxia: @rohansingh I just saw in the master logs that `e.getPath()` sometimes throws NPE because e.path is null. What should we do in that case?

rohansingh: Fixed in #568.

PR #563, opened 2015-07-21
Only send deployment events to Kafka not ZK

davidxia:
We don't need to send deployment group events to ZooKeeper.
Just send them to Kafka fire-and-forget style because this
data isn't critical. This vastly simplifies things.


davidxia:
@rohansingh

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -597,9 +597,7 @@ public void rollingUpdateStep(final DeploymentGroup deploymentGroup,
           "rolling-update on deployment-group " + deploymentGroup.getName() + " failed", e);
     }
 
-    if (deploymentGroupHistoryWriter != null) {
-      deploymentGroupHistoryWriter.saveHistoryItems(opsEvents.getEvents());
-    }
+    kafkaSender.send((KafkaRecord) opsEvents.getEvents());
```
rohansingh: Pseudo-code:
```
for (final X event : opsEvents.getEvents()) {
  kafkaSender.send(KafkaRecord.of(event.key, event.timestamp, event.toBytes()));
}
```

helios-services/src/main/java/com/spotify/helios/servicescommon/KafkaSender.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * A class that wraps {@link org.apache.kafka.clients.producer.KafkaProducer}.
+ */
+public class KafkaSender {
+
+  private static final Logger log = LoggerFactory.getLogger(KafkaSender.class);
+
+  private final Optional<KafkaProducer<String, byte[]>> kafkaProducer;
+
+  public KafkaSender(final Optional<KafkaProducer<String, byte[]>> kafkaProducer) {
```
gimaker: why optional?

gimaker: i see

gimaker:
Maybe pass an interface to ZKMasterModel to make it less coupled?

E.g. interface DeploymentEventSink { void emitEvent(DeploymentEvent ev); }

helios-services/src/main/java/com/spotify/helios/servicescommon/KafkaSender.java
```diff
@@ -0,0 +1,64 @@
...
+    this.kafkaProducer = kafkaProducer;
+  }
+
+  public void send(final KafkaRecord record) {
+    if (kafkaProducer.isPresent()) {
+      try {
+        final Future<RecordMetadata> future = kafkaProducer.get().send(
+            new ProducerRecord<String, byte[]>(record.getKafkaTopic(), record.getKafkaData()));
+        final RecordMetadata metadata = future.get(5, TimeUnit.SECONDS);
```
rohansingh: I would s/5/KAFKA_SEND_TIMEOUT/ or something similar, so it's not just a magic number in the middle of the class.

rohansingh:
This is sweet. Could we rip out the other Kafka stuff and use this for sending task history as well?

rohansingh:
Never mind, just saw #567 :)

:+1:

rohansingh:
:+1: 

:shipit:

PR #564, opened 2015-07-22
Refactor DeploymentGroupEvent

davidxia:
* Remove jobId as it's already in DeploymentGroup
* Add ability to set timestamp
* Add tests


davidxia:
@rohansingh @gimaker @rculbertson

rohansingh:
:+1:

PR #567, opened 2015-07-22
Remove Kafka functionality from QueuingHistoryWriter

davidxia:
Have the caller handle sending events to Kafka to make
QueuingHistoryWriter only be responsible for sending to
ZooKeeper. Simplifies things.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/TaskStatusEvent.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/TaskStatusEvent.java
@@ -39,7 +39,10 @@ import com.fasterxml.jackson.annotation.JsonProperty;
  * </pre>
  */
 @JsonIgnoreProperties(ignoreUnknown = true)
-public class TaskStatusEvent {
+public class TaskStatusEvent extends Descriptor {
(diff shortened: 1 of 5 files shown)


davidxia:
@rohansingh @gimaker 

I was thinking about how to use composition instead of inheritance here. We could have `TaskHistoryWriter` have `QueuingHistoryWriter` as an attribute. `QueuingHistoryWriter` wouldn't be abstract, and we'd remove the generic `TEvent` and just have it write arrays of bytes to ZK.

The abstract method `getKey()` would instead be implemented to return a checksum of the byte array. But what about `getTimestamp()` which `TaskHistoryWriter` implemented in a way to return the timestamp attribute of the `TaskStatusEvent`? If `QueuingHistoryWriter` naively wrote `byte[]` to ZK, it wouldn't be able to know the timestamp or ZK path.

We could have a `QueuingHistoryEvent` interface with `getTimestamp()` and `getZkEventsPath()`, but I'm not sure what the advantage of that is over the current design.

rohansingh:
Looks like you broke nginx.

rohansingh:
:+1:

PR #569, opened 2015-07-22
Reinstate DeploymentGroupResourceTests

davidxia:


davidxia:
@gimaker Should these tests be uncommented?

rohansingh:
:+1:

rohansingh:
I think they were commented out because they were broken.

PR #571, opened 2015-07-22
Make RolloutPlanner respect order of hosts

davidxia:
RolloutPlanner now respects the order of hosts determined by the
RollingUpdateService. This will make rolling-update deploy jobs to hosts in
alphanumeric order by their hostnames.


davidxia:
@gimaker @rohansingh @rculbertson

rohansingh:
:shipit:

PR #574, opened 2015-07-23
Get rid of ZK paths for deployment group history

rohansingh:
We're no longer writing deployment group history to ZooKeeper, just sending
it to Kafka.


rohansingh:
@davidxia

davidxia:
:+1:

PR #577, opened 2015-07-23
Only emit interesting deployment group events

rohansingh:
Don't emit events for internal state transitions that don't actually affect
the state of anything.


rohansingh:
@davidxia

PR #578, opened 2015-07-23
Fix mocking in TaskHistoryWriterTest

rohansingh:
It turns out you're not supposed to change the stubbing on Mockito mocks
once you start calling the mock from different threads. Doing so is unsafe
and can cause failures.

We were changing the stubbing on the mocked ZK client after using it in
TaskHistoryWriter, which caused weird exceptions and test failures. Fix
this so we do all the stubbing beforehand.


rohansingh:
@davidxia @gimaker

gimaker:
:+1:

PR #579, opened 2015-07-23
QueueingHistoryWriter: Get rid of mutex

rohansingh:
Added this mutex when refactoring QueueingHistoryWriter to prevent the
possibility of losing an occasional new event when we are trimming old
events, but it causes more problems than it's worth.


rohansingh:
@gimaker @davidxia

PR #580, opened 2015-07-23
Handle stopped containers during rolling-update

rculbertson:
If a container is stopped, undeploy it and deploy the new one.

Before we were not undeploying stopped containers for two reasons:
1. Stopping a job would clear out its deployment group name field.
2. If redeploying the same job, the undeploy step would see the
container is already deployed and skip it, even though it was not running.


davidxia:
:+1:

rohansingh:
I think even with this, the deployment group will still end up failed if you stop a job after it's been deployed but before the AWAIT_RUNNING step for the host succeeds, right?

I'm starting to like more and more the idea of just explicitly stopping the deployment group if you stop or undeploy any of its jobs.

PR #581, opened 2015-07-23
Remove unused throws

davidxia:


rculbertson:
:+1:

davidxia:
replaced by https://github.com/spotify/helios/pull/582

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -1597,8 +1596,7 @@ public Deployment undeployJob(final String host, final JobId jobId, final String
                                                        final String host,
                                                        final Deployment deployment,
                                                        final String token)
-      throws JobDoesNotExistException, JobAlreadyDeployedException, HostNotFoundException,
-             JobPortAllocationConflictException, TokenVerificationException {
+      throws JobDoesNotExistException, JobAlreadyDeployedException, TokenVerificationException {
```
davidxia: We aren't checking if a host exists here. We check in `getUndeployOperations()`. Should we check again here or wait for `rollingUpdateAwaitRunning()` to time out?

The chances of an agent going away between undeploy and await running are slim. But if it does happen, the error message in the deployment group status will be clearer than just `timed out while retrieving job status`.

rohansingh:
Confused. Is this PR supposed to be open or is it obsoleted?

davidxia:
It's open. Yea I changed the status a bunch.

PR #583, opened 2015-07-25
Check host exists in rolling-update

davidxia:


helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -1597,8 +1596,10 @@ public Deployment undeployJob(final String host, final JobId jobId, final String
                                                        final String host,
                                                        final Deployment deployment,
                                                        final String token)
-      throws JobDoesNotExistException, JobAlreadyDeployedException, HostNotFoundException,
```
davidxia: Removed `JobPortAllocationConflictException` since we don't check for it. Should we?

davidxia:
@rculbertson @rohansingh

PR #587, opened 2015-07-27
Add status filter to host list command

### LINKED ISSUE #469: Would it be helpful to add a status filter to helios hosts?
I often do `helios hosts | grep -v Down` to select only hosts that are "Up". Would it be helpful to have `helios hosts --status up`?

davidxia:
Fixes #469

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
+++ b/helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
@@ -441,15 +441,25 @@ public class HeliosClient implements AutoCloseable {
   }
 
   public ListenableFuture<HostStatus> hostStatus(final String host) {
-    return get(uri(path("/hosts/%s/status", host)), HostStatus.class);
+    return hostStatus(host, Collections.<String, String>emptyMap());
+  }
+
+  public ListenableFuture<HostStatus>
+  hostStatus(final String host, final Map<String, String> queryParams) {
+    return get(uri(path("/hosts/%s/status", host), queryParams), HostStatus.class);
   }
 
   public ListenableFuture<Map<String, HostStatus>> hostStatuses(final List<String> hosts) {
+    return hostStatuses(hosts, Collections.<String, String>emptyMap());
+  }
+
+  public ListenableFuture<Map<String, HostStatus>>
+  hostStatuses(final List<String> hosts, final Map<String, String> queryParams) {
     final ConvertResponseToPojo<Map<String, HostStatus>> converter = ConvertResponseToPojo.create(
         TypeFactory.defaultInstance().constructMapType(Map.class, String.class, HostStatus.class),
         ImmutableSet.of(HTTP_OK));
 
-    return transform(request(uri("/hosts/statuses"), "POST", hosts), converter);
+    return transform(request(uri("/hosts/statuses", queryParams), "POST", hosts), converter);
   }
 
   public ListenableFuture<Integer> registerHost(final String host, final String id) {
--- a/helios-services/src/main/java/com/spotify/helios/master/resources/HostsResource.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/resources/HostsResource.java
@@ -61,6 +61,7 @@ import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.Response;
 
+import static com.google.common.base.Strings.isNullOrEmpty;
 import static com.spotify.helios.common.protocol.JobUndeployResponse.Status.FORBIDDEN;
 import static com.spotify.helios.common.protocol.JobUndeployResponse.Status.HOST_NOT_FOUND;
 import static com.spotify.helios.common.protocol.JobUndeployResponse.Status.INVALID_ID;
@@ -140,6 +141,7 @@ public class HostsResource {
   /**
    * Returns various status information about the host.
    * @param host The host id.
+   * @param statusFilter An optional status filter.
    * @return The host status.
    */
   @GET
@@ -147,13 +149,22 @@ public class HostsResource {
   @Produces(APPLICATION_JSON)
   @Timed
   @ExceptionMetered
-  public Optional<HostStatus> hostStatus(@PathParam("id") final String host) {
-    return Optional.fromNullable(model.getHostStatus(host));
+  public Optional<HostStatus> hostStatus(
+      @PathParam("id") final String host,
+      @QueryParam("status") @DefaultValue("") final String statusFilter) {
+    final HostStatus status = model.getHostStatus(host);
+    if (status != null &&
+        (isNullOrEmpty(statusFilter) || statusFilter.equals(status.getStatus().toString()))) {
+      return Optional.of(status);
+    } else {
+      return Optional.absent();
+    }
   }
 
   /**
    * Returns various status information about the hosts.
    * @param hosts The hosts.
+   * @param statusFilter An optional status filter.
    * @return The response.
    */
   @POST
@@ -161,12 +172,16 @@ public class HostsResource {
   @Produces(APPLICATION_JSON)
   @Timed
   @ExceptionMetered
-  public Map<String, HostStatus> hostStatuses(final List<String> hosts) {
+  public Map<String, HostStatus> hostStatuses(
+      final List<String> hosts,
+      @QueryParam("status") @DefaultValue("") final String statusFilter) {
     final Map<String, HostStatus> statuses = Maps.newHashMap();
     for (final String current : hosts) {
       final HostStatus status = model.getHostStatus(current);
       if (status != null) {
-        statuses.put(current, status);
+        if (isNullOrEmpty(statusFilter) || statusFilter.equals(status.getStatus().toString())) {
+          statuses.put(current, status);
+        }
       }
     }
     return statuses;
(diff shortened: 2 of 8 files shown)


davidxia:
@gimaker

helios-tools/src/main/java/com/spotify/helios/cli/command/HostListCommand.java
```diff
@@ -147,11 +157,13 @@ int run(final Namespace options, final HeliosClient client, final PrintStream ou
       try {
         final Map<String, HostStatus> hostStatuses = client.hostStatuses(hosts).get();
         for (final Entry<String, HostStatus> entry : hostStatuses.entrySet()) {
-          statuses.put(entry.getKey(), Futures.immediateFuture(entry.getValue()));
+          if (statusFilter == null || entry.getValue().getStatus().equals(statusFilter)) {
```
rohansingh: Maybe just uber-paranoia but it becomes NPE-proof if you do `statusFilter.equals( ... )` instead.

davidxia: ah good point

helios-tools/src/main/java/com/spotify/helios/cli/command/HostListCommand.java
```diff
@@ -90,6 +91,10 @@ public HostListCommand(final Subparser parser) {
         .action(storeTrue())
         .help("Print full host names.");
 
+    statusArg = parser.addArgument("--status")
+        .nargs("?")
+        .help("Filter hosts by its status, i.e. \"UP\" or \"DOWN\".");
```
gimaker: You might want to checkout out the built-in choices() method. E.g.

    .choices(Collections2.transform(
        Arrays.asList(HostStatus.Status.values()), new Function<HostStatus.Status, String>() {
          @Override
          public String apply(final HostStatus.Status input) {
            return input.toString();
          }
        }).toArray(new String[0]));

(it's icky in java 7, much nicer in java 8)

helios-tools/src/main/java/com/spotify/helios/cli/command/HostListCommand.java
```diff
@@ -107,6 +112,11 @@ int run(final Namespace options, final HeliosClient client, final PrintStream ou
         .from(client.listHosts().get())
         .filter(containsPattern(pattern))
         .toList();
+    HostStatus.Status statusFilter = null;
+    try {
+      statusFilter = HostStatus.Status.valueOf(options.getString(statusArg.getDest()));
+    } catch (IllegalArgumentException | NullPointerException ignored) {
```
gimaker: erroring out here would probably be nice, and avoid confusion when you have typos.

E.g. "list --status up" or "list --status DWN" will list all hosts but not actually filter anything out

gimaker:
:+1: with a few nitpicks

helios-tools/src/main/java/com/spotify/helios/cli/command/HostListCommand.java
davidxia: Yea, quite gross

helios-tools/src/main/java/com/spotify/helios/cli/command/HostListCommand.java
davidxia: good point

helios-tools/src/main/java/com/spotify/helios/cli/command/HostListCommand.java
```diff
@@ -107,6 +125,16 @@ int run(final Namespace options, final HeliosClient client, final PrintStream ou
         .from(client.listHosts().get())
         .filter(containsPattern(pattern))
         .toList();
+    HostStatus.Status statusFilter = null;
+    final String statusFilterStr = options.getString(statusArg.getDest());
+    if (!isNullOrEmpty(statusFilterStr)) {
+      try {
+        statusFilter = HostStatus.Status.valueOf(statusFilterStr);
```
gimaker: statusFilterStr.toUpperCase() ?

gimaker:
:+1:

helios-tools/src/main/java/com/spotify/helios/cli/command/HostListCommand.java
davidxia: Yea I was wondering if the client should be smart like that. You are too nice to the user. I can change it.

davidxia: Ah yes. Because of the argparse `Choices`, if the user uses lowercase, an error will occur before we even get here.

davidxia:
Hm now I'm thinking this whole filtering thing should be done server-side.

gimaker:
:+1: 

No strong opinions on server/client side filtering for this case. That said, I'm tempted to move the label filtering to the server-side so that it works when --json is specified (or, alternatively, just do the label filtering on the client-side as well).

davidxia:
Yea I think server-side will have better performance and easier to maintain.

On Tue, Jul 28, 2015 at 12:53 PM, Staffan Gimåker <notifications@github.com>
wrote:

> [image: :+1:]
>
> No strong opinions on server/client side filtering for this case. That
> said, I'm tempted to move the label filtering to the server-side so that it
> works when --json is specified (or, alternatively, just do the label
> filtering on the client-side as well).
>
> —
> <https://github.com/spotify/helios/pull/587#issuecomment-125677685>.
>



-- 
David Xia
Software Engineer
dxia@spotify.com

davidxia:
@gimaker I moved the logic to the server.

helios-services/src/main/java/com/spotify/helios/master/resources/HostsResource.java
```diff
@@ -147,13 +148,21 @@ public HostDeregisterResponse delete(@PathParam("id") final String host) {
   @Produces(APPLICATION_JSON)
   @Timed
   @ExceptionMetered
-  public Optional<HostStatus> hostStatus(@PathParam("id") final String host) {
-    return Optional.fromNullable(model.getHostStatus(host));
+  public Optional<HostStatus> hostStatus(@PathParam("id") final String host,
+                                         @QueryParam("status") final String statusFilter) {
```
gimaker: What happens in the status query-param is missing? It should still work, to remain backwards-compatible with older versions of the CLI - does it?

If it doesn't work without specifying the query-param you might want to use `@DefaultValue("")` or [guava] Optional.

helios-services/src/main/java/com/spotify/helios/master/resources/HostsResource.java
```diff
@@ -161,12 +170,15 @@ public HostDeregisterResponse delete(@PathParam("id") final String host) {
   @Produces(APPLICATION_JSON)
   @Timed
   @ExceptionMetered
-  public Map<String, HostStatus> hostStatuses(final List<String> hosts) {
+  public Map<String, HostStatus> hostStatuses(final List<String> hosts,
+                                              @QueryParam("status") final String statusFilter) {
```
gimaker: Same thing here.

A test checking that it works to call this endpoint wihtout the status param would be nice.

helios-services/src/main/java/com/spotify/helios/master/resources/HostsResource.java
davidxia: It should work. Here's a test: https://github.com/spotify/helios/pull/587/files#diff-dc34e9cadd23a8d110d25e12c41041e8R81

davidxia: Just tested both manually too just to make sure.

helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
```diff
@@ -441,15 +441,25 @@ public Integer apply(final Response reply) {
   }
 
   public ListenableFuture<HostStatus> hostStatus(final String host) {
-    return get(uri(path("/hosts/%s/status", host)), HostStatus.class);
+    return hostStatus(host, Collections.<String, String>emptyMap());
+  }
+
+  public ListenableFuture<HostStatus>
+  hostStatus(final String host, final Map<String, String> queryParams) {
```
gimaker:
```
hostStatus(final String host, final Map<String, String> queryParams)
```

-->

```
hostStatus(final String host, @Nullable final String statusFilter)
```

or similar ?

gimaker: or Optional instead of Nullable

davidxia: I made it a map in case we want to add more query parameters in the future. And `uri()` takes a map anyways.

davidxia:
@rohansingh @gimaker There are probably a few more cases where we can optimize the cli commands and move logic into the server side.

davidxia:
@gimaker How does it look now?

helios-services/src/main/java/com/spotify/helios/master/resources/HostsResource.java
davidxia: I added the `@DefaultValue("")` since it's more explicit.

helios-services/src/main/java/com/spotify/helios/master/resources/HostsResource.java
davidxia: I added the `@DefaultValue("")` since it's more explicit.

PR #588, opened 2015-07-27
KafkaSender: Change failures from ERROR to WARN

rohansingh:
It's not really an exceptional error if there's an occasional issue sending
an event to Kafka. Avoid causing alerts based on this.


rohansingh:
@davidxia

davidxia:
:+1:

PR #589, opened 2015-07-27
Add more data to DeploymentGroupEvent

davidxia:
Add DeploymentGroupStatus and domain for more info


davidxia:
@rohansingh

gimaker:
Why is the additional information needed?

Emitting the full DeploymentGroupStatus blob makes it harder to change how deployment groups works internally. I.e. if we change DeploymentGroupStatus that will also break anything that depend on the format of the emitted events. But maybe that's okay?

davidxia:
Yea, it's a lot more info. We wanted the error message from DeploymentGroupStatus and were lazy...

gimaker:
:+1:

davidxia:
Thanks. Let's merge for now and change later as needed.

PR #591, opened 2015-07-28
Don't throw exception if Kafka broker not found

rohansingh:
Right now the agent won't start if DNS lookup for a Kafka broker fails.
Instead, just log the exception and move on.


davidxia:
:+1:

gimaker:
:+1:

PR #595, opened 2015-07-29
Show endpoints to which HeliosClient connects

davidxia:


davidxia:
This is useful in debugging our users' failed tempjobs. I often want to see where their client is connecting. This won't print to console when you run `helios ...`.

davidxia:
@gimaker ^

gimaker:
:+1:

PR #596, opened 2015-07-29
Make client support https endpoints

### LINKED ISSUE #491: adding https encryption to the client
The help function tells me, that the Master endpoints must be of the form "http[s]://heliosmaster.domain.net:<port>".
Using Helios Master behind a https encrypted nginx throws an exception. i looks to me like the helios client is ignoring the https and always using http.

davidxia:
Fixes #491


rohansingh:
:+1:

PR #597, opened 2015-07-30
Fail softly when agent cannot save task history

davidxia:
Bad data can get into the agent's task history backing store.
Fail softly to not screw up the important Helios job operations.


davidxia:
@rohansingh @gimaker

PR #598, opened 2015-07-30
Add logging around ZooKeeper cluster ID check

rohansingh:
We have issues where nodes don't recover nicely after losing connectivity
to ZooKeeper due to some issue with the cluster ID check. Add some logging
so we can see what's going on.


davidxia:
:+1:

PR #600, opened 2015-07-30
Emit event on rolling-update

davidxia:


davidxia:
@rohansingh

gimaker:
how do you tell this DeploymentGroupEvent (DGE) apart from other DGEs?

I.e. how do I tell this DGE apart from a DGE emitted when a rollout is started due to more hosts being added to the deployment group?

PR #603, opened 2015-08-01
Create test method to start many default masters

davidxia:
Will be useful for integration tests involving many masters.


davidxia:
@gimaker @rculbertson

rculbertson:
:+1:

PR #613, opened 2015-08-06
Set default query param values for master endpoints

davidxia:
Add tests for bad HTTP requests


davidxia:
inspired by @gimaker here https://github.com/spotify/helios/pull/587#discussion_r35781309

PR #614, opened 2015-08-07
Add more rolling update reactor threads

davidxia:
This is a temporary measure to speed up rolling updates
while we come up with a more scalable solution.


davidxia:
@rohansingh @gimaker Can it be as simple as this?

gimaker:
That should work, albeit a little ugly (would be nicer to have a single underlying ExecutorService with multiple threads).

Also, with 10 concurrent threads there's a chance they might contend with each other (e.g. two threads doing the same operations, and then one failing to commit the results). Not sure it will be a problem but you could do something like randomize the order in which deployments are processed to reduce any potential problem (or add a delay).

PR #615, opened 2015-08-07
Parallelize execution of rolling updates

rohansingh:
Have a thread pool of 5 threads which we use to run the rolling update step
for each deployment group. Hopefully this speeds things up a bit.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdateService.java
+++ b/helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdateService.java
@@ -26,6 +26,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.util.concurrent.AbstractIdleService;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 
 import com.spotify.helios.common.descriptors.DeploymentGroup;
 import com.spotify.helios.common.descriptors.HostSelector;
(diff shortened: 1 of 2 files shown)


davidxia:
assuming tests pass, :+1:

rohansingh:
@gimaker Want to take a look as well? I think you know Java stuff better than either of us.

helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdateService.java
```diff
@@ -132,12 +145,17 @@ public void run(final boolean timeout) throws InterruptedException {
     @Override
     public void run(final boolean timeout) throws InterruptedException {
       for (final DeploymentGroup dg : masterModel.getDeploymentGroups().values()) {
-        try {
-          masterModel.rollingUpdateStep(dg, DefaultRolloutPlanner.of(dg));
-        } catch (Exception e) {
-          log.warn("error processing rolling update step for deployment group: {} - {}",
-                   dg.getName(), e);
-        }
+        rollingUpdateExecutor.execute(new Runnable() {
```
gimaker: probably very much a non-issue but this looks like an unbounded queue

helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdateService.java
```diff
@@ -60,6 +65,14 @@
   private final Reactor hostUpdateReactor;
   private final Reactor rollingUpdateReactor;
 
+  private final ExecutorService rollingUpdateExecutor = MoreExecutors.getExitingExecutorService(
```
gimaker: The exiting executor service seems unnecessary

PR #623, opened 2015-08-13
Set kafka metadata fetch timeout to 5 seconds

davidxia:
instead of the default 1 minute. If kafka is borked,
the minute timeout will make rolling-update timeout.


davidxia:
@rculbertson

rculbertson:
:+1:

Issue #599, opened 2015-07-30
deployment-group-status shows DOWN hosts


gimaker:
...and it's confusing. It should probably either filter out DOWN hosts, or show the host status as well.

In the example below the non-up-to-date host is actually DOWN but there's no obvious way to tell:

```
$ helios -d <X> deployment-group-status megatest
Name: megatest
Job Id: megatest-service:0.0.1-SNAPSHOT-e7e221e:6d6153c
Status: ACTIVE
Host selectors:
  role = heliosagent

HOST                                    UP-TO-DATE    JOB                                                STATE      
xxxx-heliosagent-a1.xxxx                X             megatest-service:0.0.1-SNAPSHOT-e7e221e:6d6153c    RUNNING    
xxxx-heliosagent-a1-v1438210071.xxxx                  -                                                  -          
xxxx-heliosagent-staffan1.xxxx          X             megatest-service:0.0.1-SNAPSHOT-e7e221e:6d6153c    RUNNING
```

davidxia:
This is a slightly related question. Should `helios status --job jobID` show hosts that are DOWN?

```
JOB ID                                HOST                                       GOAL     STATE            CONTAINER ID    PORTS
helioscanary:0.8.0-74b6afb:3d335f4    host1              START    RUNNING          b4d570e         hm=20000:20000 http=20001:20001
```

`helios hosts host1` will show it as down.

```
HOST                             STATUS        DEPLOYED    RUNNING    CPUS    MEM     LOAD AVG    MEM USAGE    OS                         HELIOS     DOCKER          LABELS
host1    Down 1 day    1           1          1       3 gb    0.96        0.87         Linux 3.13.0-34-generic    0.8.447    1.6.0 (1.18)    helioscanary=true, role=heliosagent
```

davidxia:
fixed by #630 and #624

PR #624, opened 2015-08-19
Fix total number of deployment group hosts

### LINKED ISSUE #620: rolling-update misreports total number of hosts
The `rolling-update` command misreports the number of hosts in the `-> RUNNING (n/m)` output. `m` includes `DOWN` hosts. It shouldn't.

davidxia:
Only include UP hosts in rolling-update and deployment-group-status
output, i.e. don't include DOWN hosts.

Fixes #620 #599

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/master/resources/DeploymentGroupResource.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/resources/DeploymentGroupResource.java
@@ -192,7 +192,7 @@ public class DeploymentGroupResource {
         JobId deployedJobId = null;
         TaskStatus.State state = null;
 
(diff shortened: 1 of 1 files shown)


davidxia:
@staffan

gimaker:
I'd prefer to filter things in the status endpoint, e.g. add this check (status == UP) to DeploymentGroupResource.java:195. Or even return the DOWN hosts but include the status in the response, so that you can filter on the client side.

With this change, hosts going DOWN temporarily will trigger rolling updates (since the hosts are changed) which seems a bit counter-intuitive.

davidxia:
Ah I see. Good catch.

gimaker:
:+1:

PR #625, opened 2015-08-19
Don't abort rollingUpdateStep if a deployment group fails

gimaker:
Just log it instead.


gimaker:
@davidxia

davidxia:
:+1:

davidxia:
Should we log in places like this? https://github.com/spotify/helios/blob/master/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java#L1664

gimaker:
yeah we should catch and log those exceptions in rollingUpdateStep()

let me add that to the PR

davidxia:
:+1:

PR #627, opened 2015-08-19
Emit event when rolling-update is started because hosts were added

gimaker:


gimaker:
Not completely sure how to test this in a reasonable manner -- wokring on that part...

gimaker:
@davidxia

davidxia:
:+1:

PR #629, opened 2015-08-20
Deployment groups: Always take ownership of manually deployed jobs

gimaker:
If the deployment group's target job is already deployed on the host,
assume ownership if the job is exactly the same and as not deployed
by a deployment group, even if the --migrate flag was not specified.


gimaker:
@davidxia @matslina 

Not sure I like this. With this in place there are more edge cases. I.e.

A deployment will only undeploy jobs previously deployed by it, unless:
 * the job is exactly the same and was manually deployed, OR
 * the job is exactly the same and --migrate was specified

PR #631, opened 2015-08-23
Add failure threshold option to deployment groups

davidxia:


davidxia:
@staffan @rohansingh

helios-client/src/main/java/com/spotify/helios/common/descriptors/RolloutOptions.java
```diff
@@ -31,20 +31,24 @@
...
   private final boolean migrate;
   private final boolean overlap;
+  private final int failureThreshold;
 
   public RolloutOptions(@JsonProperty("timeout") final long timeout,
                         @JsonProperty("parallelism") final int parallelism,
                         @JsonProperty("migrate") final boolean migrate,
-                        @JsonProperty("overlap") boolean overlap) {
+                        @JsonProperty("overlap") boolean overlap,
+                        @JsonProperty("failureThreshold") final int failureThreshold) {
```
gimaker: for existing deployment groups this field will be missing. what will happen for them?

will failureThreshold be set to 0?

davidxia: Ah thanks. We use @Nullable and `    this.failureThreshold = Optional.fromNullable(failureThreshold).or(0);` because `@DefaultValue()` only takes String.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -733,13 +740,23 @@ private RollingUpdateOp rollingUpdateAwaitRunning(final ZooKeeperClient client,
       // then manually undeployed. The job will not get redeployed, so treat this as a failure.
       final Deployment deployment = getDeployment(host, deploymentGroup.getJobId());
       if (deployment == null) {
-        return opFactory.error(
-            "Job unexpectedly undeployed. Perhaps it was manually undeployed?", host);
+        opFactory.failTask();
```
gimaker: Baking this logic into RollingUpdateOpFactory this way seems a bit out of place and conflated, but I see that you need to emit an event AND move to the next task. Looks like it's time for some refactoring of RollingUpdateOpFactory...

helios-client/src/main/java/com/spotify/helios/common/descriptors/RolloutOptions.java
```diff
@@ -31,20 +35,25 @@
 
   public static final long DEFAULT_TIMEOUT = TimeUnit.MINUTES.toSeconds(5);
   public static final int DEFAULT_PARALLELISM = 1;
+  public static final int DEFAULT_FAILURE_THRESHOLD = 1;
```
gimaker: add comment or rename to make it clear that this is a percentage value?

helios-client/src/main/java/com/spotify/helios/common/descriptors/RolloutOptions.java
```diff
@@ -31,20 +35,25 @@
...
-                        @JsonProperty("migrate") final boolean migrate,
-                        @JsonProperty("overlap") boolean overlap) {
+  private final int failureThreshold;
+
+  public RolloutOptions(
+      @JsonProperty("timeout") final long timeout,
+      @JsonProperty("parallelism") final int parallelism,
+      @JsonProperty("migrate") final boolean migrate,
+      @JsonProperty("overlap") boolean overlap,
+      @JsonProperty("failureThreshold") @Nullable final Integer failureThreshold) {
```
gimaker: Why `@Nullable Integer` and not just an int? Using `@DefaultValue` to set a default seems nicer, alternatively you can let it be 0 for existing deployment groups. This should be safe since we do a greater-than comparison (as opposed to a >=). x > 0 is only true if x is non-zero, which means one host failed -- which is the current behaviour.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -793,19 +800,21 @@ private RollingUpdateOp rollingUpdateDeploy(final ZooKeeperClient client,
                                               final RollingUpdateOpFactory opFactory,
                                               final DeploymentGroup deploymentGroup,
                                               final String host) {
+    final List<ZooKeeperOperation> operations = Lists.newArrayList();
```
gimaker: Looks like spurious changes with no changed behaviour in this file. Necessary?

helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdateOpFactory.java
```diff
@@ -78,10 +83,18 @@ public RollingUpdateOp nextTask(final List<ZooKeeperOperation> operations) {
       // Emit an event signalling that we're DONE!
       events.add(eventFactory.rollingUpdateDone(deploymentGroup));
     } else {
+      final DeploymentGroupTasks.Builder taskBuilder = tasks.toBuilder()
```
gimaker: in the above ^ if block a task failed event should be emitted if `failedTarget` is non-null, right?

helios-client/src/main/java/com/spotify/helios/common/descriptors/RolloutOptions.java
davidxia: Good point.

helios-client/src/main/java/com/spotify/helios/common/descriptors/RolloutOptions.java
davidxia: `@DefaultValue()` only takes String and seems like it's used for a different purpose.

```
 * Defines the default value of request metadata that is bound using one of the 
 * following annotations:
 * {@link javax.ws.rs.PathParam}, 
 * {@link javax.ws.rs.QueryParam}, 
 * {@link javax.ws.rs.MatrixParam},
 * {@link javax.ws.rs.CookieParam},
 * {@link javax.ws.rs.FormParam},
 * or {@link javax.ws.rs.HeaderParam}.
 * The default value is used if the corresponding metadata is not present in the
 * request.
 ```

File: helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java (hunk 5 above)
Comment (davidxia) [inline helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java]: My IntelliJ does this stuff automatically :(

Comment (davidxia) [inline helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java]: I thought it made it easier to read. But I can revert.

File: helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdateOpFactory.java (hunk 6 above)
Comment (davidxia) [inline helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdateOpFactory.java]: yes yes

**Comment by davidxia** [comment]
Still need to update RollingUpdateOpFactoryTest

**Comment by davidxia** [comment]
Added tests.

File: helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdateOpFactory.java [hunk 7]
```diff
@@ -103,41 +120,49 @@ public RollingUpdateOp yield() {
...
+  public RollingUpdateOp taskError(final Exception e, final String host,
+                                   final RollingUpdateError errorCode) {
+    return taskError(e.getMessage(), host, errorCode);
   }
 
-  public RollingUpdateOp error(final Exception e, final String host,
-                               final RollingUpdateError errorCode) {
-    return error(e.getMessage(), host, errorCode);
+  private boolean isOverFailureThreshold() {
+    return ((float) (tasks.getFailedTargets().size() + 1) / tasks.getNumTargets() * 100)
```
rohansingh: Maybe move the failed target count to be a param, so that if somebody calls `isOverFailureThreshold` in the future they aren't confused why it always counts at least one host as failed.

helios-client/src/main/java/com/spotify/helios/common/descriptors/DeploymentGroupTasks.java
```diff
@@ -22,42 +22,57 @@
...
   private final DeploymentGroup deploymentGroup;
+  private final Set<String> failedTargets;
+  private final int numTargets;
 
   private DeploymentGroupTasks(
       @JsonProperty("rolloutTasks") final List<RolloutTask> rolloutTasks,
       @JsonProperty("taskIndex") final int taskIndex,
-      @JsonProperty("deploymentGroup") final DeploymentGroup deploymentGroup) {
+      @JsonProperty("deploymentGroup") final DeploymentGroup deploymentGroup,
+      @JsonProperty("numTargets") final int numTargets,
```
rohansingh: Have `numTargets` be an `Integer` and verify that it's not null.

helios-tools/src/main/java/com/spotify/helios/cli/command/RollingUpdateCommand.java
```diff
@@ -141,10 +147,13 @@ protected int runWithJobId(final Namespace options, final HeliosClient client,
     final long rolloutTimeout = options.getLong(rolloutTimeoutArg.getDest());
     final boolean migrate = options.getBoolean(migrateArg.getDest());
     final boolean overlap = options.getBoolean(overlapArg.getDest());
+    final int failureThreshold = options.getInt(failureThresholdArg.getDest());
 
     checkArgument(timeout > 0, "Timeout must be greater than 0");
     checkArgument(parallelism > 0, "Parallelism must be greater than 0");
     checkArgument(rolloutTimeout > 0, "Rollout timeout must be greater than 0");
+    checkArgument(failureThreshold > 0 && failureThreshold < 100,
```
rohansingh: Change semantics so that 0 and 100 are allowed values, with 0 meaning any failures cause the rolling update to fail.

helios-client/src/main/java/com/spotify/helios/common/descriptors/DeploymentGroupTasks.java
davidxia: I'll set the builder's `numTargets` explicitly to `private int numTargets = DEFAULT_FAILURE_THRESHOLD_PERCENTAGE;` below.

helios-tools/src/main/java/com/spotify/helios/cli/command/RollingUpdateCommand.java
davidxia: Done.

davidxia:
@rohansingh Updated according to your comments. Also changed the failure threshold to be a float instead of int.

helios-client/src/main/java/com/spotify/helios/common/descriptors/DeploymentGroupTasks.java
gimaker: But it will be null (for existing deployment groups that are currently being rolled out). What do you do if it's null?

gimaker: Note: It being 0 is fine -- it just means that any host failing will cause the DG to fail.

davidxia:
@gimaker Added `testTaskErrorTransitionToFailed()` and `testTaskErrorTransitionToDone()` to `RollingUpdateOpFactoryTest`.

helios-client/src/main/java/com/spotify/helios/common/descriptors/DeploymentGroupTasks.java
davidxia: For existing deployment group data in ZK, I think we can do:

`      @JsonProperty("numTargets") @Nullable final Integer numTargets,`
and
`    this.numTargets = Optional.fromNullable(numTargets).or(0);`

gimaker: The result is the same. I have no strong opinions on how to get there.

davidxia:
I tweaked `isOverFailureThreshold()` so that a rollout fails when 1 out of 2 hosts fails with a 50% failure threshold. Before you had to set a 51% threshold.

gimaker:
Can we hold off on merging this? I suspect there will be ramifications elsewhere, and would like to look into those before merging anything. Specifically:

 * How will rolling-update behave if a host fails, but the rollout does not? I suspect the failed host will not be shown at all in the output.
 * "deployment-group-status" should probably show the failure threshold, and be explicit about which host failed.

etc.

davidxia:
Replaced by https://github.com/spotify/helios/pull/650

PR #633, opened 2015-08-25
Go back to old implementation of TaskHistoryWriter

rohansingh:
This implementation actually worked, whereas the `helios history` command
no longer works after our changes.


rohansingh:
@davidxia

rohansingh:
Basically grabbed the old implementation out of the log, renamed it, and ripped the Kafka out of it again.

PR #634, opened 2015-08-25
Add previous job states to deployment group errors

davidxia:
For debugging failed rollouts.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
@@ -21,6 +21,7 @@
 
 package com.spotify.helios.master;
 
+import com.google.common.base.Function;
 import com.google.common.base.Optional;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
@@ -85,6 +86,7 @@ import java.util.UUID;
 import static com.google.common.base.Charsets.UTF_8;
 import static com.google.common.base.Optional.fromNullable;
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Strings.isNullOrEmpty;
 import static com.google.common.collect.Lists.newArrayList;
 import static com.google.common.collect.Lists.reverse;
 import static com.spotify.helios.common.descriptors.DeploymentGroupStatus.State.DONE;
@@ -99,6 +101,7 @@ import static com.spotify.helios.servicescommon.coordination.ZooKeeperOperations
 import static com.spotify.helios.servicescommon.coordination.ZooKeeperOperations.set;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.emptyMap;
+import static java.util.Collections.singletonList;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 /**
@@ -361,18 +364,30 @@ public class ZooKeeperMasterModel implements MasterModel {
   }
 
   /**
-   * Given a jobId, returns the N most recent events in it's history in the cluster.
+   * Given a jobId, returns the N most recent events in its history in the cluster.
    */
   @Override
   public List<TaskStatusEvent> getJobHistory(final JobId jobId) throws JobDoesNotExistException {
+    return getJobHistory(jobId, null);
+  }
+
+  /**
+   * Given a jobId and host, returns the N most recent events in its history on that host in the
+   * cluster.
+   */
+  @Override
+  public List<TaskStatusEvent> getJobHistory(final JobId jobId, final String host)
+      throws JobDoesNotExistException {
     final Job descriptor = getJob(jobId);
     if (descriptor == null) {
       throw new JobDoesNotExistException(jobId);
     }
     final ZooKeeperClient client = provider.get("getJobHistory");
     final List<String> hosts;
+
     try {
-      hosts = client.getChildren(Paths.historyJobHosts(jobId));
+      hosts = (!isNullOrEmpty(host)) ? singletonList(host) :
+              client.getChildren(Paths.historyJobHosts(jobId));
     } catch (NoNodeException e) {
       return emptyList();
     } catch (KeeperException e) {
@@ -381,10 +396,10 @@ public class ZooKeeperMasterModel implements MasterModel {
 
     final List<TaskStatusEvent> jsEvents = Lists.newArrayList();
 
-    for (String host : hosts) {
+    for (String h : hosts) {
       final List<String> events;
       try {
-        events = client.getChildren(Paths.historyJobHostEvents(jobId, host));
+        events = client.getChildren(Paths.historyJobHostEvents(jobId, h));
       } catch (NoNodeException e) {
         continue;
       } catch (KeeperException e) {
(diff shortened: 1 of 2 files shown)


davidxia:
@gimaker @rohansingh

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -1705,4 +1735,26 @@ private static void verifyToken(final String token, final Job job)
     }
   }
 
+  private String getPreviousJobStates(final JobId jobId, final String host) {
```
davidxia: This seems ugly. And which class should such a function live (this class is already too big).

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -744,14 +767,19 @@ private RollingUpdateOp rollingUpdateAwaitRunning(final ZooKeeperClient client,
...
 
       if (isRolloutTimedOut(client, deploymentGroup)) {
         // time exceeding the configured deploy timeout has passed, and this job is still not
         // running
-        return opFactory.error("timed out waiting for job to reach RUNNING", host);
+        String errMsg = String.format("timed out waiting for job %s to reach RUNNING.", jobId);
+
+        final TaskStatus taskStatus = getTaskStatus(client, host, jobId);
+        final String currentState = taskStatus == null ? "" : taskStatus.getState().toString();
+        final String previousStates = getPreviousJobStates(jobId, host);
```
gimaker: how many? should we limit these more?

davidxia: I'll limit to 10

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -360,18 +364,36 @@ public void addJob(final Job job) throws JobExistsException {
...
     if (descriptor == null) {
       throw new JobDoesNotExistException(jobId);
     }
     final ZooKeeperClient client = provider.get("getJobHistory");
-    final List<String> hosts;
+    List<String> hosts;
+
     try {
       hosts = client.getChildren(Paths.historyJobHosts(jobId));
+      if (!isNullOrEmpty(host)) {
```
rohansingh: Instead of having this conditional, how about something like:

```java
hosts = (!isNullOrEmpty(host)) ? singletonList(host) : client.getChildren(...);
```

That is to say, there's no need to do the ZK operation if a host is provided. We can just trust that it's an actual host for the job. If it isn't, then it will get filtered out later when we loop through the hosts list.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -360,18 +364,36 @@ public void addJob(final Job job) throws JobExistsException {
...
+  @Override
+  public List<TaskStatusEvent> getJobHistory(final JobId jobId, final String host)
+      throws JobDoesNotExistException {
     final Job descriptor = getJob(jobId);
     if (descriptor == null) {
       throw new JobDoesNotExistException(jobId);
     }
     final ZooKeeperClient client = provider.get("getJobHistory");
-    final List<String> hosts;
+    List<String> hosts;
```
rohansingh: This can become `final` again if you decide to use the approach in the next comment.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -744,14 +767,20 @@ private RollingUpdateOp rollingUpdateAwaitRunning(final ZooKeeperClient client,
...
-    } else if (!taskStatuses.get(deploymentGroup.getJobId()).getState()
-        .equals(TaskStatus.State.RUNNING)) {
+    } else if (!taskStatuses.get(jobId).getState().equals(TaskStatus.State.RUNNING)) {
       // job isn't running yet
 
       if (isRolloutTimedOut(client, deploymentGroup)) {
         // time exceeding the configured deploy timeout has passed, and this job is still not
         // running
-        return opFactory.error("timed out waiting for job to reach RUNNING", host);
+        String errMsg = String.format("timed out waiting for job %s to reach RUNNING.", jobId);
```
rohansingh: Any reason to not just defer this until we have the previous states and construct the error message all in one go?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -1705,4 +1734,27 @@ private static void verifyToken(final String token, final Job job)
     }
   }
 
+  private List<TaskStatus.State> getPreviousJobStates(final JobId jobId, final String host) {
+    List<TaskStatus.State> previousStates;
```
rohansingh: Can this be `final` since its assignment is the last statement in the `try` below? Not sure. Java... how does it work?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -1705,4 +1734,27 @@ private static void verifyToken(final String token, final Job job)
...
+    try {
+      final List<TaskStatusEvent> jobHistory = getJobHistory(jobId, host);
+      jobHistory.subList(0, Math.min(10, jobHistory.size()));
```
rohansingh: Maybe parameterize the function with `int maxStates` or something like that and use that here instead of just hardcoding `10`.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -1705,4 +1734,27 @@ private static void verifyToken(final String token, final Job job)
...
+      Function<TaskStatusEvent, TaskStatus.State> statusesToStrings =
+          new Function<TaskStatusEvent, TaskStatus.State>() {
+            @javax.annotation.Nullable
```
rohansingh: Eh? Should this be here? Also why is it `@javax.annotation.Nullable` instead of just `@Nullable`?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
davidxia: Good idea.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
davidxia: IntelliJ added it for me. Idk why. I'll remove.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
davidxia: Seems reasonable, but IntelliJ complains that `previousStates = emptyList();` in the catch might have already been assigned to.

rohansingh:
:+1:

PR #635, opened 2015-08-25
Tweak deployment groups metrics

gimaker:
* Always include the full deployment group in all events.
 * For event that either fail or succeed, indicate the result via a
   numeric "success" field.
 * Include a "machine readable" error code for failed tasks.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/rollingupdate/DeploymentGroupEventFactory.java
+++ b/helios-services/src/main/java/com/spotify/helios/rollingupdate/DeploymentGroupEventFactory.java
@@ -97,7 +97,7 @@ public class DeploymentGroupEventFactory {
                                                  final Map<String, Object> failEvent) {
     final Map<String, Object> ev = createEvent("rollingUpdateFinished", deploymentGroup);
     ev.put("success", 0);
-    ev.put("failEvent", failEvent);
+    ev.put("failedTask", failEvent);
     return ev;
   }
 }
--- a/helios-services/src/main/java/com/spotify/helios/rollingupdate/DeploymentGroupEventFactory.java
+++ b/helios-services/src/main/java/com/spotify/helios/rollingupdate/DeploymentGroupEventFactory.java
@@ -26,6 +26,7 @@ import com.google.common.collect.Maps;
 import com.spotify.helios.common.descriptors.DeploymentGroup;
 import com.spotify.helios.common.descriptors.RolloutTask;
 
+import java.util.Collections;
 import java.util.Map;
 
 public class DeploymentGroupEventFactory {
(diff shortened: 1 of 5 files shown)


gimaker:
@rohansingh @davidxia

rohansingh:
`helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java:812:12: WhitespaceAround: 'catch' is not followed by whitespace.`

helios-services/src/main/java/com/spotify/helios/rollingupdate/DeploymentGroupEventFactory.java
```diff
@@ -37,51 +35,58 @@
...
     return ev;
   }
 
   public Map<String, Object> rollingUpdateDone(final DeploymentGroup deploymentGroup) {
-    final Map<String, Object> ev = Maps.newHashMap();
-    ev.put("eventType", "rollingUpdateDone");
-    ev.put("timestamp", System.currentTimeMillis());
-    ev.put("deploymentGroupName", deploymentGroup.getName());
-    ev.put("state", DeploymentGroupStatus.State.DONE);
+    final Map<String, Object> ev = createEvent("rollingUpdateFinished", deploymentGroup);
```
davidxia: Can we make the method names and eventType the same?

gimaker:
@matslina

helios-services/src/main/java/com/spotify/helios/rollingupdate/DeploymentGroupEventFactory.java
davidxia: Oh nvm/

gimaker: No. The eventType is rollingUpdatedFinished both for rollingUpdateDone and rollingUpdateFailed

helios-services/src/main/java/com/spotify/helios/rollingupdate/DeploymentGroupEventFactory.java
```diff
@@ -37,51 +36,68 @@
     HOSTS_CHANGED
   }
 
-  public Map<String, Object> rollingUpdateTaskSucceeded(final DeploymentGroup deploymentGroup,
-                                                        final RolloutTask task) {
+  private Map<String, Object> createEvent(final String eventType,
```
davidxia: Is it worth it to create classes for these objects?

gimaker: For what objects?

A DeploymentGroupEvent class? Probably. But I'd rather do that in a separate commit.

Or specific classes for specific events?

davidxia: Yea a separate commit sounds good. Something like the below?

DeploymentGroupEvent class:
* eventType
* timestamp
* deploymentGroup
* success
* reason
* failedTask

DeploymentGroupTaskEvent class inherits from DeploymentGroupTask and also includes:
* action
* target
* error
* errorCode
* metadata

davidxia:
:+1:

helios-services/src/main/java/com/spotify/helios/rollingupdate/DeploymentGroupEventFactory.java
gimaker: Not sure exactly how it'd look (not sure a class hierachy makes sense, especially since things like failedTask make no sense what-so-ever for rollout-task events), but passing in a metadata `Map<String, Object>` is surely kludgy.

PR #638, opened 2015-08-27
Fix bug in the logic for determining if a rolling-update is timed out

gimaker:
Previously, we'd update the deployment-group status on every task and
as a consequence we looked at the mtime of the deployment-group status
node to determine how long ago a task started. With the recent changes
to how deployment-groups work, i.e. introducing the separate tasks node,
this is no longer the case.

So, instead look at the mtime of the deployment-group tasks path --
which we write to whenever we proceeded to a new task.


gimaker:
@davidxia @rohansingh

davidxia:
:+1: Test?

gimaker:
@davidxia yes please, patches welcome ;)

PR #639, opened 2015-08-28
Emit done event on rolling-update with no hosts

davidxia:


davidxia:
@staffan @rohansingh

gimaker:
:-1: The rolling-update op factory today contains only logic for acting on and moving between tasks in a rollout. Adding logic to it where around how to start rolling-updates seems out of place -- for example, the `start()` method cannot be called once a rolling-update is started, and `nextTask()` and `error()` cannot be called unless it has already been started.

IMO, a better, but still not that nice, solution would be to just return a `RollingUpdateOp` from ` getInitRollingUpdateOps` (it doesn't need to be created by the factory).

Although, at that point maybe it should be renamed to `ZkOpsAndEventsPair` or something like that :)

davidxia:
So I believe you can call `start()` in the middle of a rolling-update. It would be when the user runs `rolling-update` twice, the second time before the first has finished.

`nextTask()` and `error()` will probably throw exceptions if called before `start()`, but the `ZooKeeperMasterModel` that was calling `RollingUpdateOpFactory` needs to know this now anyways.

It's probably not the best way, but I think it's better than now for these reasons:

1. Fixes a bug
2. Consolidates more ZK ops and event data into one place where it's less likely we'll forget to pair ops and events
3. Increases test coverage of the business logic around starting a rolling-update

rohansingh:
:+1:

PR #643, opened 2015-09-01
Make deployment groups support insecure access tokens

davidxia:
`helios rolling-update` now takes a `--token` switch which allows it to
undeploy and deploy jobs with tokens.


davidxia:
@rohansingh

rohansingh:
Looks great.

:+1:

PR #644, opened 2015-09-01

davidxia:


helios-system-tests/src/main/java/com/spotify/helios/system/DeregisterTest.java
```diff
@@ -102,4 +103,27 @@ public void test() throws Exception {
...
+    awaitHostRegistered(client, host, LONG_WAIT_SECONDS, SECONDS);
+    awaitHostStatus(client, host, UP, LONG_WAIT_SECONDS, SECONDS);
+
+    // Kill off agent
+    agent.stopAsync().awaitTerminated();
+    awaitHostStatus(client, host, DOWN, LONG_WAIT_SECONDS, SECONDS);
+
+    // Start a new agent with the same hostname but have it generate a different ID
+    resetAgentStateDir();
+    startDefaultAgent(testHost(), "--zk-registration-ttl", "0");
```
davidxia: Couldn't find an easy way to set the ttl to a positive number but not slow down the test by passing in a fake `Clock` object via CLI args here. See https://github.com/spotify/helios/pull/644/files#diff-dce8417dd28c57a529199badc738ddf3R80

PR #648, opened 2015-09-03
Fix NPE in deployment-group-status command

davidxia:
Fix NPE when user runs `deployment-group-status` before
running `rolling-update` for a deployment group.


davidxia:
@rohansingh @gimaker

rohansingh:
:+1:

PR #649, opened 2015-09-03
Default to EMPTY_TOKEN in RolloutOptions

davidxia:


davidxia:
@rohansingh

davidxia:
We seem to follow this pattern elsewhere.

davidxia:
See https://github.com/spotify/helios/blob/master/helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java#L201

PR #650, opened 2015-09-03
Add failure threshold option to deployment groups

davidxia:


davidxia:
@rohansingh @gimaker

davidxia:
Even though helioscanary is the only one that needs this right now, I feel that's an important enough piece of infra that's so blocked by deployment group failures that this feature will be valuable for that.

gimaker:
:-1: I think saying to the user that everything went fine when the service is in partially rolled out state is a recipe for confusion.

PR #654, opened 2015-09-08

davidxia:


helios-system-tests/src/main/java/com/spotify/helios/system/DeregisterTest.java
```diff
@@ -102,4 +103,27 @@ public void test() throws Exception {
...
+    awaitHostRegistered(client, host, LONG_WAIT_SECONDS, SECONDS);
+    awaitHostStatus(client, host, UP, LONG_WAIT_SECONDS, SECONDS);
+
+    // Kill off agent
+    agent.stopAsync().awaitTerminated();
+    awaitHostStatus(client, host, DOWN, LONG_WAIT_SECONDS, SECONDS);
+
+    // Start a new agent with the same hostname but have it generate a different ID
+    resetAgentStateDir();
+    startDefaultAgent(testHost(), "--zk-registration-ttl", "0");
```
davidxia: Couldn't find an easy way to set the ttl to a positive number but not slow down the test by passing in a fake Clock object via CLI args here. See https://github.com/spotify/helios/pull/654/files#diff-dce8417dd28c57a529199badc738ddf3R71

helios-services/src/main/java/com/spotify/helios/agent/HostInfoReporter.java
```diff
@@ -93,6 +96,10 @@ protected void runOneIteration() throws InterruptedException {
         .build();
 
     nodeUpdater.update(hostInfo.toJsonBytes());
+
+    // Touch the host ID node in ZK so that later on AgentZooKeeperRegistrar can know how long a
+    // host has been down.
+    //hostIdUpdater.touch();
```
davidxia: Adding this breaks the tests. @gimaker @rohansingh @mattnworb maybe you could take a second look with me when you guys get a chance.

helios-services/src/main/java/com/spotify/helios/agent/AgentConfig.java
```diff
@@ -43,6 +43,7 @@
   private int zooKeeperConnectionTimeoutMillis;
   private String zooKeeperNamespace;
   private String zooKeeperClusterId;
+  private int setZooKeeperRegistrationTtlMinutes;
```
mattnworb: I think the `set` prefix in the field name is confusing (or accidental?)

helios-services/src/main/java/com/spotify/helios/agent/AgentZooKeeperRegistrar.java
```diff
@@ -53,13 +56,19 @@
...
 
-  public AgentZooKeeperRegistrar(final Service agentService, final String name, final String id) {
+  public AgentZooKeeperRegistrar(final Service agentService, final String name, final String id,
+                                 final int zooKeeperRegistrationTtlMinutes) {
     this.agentService = agentService;
     this.name = name;
     this.id = id;
+    this.zooKeeperRegistrationTtlMillis =
+        TimeUnit.MILLISECONDS.convert(zooKeeperRegistrationTtlMinutes, TimeUnit.MINUTES);
+    this.clock = new SystemClock();
```
mattnworb: I think it would help your test to have the Clock as a constructor arg

davidxia: Yea, but I have to figure out a way to pass a fake clock via the CLI args. That isn't straight forward :(

gimaker: Can you not test `AgentZooKeeperRegistrar` in isolation, without CLI and crap?

At a glance, `agentService` seems easily mockable (only `stopAsync()` is called). `name` and `id` should be as well.

mattnworb:
This approach would seem to assume that the clocks on agents are the same, wouldn't it? Maybe this is totally implausible but it seems like if there was a `host1` healthy and alive and registered and I started up a duplicate `host1` with the clock set 15 minutes in the future, the latter would deregister the former.

You may not even need to add the `touch()` method to NodeUpdater as you could check the mtime of the  host agent info path, right?

davidxia:
Starting a helios host whose clock is ahead and has the same hostname as an already registered agent, will have the same effect as a user deregistering a perfectly OK host. I'm not sure we should worry about this case.

mattnworb:
> the same effect as a user deregistering a perfectly OK host.

Good point 😀

helios-services/src/main/java/com/spotify/helios/agent/AgentZooKeeperRegistrar.java
davidxia: Yea, let me try that out.

davidxia: Actually I already have an end-to-end test for this in `DeregisterTest.testRegistrationResolution()`

davidxia:
@mattnworb @gimaker Updated with another commit.

mattnworb:
@davidxia stat-ing the `hostInfoPath` looks a lot cleaner :+1:

mattnworb:
re: b970a89, makes sense to me. I think ideally it wouldn't be necessary to signal between those components because it seems like the other two should not be started until the former has registered at least once, but this looks like a good way to solve the immediate problem.

rohansingh:
:+1: 

Looks good.

davidxia:
@rohansingh Do you know why we start LabelReporter and EnvironmentVariableReporter before knowing if the agent registered in ZK successfully? https://github.com/spotify/helios/pull/654/files#diff-144948f2deffedddb6b1bda1077673a9R351

rohansingh:
I think the assumption was that when `zkRegistrar.startAsync().awaitRunning()` returns, registration should be complete.

davidxia:
@rohansingh Should we make label and env var reporter wait on zk registration with the countdown latch or use another way?

PR #655, opened 2015-09-09
Try to get HTTPS SRV records first, fallback to HTTP

davidxia:


helios-client/src/main/java/com/spotify/helios/common/Resolver.java
```diff
@@ -44,7 +44,8 @@
 
 public abstract class Resolver {
 
-  private static final String SRV_FORMAT = env("HELIOS_SRV_FORMAT", "_%s._http.%s");
```
davidxia: Making a new env var to not break users, if any, that were using `HELIOS_SRV_FORMAT`.

davidxia:
@rohansingh @gimaker Is it worth it to add a test for this? I'd have to refactor `Resolver.java` to be able to mock SRV lookups.

davidxia:
@gimaker @rohansingh @mattnworb

mattnworb:
If you are adding the findbugs-annotations `<dependency>` just to satisfy maven-enforcer, you can put that in the `<dependencyManagement>` section instead (possibly just in the root pom). That way if we ever remove the dns dependency then the findbugs-annotation one doesn't linger on the classpath.

mattnworb:
Do you think it would be useful to somehow configure it to _not_ fallback to http, for anyone who wants to enforce only https communication?

davidxia:
@mattnworb Yea I was wondering about the best default.

1. default to https fallback to http
2. default to http and user opts into https
3. default to https and user opts out of it
...

mattnworb:
On second thought I suppose that any users who wanted their masters to be available only over https or only over http could just not register those DNS records, so maybe the client doesn't need to care about flags or options.

davidxia:
@mattnworb Yea, we can just leave it up to the operators of helios masters to decide. If they only want clients to talk https, they'd make masters only listen on localhost. If they only want http, they just don't setup SRV records.

davidxia:
@mattnworb Thanks, tests pass now.

helios-tools/pom.xml
```diff
@@ -138,6 +138,12 @@
                 <include>com/sun/jna/**</include>
               </includes>
             </filter>
+            <filter>
+              <artifact>commons-logging</artifact>
+              <includes>
+                <include>org/apache/commons/logging/**</include>
```
mattnworb: should comment/document why this is here too

mattnworb:
+1 if you added something to the docs about the new A record resolution logic.

davidxia:
@mattnworb Good points. I added comments.

mattnworb:
thanks, now here is your +1

davidxia:
Thanks, I'll wait for @gimaker and/or @rohansingh to take a look too.

helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
```diff
@@ -344,20 +366,34 @@ private HttpURLConnection connect(final URI uri, final String method, final byte
...
+        @Override
+        public boolean verify(String ip, SSLSession sslSession) {
+          final String tHostname = hostname.endsWith(".") ?
+                                   hostname.substring(0, hostname.length() - 1) : hostname;
+          return new DefaultHostnameVerifier().verify(tHostname, sslSession);
+        }
+      });
+    }
+
     connection.setRequestProperty("Accept-Encoding", "gzip");
```
rohansingh: Probably should add:

    System.setProperty("sun.net.http.allowRestrictedHeaders", "true");
    connection.setRequestProperty("Host", hostname);

For the case where Helios is behind nginx or some other proxy that serves multiple hostnames on the same IP.

Another way to approach this rather than creating an IP URI, etc., is to just set the request proxy to the IP:

    Proxy proxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress("10.0.0.1", 8080));
    conn = new URL(urlString).openConnection(proxy);

That might actually simplify the code quite a bit, though I'm not sure if it would work with HTTPS.

davidxia: Did the first. Maybe your second point would work? `Proxy.Type` only has HTTP, SOCKS, and DIRECT.

helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
```diff
@@ -312,13 +316,30 @@ private HttpURLConnection connect(final URI uri, final String method, final byte
...
-        final String scheme = endpoint.getScheme();
-        final String host = endpoint.getHost();
-        final int port = endpoint.getPort();
+      // Resolve hostname into IPs so client will round-robin and retry for multiple A records.
+      // Keep a mapping of IPs to hostnames for TLS verification.
+      final List<URI> ipEndpoints = Lists.newArrayList();
+      final Map<URI, URI> ipToHostnameUris = Maps.newHashMap();
+
+      for (final URI hnUri : endpoints) {
+        final InetAddress[] ips = InetAddress.getAllByName(hnUri.getHost());
```
gimaker: Catch `UnknownHostException` here and emit a warning?

helios-tools/src/main/java/com/spotify/helios/cli/Utils.java
```diff
@@ -63,6 +63,7 @@ public static HeliosClient getClient(final Target target, final PrintStream err,
     try {
       endpoints = target.getEndpointSupplier().get();
     } catch (Exception ignore) {
+      err.println(ignore.getMessage());
```
gimaker: intentional?

helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
rohansingh: `Proxy.Type.HTTP` includes both HTTP and HTTPS proxies. The master endpoint will work fine as a "proxy" for HTTP, but not sure if it will forward HTTPS correctly. The nice thing about doing it this way would be that you could probably dump the hostname verification stuff.

But it might also not work at all. Up to you if you want to investigate or just punt on it.

I think long-term the best solution is to use the Apache HTTP client that lets you override DNS resolution and have <hostname> explicitly resolve to the single IP that we've selected.

helios-tools/src/main/java/com/spotify/helios/cli/Utils.java
davidxia: Must've left this by accident.

davidxia:
@rohansingh I think I am setting the right SNI now.

rohansingh:
:+1:

PR #662, opened 2015-09-22
Agent checks port availability when starting job

davidxia:
Have the agent check if a port is available from the perspective of the
host. Before this the agent was only checking if the port hadn't been allocated by itself.

This is racy, but better than nothing.


davidxia:
@gimaker @rohansingh

davidxia:
@danielnorberg

gimaker:
:+1:

rohansingh:
:+1:

PR #664, opened 2015-09-24
Refactor ZK master + agent registration

davidxia:
Move redundant code into ZooKeeperRegistrarUtil.java

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/servicescommon/ZooKeeperRegistrarUtil.java
+++ b/helios-services/src/main/java/com/spotify/helios/servicescommon/ZooKeeperRegistrarUtil.java
@@ -1,0 +1,329 @@
+/*
+ * Copyright (c) 2014 Spotify AB.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.helios.servicescommon;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.spotify.helios.common.HeliosRuntimeException;
+import com.spotify.helios.common.Json;
+import com.spotify.helios.common.descriptors.AgentInfo;
+import com.spotify.helios.common.descriptors.Deployment;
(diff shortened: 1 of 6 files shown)


davidxia:
@gimaker @rohansingh

helios-services/src/main/java/com/spotify/helios/servicescommon/ZooKeeperRegistrarUtil.java
```diff
@@ -0,0 +1,329 @@
...
+                               : Collections.<JobId>emptyList();
+      throw new HostStillInUseException(host, jobs);
+    } catch (NoNodeException e) {
+      throw new HostNotFoundException(host);
+    } catch (KeeperException e) {
+      throw new HeliosRuntimeException(e);
+    }
+  }
+
+  public static List<JobId> listHostJobs(final ZooKeeperClient client, final String host) {
```
rohansingh: Should be `private` (maybe)?

helios-services/src/test/java/com/spotify/helios/servicescommon/ZooKeeperRegistrarUtilTest.java
```diff
@@ -0,0 +1,57 @@
...
+    final String idPath = Paths.configHostId(HOSTNAME);
+    ZooKeeperRegistrarUtil.registerHost(zkClient, idPath, HOSTNAME, ID);
+    verify(zkClient).ensurePath(Paths.configHost(HOSTNAME));
+    verify(zkClient).ensurePath(Paths.configHostJobs(HOSTNAME));
+    verify(zkClient).ensurePath(Paths.configHostPorts(HOSTNAME));
+    verify(zkClient).ensurePath(Paths.statusHost(HOSTNAME));
+    verify(zkClient).ensurePath(Paths.statusHostJobs(HOSTNAME));
+    verify(zkClient).createAndSetData(idPath, ID.getBytes(UTF_8));
+  }
+
```
rohansingh: Maybe have a deregister test?

rohansingh:
:+1: other than inline comments

helios-services/src/test/java/com/spotify/helios/servicescommon/ZooKeeperRegistrarUtilTest.java
```diff
@@ -0,0 +1,129 @@
...
+    ops.add(new Delete(Paths.configHostPorts(HOSTNAME)));
+
+    // Remove host id
+    verify(zkClient).exists(Paths.configHostId(HOSTNAME));
+    ops.add(new Delete(Paths.configHostId(HOSTNAME)));
+
+    // Remove host config root
+    ops.add(new Delete(Paths.configHost(HOSTNAME)));
+
+    verify(zkClient).transaction(ops);
```
davidxia: @rohansingh Should we check that `transaction()` was called with these particular ops or just `verify` that the call was made? Test is brittle because if implementation changes, it will fail.

rohansingh: Yeah, it is really tied to the implementation right now. If anything changes, either the verification or the mock will break. Perhaps just verifying that *something* happened is OK.

Another approach would be to use an `org.apache.curator.test.TestingServer` to actually bring up an in-memory ZK, write some nodes into it, and then ensure the nodes are gone after running `deregisterHost`. Though not sure that this is the right place to do that kind of test (maybe?).

davidxia: I'm going to go with the first approach. We have ITs for deregistering in `DeregisterTest`.

rohansingh:
:+1:

PR #665, opened 2015-09-30
Support for creating/deploying jobs with image digests

rohansingh:
This updates the JobValidator to accept image names that have digests, and
adds some unit tests and an integration test to make sure it all works.


rohansingh:
@davidxia @gimaker

davidxia:
:+1:

PR #667, opened 2015-10-01
add a 'metadata' field to Jobs

mattnworb:
The "metadata" field allows users to optionally store an arbitrary list of
key-value pairs to be associated with the Job.

The Helios master or agent doesn't act on the contents of the Job metadata,
but the field can be used for storing information alongside the Job
definition such as Git Commit ID, etc.

The (optional) list of metadata, as key-value pairs, can be sent in the
JobCreateCommand with the `-m` or `--metadata` switch, like:

```
$ helios create -m a=1 -m b=2 myjob:1 image
```

The format follows the same as the `--env` flag, in that the value is
required to be key-values delimited with `=`.

I don't have very strong opinions on if this is called "metadata" or "labels"
or "tags" or something else, although I do think "metadata" helps imply that
the data is just informational and won't be acted on (unlike "labels"/"tags").

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
@@ -508,7 +508,7 @@ public class Job extends Descriptor implements Comparable<Job> {
         this.token = EMPTY_TOKEN;
         this.healthCheck = EMPTY_HEALTH_CHECK;
         this.securityOpt = EMPTY_SECURITY_OPT;
-        this.metadata = EMPTY_METADATA;
+        this.metadata = Maps.newHashMap();
       }
 
       private Parameters(final Parameters p) {
@@ -655,7 +655,7 @@ public class Job extends Descriptor implements Comparable<Job> {
     }
 
     public Builder setMetadata(final Map<String, String> metadata) {
-      p.metadata = metadata;
+      p.metadata = Maps.newHashMap(metadata);
       return this;
     }
 
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
@@ -68,6 +68,9 @@ import static java.util.Collections.emptyMap;
  *   "id" : "myservice:0.5:3539b7bc2235d53f79e6e8511942bbeaa8816265",
  *   "image" : "myregistry:80/janedoe/myservice:0.5-98c6ff4",
  *   "hostname": "myhost",
+ *   "metadata": {
+ *     "foo": "bar
+ *   },
  *   "networkMode" : "bridge",
  *   "ports" : {
  *     "http" : {
@@ -116,6 +119,7 @@ public class Job extends Descriptor implements Comparable<Job> {
   public static final List<String> EMPTY_SECURITY_OPT = emptyList();
   public static final String DEFAULT_NETWORK_MODE = "bridge";
   public static final String EMPTY_HOSTNAME = null;
+  public static final Map<String, String> EMPTY_METADATA = emptyMap();
 
   private final JobId id;
   private final String image;
@@ -134,6 +138,7 @@ public class Job extends Descriptor implements Comparable<Job> {
   private final HealthCheck healthCheck;
   private final List<String> securityOpt;
   private final String networkMode;
+  private final Map<String, String> metadata;
 
   /**
    * Create a Job.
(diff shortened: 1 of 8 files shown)


mattnworb:

helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
```diff
@@ -633,6 +654,16 @@ public Builder setNetworkMode(final String networkMode) {
       return this;
     }
 
+    public Builder setMetadata(final Map<String, String> metadata) {
+      p.metadata = metadata;
+      return this;
+    }
+
+    public Builder addMetadata(String name, String value) {
```
gimaker: Calling `addMetadata()` will throw an exception unless you called `setMetadata()` (with a mutable map) first, because `p.metadata` defaults to `emptyMap()` which is immutable.

`metadata` in the builder should probably default to `Maps.newHashMap()`.
`setMetadata()` should make a copy of the input map.

gimaker: ...so that cases where you call `setMetadata` with an immutable map and then call `addMetadata` will still work.

helios-tools/src/main/java/com/spotify/helios/cli/command/JobCreateCommand.java
```diff
@@ -523,6 +532,27 @@ int run(final Namespace options, final HeliosClient client, final PrintStream ou
     }
   }
 
+  private static Map<String, String> parseListOfPairs(List<String> list, String fieldName) {
```
gimaker: There's a `argToStringMap` method in `com.spotify.helios.cli.Utils` that looks similar, AFAICT. Would be nice to kill off one of them maybe. (I'm not very fond of `argToStringMap`).

helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
mattnworb: good catch! fixed in 024eb58

helios-tools/src/main/java/com/spotify/helios/cli/command/JobCreateCommand.java
mattnworb: The difference with argToStringMap is that it is used for Arguments that have `nargs("+")` like `--labels` in the `host` command.

I don't totally get the nuances of the nargs part in argparse4j but I modeled this one after the `--env` command, meaning the input is expected to be like

```
create -m a=1 -m b=2 -m c=3 job:version image args...
```

with `nargs("+")` here it would have just one `-m` like

```
create -m a=1 b=2 c=3 job:version image args...
```

I'm not sure if that would be a problem with all of the other arguments in JobCreateCommand though, none of them have `nargs("+")` (even though they often have multiple values like `--env` and `-p` or `--volume`) except for `args` at the end which has `nargs("*")`

gimaker:
:+1:

mattnworb:
I am wondering if it would make more sense to use [Docker's built-in support for labels for images/containers](http://rancher.com/docker-labels/) rather than adding a metadata field for Jobs in Helios. It would seem to duplicate the same idea.

I think for the use case we are interested in at Spotify (git commit id of the image) an image label would suffice just fine

gimaker:
@mattnworb: We could tweak the docker-maven-plugin to emit such a label and use that, but if you don't use docker-maven-plugin or if you supply your own Dockerfile you'd have to add and maintain that line by yourself -- which seems error-prone and tedious.

helios-tools/src/main/java/com/spotify/helios/cli/command/JobCreateCommand.java
```diff
@@ -523,6 +532,27 @@ int run(final Namespace options, final HeliosClient client, final PrintStream ou
...
+    return parseListOfPairs(list, fieldName, "=");
+  }
+
+  private static Map<String, String> parseListOfPairs(List<String> list, String fieldName,
```
davidxia: Consider using `final`s here. We seem to like that.

helios-tools/src/main/java/com/spotify/helios/cli/command/JobCreateCommand.java
```diff
@@ -523,6 +532,27 @@ int run(final Namespace options, final HeliosClient client, final PrintStream ou
...
+                                                      String delimiter) {
+    Map<String, String> pairs = new HashMap<>();
+    for (final String s : list) {
+      final String[] parts = s.split("=", 2);
```
davidxia: s/=/delimiter/

helios-tools/src/main/java/com/spotify/helios/cli/command/JobCreateCommand.java
mattnworb: good catch, updated

helios-tools/src/main/java/com/spotify/helios/cli/command/JobCreateCommand.java
mattnworb: very good catch, updated

mattnworb:
After more discussion @gimaker and I decided to keep the addition of job metadata rather than using image labels, as it might be convenient to set metadata of a job independent of how the image was built (or if deploying a job for an image you did not build yourself, etc.).

6921b14 adds a `GIT_COMMIT_ID` entry to the job metadata by default if it is set as an environment variable when the job is created.

gimaker:
:+1:

PR #668, opened 2015-10-05
cosmetic changes to user manual / All Job Options

mattnworb:
reformatted the `All Helios Job Options` section of the manual to change
the markup from a list of bullet points to a sub-heading for each job
field. The intent here is to make the text on the page less dense, for
example when trying to view/understand the format of the port mapping or
service registration sections (including the many examples and defaults)
which is rendered as a big blob of text in HTML.

A nice benefit of having headers for each field is that it becomes
possible to link to each section on Github as well.

### DIFF
--- a/docs/user_manual.md
+++ b/docs/user_manual.md
@@ -129,7 +129,7 @@ example that uses all the available configuration keys with an explanation of ea
 
 ### All Helios Job Options
 
-```
+```json
 {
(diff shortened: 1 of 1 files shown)


mattnworb:
change is best viewed in preview mode 

davidxia:
:+1:

gimaker:
:+1:

PR #669, opened 2015-10-05
Include ThrottleState in deployment group metrics

gimaker:
... when a deplyoment group fails because the job failed to reach
RUNNING before the configured time out. This is useful to distinguish
between likely infrastructural issues (e.g. IMAGE_MISSING) and likely
"user errors" (e.g. FLAPPING).


gimaker:

mattnworb:
+1

or just log the whole taskStatus?

gimaker:
@mattnworb Aye, we could. I don't have a strong opinion either way. TaskStatus is quite hefty though, as it contains a reference to the `Job`.

davidxia:
:+1:

PR #670, opened 2015-10-05
Add Codota semantic code search to index

drorweiss:
For example, see results for HeliosClient.close() in http://www.codota.com/xref/#/github_spotify_helios_5612e5f19b224403007e2fd6/findUsages?query=com.spotify.helios.client.HeliosClient&method=Lcom%252Fspotify%252Fhelios%252Fclient%252FHeliosClient;close()V

Would appreciate feedback


davidxia:
Thanks, @drorweiss. I see you're the CEO of codota. :) I'm wondering if there are Helios users that want this.

drorweiss:
Hi @davidxia. That's right, I'm Codota's CEO :) . 
Yes, I think that users who read the Helios code, and try to learn from it will find Codota useful in figuring out the code. I believe it's a compelling code browsing experience.

Would be great if you try it and tell us what you think (possible also via email - dror at codota dot com)

davidxia:
Thanks, we'll try to check it out.

davidxia:
I'm going to close for now. We can revisit this if there are users who want the tool.

PR #671, opened 2015-10-06
fix broken JobCreateCommandTest in Jenkins

mattnworb:
The logic in JobCreateCommand for setting job metadata based on the
environment variables present in the JVM has the side effect of breaking
if the unit test is run by Jenkins which is passing the `GIT_COMMIT`
value to the test itself.

Since it is not simple to mock out a call to `System.getenv()`, refactor
the "get the environment variables" logic to use a `Supplier` which can
be passed in in the JobCreateCommandTest.

This test was broken in #667.


mattnworb:
@davidxia @gimaker

gimaker:
:+1:

PR #672, opened 2015-10-06
additional fixes for job metadata / GIT_COMMIT

mattnworb:
This fixes failures in the system tests when run under Jenkins (or any
agent with the `GIT_COMMIT` environment variable set) that are caused by
the job being created with the environment variable automatically
included but the test does a deep-equality assertion against an expected
Job value.

The hacky fix here is to modify `assertJobEquals()` to remove any
automatically-included-metadata from being compared in the assertion.

This is a continuation of #671

### DIFF
--- a/helios-system-tests/src/main/java/com/spotify/helios/system/SystemTestBase.java
+++ b/helios-system-tests/src/main/java/com/spotify/helios/system/SystemTestBase.java
@@ -58,12 +58,14 @@ import com.spotify.helios.ZooKeeperTestManager;
 import com.spotify.helios.ZooKeeperTestingServerManager;
 import com.spotify.helios.agent.AgentMain;
 import com.spotify.helios.cli.CliMain;
+import com.spotify.helios.cli.command.JobCreateCommand;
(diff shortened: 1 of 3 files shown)


mattnworb:
@gimaker @davidxia

helios-system-tests/src/main/java/com/spotify/helios/system/SystemTestBase.java
```diff
@@ -1094,7 +1096,29 @@ public Object call() throws Exception {
   }
 
   protected void assertJobEquals(final Job expected, final Job actual) {
-    assertEquals(expected.toBuilder().setHash(actual.getId().getHash()).build(), actual);
+    final Builder expectedBuilder = expected.toBuilder();
+
+    // hack to make sure that any environment variables that were folded into the created job
+    // because of environment variables set at runtime on the test-running-agent are removed
+    // from the actual when we assert the equality below
```
mattnworb: allow me to preemptively apologize for this

davidxia:
ugly but I guess it gets the job done?

PR #673, opened 2015-10-06
Add created Date field to Job

davidxia:
So we know when a job was created.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
@@ -54,6 +54,7 @@ import static java.util.Collections.emptyMap;
  * An sample expression of it in JSON might be:
  * <pre>
  * {
+ *   "created" : 1410308461448,
  *   "command" : [ "server", "serverconfig.yaml" ],
  *   "env" : {
  *     "JVM_ARGS" : "-Ddw.feature.randomFeatureFlagEnabled=true"
@@ -106,6 +107,7 @@ public class Job extends Descriptor implements Comparable<Job> {
   public static final Map<String, String> EMPTY_ENV = emptyMap();
   public static final Resources EMPTY_RESOURCES = null;
   public static final Map<String, PortMapping> EMPTY_PORTS = emptyMap();
+  public static final Long EMPTY_CREATED = null;
   public static final List<String> EMPTY_COMMAND = emptyList();
   public static final Map<ServiceEndpoint, ServicePorts> EMPTY_REGISTRATION = emptyMap();
   public static final Integer EMPTY_GRACE_PERIOD = null;
@@ -124,6 +126,7 @@ public class Job extends Descriptor implements Comparable<Job> {
   private final JobId id;
   private final String image;
   private final String hostname;
+  private final Long created;
   private final List<String> command;
   private final Map<String, String> env;
   private final Resources resources;
@@ -146,6 +149,8 @@ public class Job extends Descriptor implements Comparable<Job> {
    * @param id The id of the job.
    * @param image The docker image to use.
    * @param hostname The hostname to pass to the container.
+   * @param created The timestamp in milliseconds of when the job was created.
+   *                This should only be set by the server.
    * @param command The command to pass to the container.
    * @param env Environment variables to set
    * @param resources Resource specification for the container.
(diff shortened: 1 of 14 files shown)


helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
```diff
@@ -190,6 +192,7 @@ public Job(@JsonProperty("id") final JobId id,
              @JsonProperty("metadata") @Nullable final Map<String, String> metadata) {
     this.id = id;
     this.image = image;
+    this.created = new Date();
```
gimaker: maybe use a millisecond timestamp to be consistent with the timestamps reported in the task status history?

mattnworb:
I suggest using joda-time's `DateTime` instead, which would allow you to specify the timezone you want the value to be in, as `Date` can have unexpected timezone logic.

Datetime is also easier to test against

helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
```diff
@@ -246,6 +250,10 @@ public String getHostname() {
     return hostname;
   }
 
+  public Date getCreated() {
+    return created;
+  }
```
mattnworb: if Job is meant to be immutable, then this method breaks that, as Date can be altered

davidxia: Isn't this a getter?

davidxia:
I considered using Joda time and milliseconds. I used Date to be consistent with the expires field in Job class. DateTime seems more human readable than milliseconds. But if we do that itd be nice to also change expires which might be tricky to be backwards compatible. 

I'm also fine with keeping it as Date. Thoughts?

helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
gimaker: `Date` is itself not immutable (you can call e.g. `setHour()` on `Date`), so even if the reference remains unchanged this effectively makes `Job` mutable.

davidxia:
I made `created` optional and to be set by the caller. Otherwise the tests become a pain to fix.

mattnworb:
A few other thoughts on this: 

1. I'm not sure if we really care about the JSON format of a job but I think that `"Tue Oct 06 16:02:46 EDT 2015"` is a really ugly format for dates. I think it would be simpler to store a date as milliseconds since epoch, *UTC*, and relying on `Date.toString()` (if that is what is happening here) means that you might get timezone ids in the string based on where the user created the job from, which feels irrelevant. If we don't store as milliseconds then ISO-8601 in utc time still feels better.

2. Can the master assigned the created time to the job instead of generating it on the CLI side? The latter gives us the time on the CLI host, whereas the time on the master feels more authoritative.

davidxia:
@mattnworb Good points, I'll do milliseconds and have the server set it. But we'll need to ignore date when generating the job hash right?

helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
```diff
@@ -480,6 +498,7 @@ public Builder(final String hash, final Parameters parameters) {
       public String version;
       public String image;
       public String hostname;
+      public Long created;
```
rohansingh: Also add something to clear this out in `withoutMetaParameters` below. That keeps it from affecting the hash.

rohansingh:
Milliseconds since epoch is also what we use for the timestamp in task status events. :+1: for consistency.

helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
davidxia: Thanks. I was wondering whether it belonged there.

mattnworb:
+1 on the changes in 91262d6, although it feels like it should use the `Clock` abstraction instead of `System.currentTimeMillis()` (but this feels very minor)

davidxia:
@mattnworb yea, it might make testing easier in the future. See latest commit

mattnworb:
I was assuming that the thing that created the JobsResource had a `Clock` instance it would pass in, but :+1: either way

gimaker:
:+1:

Issue #610, opened 2015-08-06
Creating the id hash when using the rest interface


cldmnky:
We are working on a Helios Saltstack module using the Helios rest interface. Most of the requests are straight forward, but the POST to /jobs/ requires a sha1 pregenerated id hash included in the JSON job.

What would be the flow for creating the hash?

cldmnky:
We decided to use the cli interface in the module which worked out well, so this is kind of solved on our side :-)

davidxia:
@cldmnky Glad to hear you found a workaround. FYI clients can now create jobs without specifying a hash. See https://github.com/spotify/helios/pull/678

PR #678, opened 2015-10-07
master: allow for hash-less job create

### LINKED ISSUE #677: Remove client-side hash creation when creating a job
Idea: remove the requirement to specify the job hash when the Job JSON blob is submitted to the master. It may be hard for non-Java clients to create the job hash in the same exact way that the helios-client Java code does (which is dependent on JSON serialization settings in Jackson).

Additionally, the current requirement that the server is able to create the same hash from the Job arguments and have it match what the client sends means it is impossible to have the client send along new fields in the Job that the master does not yet know about. This means it's not really possible to deploy a new version of the client/CLI ahead of upgrading the master.

related to https://github.com/spotify/helios/pull/378

rohansingh:
* Validate job after generating the hash, if not present.
* Add regression test.

Fixes #677, #610.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
@@ -71,6 +71,16 @@ public class JobValidator {
 
   public static final List<String> VALID_NETWORK_MODES = ImmutableList.of("bridge", "host");
 
+  private final boolean shouldValidateJobHash;
+
+  public JobValidator() {
+    this(true);
+  }
+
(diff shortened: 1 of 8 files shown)


mattnworb:
:+1: :100:

davidxia:
Client will fail here https://github.com/spotify/helios/blob/master/helios-client/src/main/java/com/spotify/helios/common/JobValidator.java#L263 as there's no hash.

davidxia:
`helios inspect` with data from shared.cloud works. `helios create` with old clients works.

PR #679, opened 2015-10-07
Delete job history when a job is removed

gimaker:
There's no way to access job history once the job is removed. Keeping
it around seems of little value. Deleting the history with the job also
keeps the ZK data set size from growing endlessly over time.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
@@ -1056,6 +1056,15 @@ public class ZooKeeperMasterModel implements MasterModel {
       throw new HeliosRuntimeException("removing job " + id + " failed", e);
     }
 
+    // Delete job history on a best effort basis
+    try {
+      client.deleteRecursive(Paths.historyJob(id));
+    } catch (NoNodeException ignored) {
+      // There's no history for this job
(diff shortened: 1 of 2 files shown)


gimaker:

davidxia:
:100:

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -1056,6 +1056,15 @@ public Job removeJob(final JobId id, final String token)
       throw new HeliosRuntimeException("removing job " + id + " failed", e);
     }
 
+    // Delete job history on a best effort basis
+    try {
+      client.deleteRecursive(Paths.historyJob(id));
```
mattnworb: just curious - is it not possible to have the recursive delete added to the transaction above? or just not desired in this case?

mattnworb:
:+1:

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
gimaker: It's possible, but there's no recursive delete in ZK so first have to do a recursive list outside the transaction and then add a delete for each node to the transaction. This seemed like more work - and the recursive delete after the fact seemed good enough so I opted for being lazy.

PR #681, opened 2015-10-08
fix failure in CliConfigTest if HELIOS_MASTER is set on host

mattnworb:
If you happen to run CliConfigTest#testConfigFromFile() on a host that
has the environment variable `HELIOS_MASTER` set, the test will fail as
it implicitly assumes no environment variables are set. To be safe when
running the test, always wipe out whatever env vars are inherited by the
process.

### DIFF
--- a/helios-tools/src/test/java/com/spotify/helios/cli/CliConfigTest.java
+++ b/helios-tools/src/test/java/com/spotify/helios/cli/CliConfigTest.java
@@ -7,6 +7,7 @@ import com.google.common.collect.ImmutableMap;
 import com.typesafe.config.ConfigException;
 
 import org.hamcrest.Matchers;
+import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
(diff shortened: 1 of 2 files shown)


mattnworb:

note the test failure seems to happen on master branch as well

gimaker:
:+1:

Although I have to say that the general practice in that file seems very smelly. static variables don't get reset between tests, or even different test classes -- changing a static variable could potentially affect other tests that depend on it, and make it hard to figure out why things blow up.

davidxia:
@gimaker Do you mean static variables like `ENDPOINT1` and `SITE1`?

gimaker:
@davidxia Sorry I was referring to mutating static *non-final* variables, like the `CliConfigTest` does.

helios-tools/src/test/java/com/spotify/helios/cli/CliConfigTest.java
```diff
@@ -33,6 +34,14 @@
 
   @Rule public final ExpectedException expectedEx = ExpectedException.none();
 
+  @Before
```
gimaker: If you want to make sure `environment` is always empty unless explicitly set, should you not do this in `@After` also?

mattnworb:
I was considering rewriting the test for the same static var reasons (it also builds a bunch of Json with string literals), decided to do that in a separate PR.

helios-tools/src/test/java/com/spotify/helios/cli/CliConfigTest.java
mattnworb: yes very true

PR #684, opened 2015-10-08
refactor CliConfig to remove static reference to System.getenv()

mattnworb:
Storing the result of `System.getenv()` in a non-private static
reference causes the tests that touch `CliConfig` to need to worry about
setting or resetting this reference for the environment variables it
wants to simulate as being present for it's test case.

Refactoring `CliConfig.fromUserConfig()` to require the caller to pass
along what it thinks it's environment variables are (which is only
called from one place in non-test code) removes this need for the tests
worry about cleanup and state.

### DIFF
--- a/helios-tools/src/main/java/com/spotify/helios/cli/CliConfig.java
+++ b/helios-tools/src/main/java/com/spotify/helios/cli/CliConfig.java
@@ -21,7 +21,6 @@
 
 package com.spotify.helios.cli;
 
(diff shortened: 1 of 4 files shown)


mattnworb:
@gimaker

the tests that fail also fail on master :confused:

davidxia:
@mattnworb I fixed the DeregisterTests. Try rebasing on master.

PR #686, opened 2015-10-09
Add authentication module

davidxia:
Add helios-authentication service provider framework.


helios-authentication/src/main/java/com/spotify/helios/authentication/AuthProviderLoader.java
```diff
@@ -0,0 +1,198 @@
...
+      throw new RuntimeException("Failed to load plugin jar " + plugin, e);
+    }
+    final ClassLoader providedClassLoader = new FilteringClassLoader(PROVIDED, environment, parent);
+    return new URLClassLoader(new URL[]{url}, providedClassLoader);
+  }
+
+  /**
+   * A class loader that exposes a specific list of packages from another class loader.
+   */
+  private static class FilteringClassLoader extends ClassLoader {
```
mattnworb: this seems copied verbatim from `com.spotify.helios.serviceregistration.ServiceRegistrarLoader`, would be nice to not have to duplicate those

helios-authentication/src/main/java/com/spotify/helios/authentication/AuthClient.java
```diff
@@ -0,0 +1,27 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.helios.authentication;
+
+public interface AuthClient {
```
mattnworb: would be helpful to have a one-line summary javadoc here

helios-authentication/src/main/java/com/spotify/helios/authentication/AuthProviders.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+    return factory;
+  }
+
+  /**
+   * Get an authentication factory from the application class loader.
+   *
+   * @return The {@link AuthProviderFactory} object.
+   */
+  private static AuthProviderFactory createFactory() {
```
mattnworb: feels like `createFactory(Path)` and `createFactory()` could be one method

helios-authentication/src/main/java/com/spotify/helios/authentication/NopAuthClient.java
```diff
@@ -0,0 +1,30 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.helios.authentication;
+
+public class NopAuthClient implements AuthClient {
```
mattnworb: NopAuthClient and NopAuthHeader could probably be package-private instead of `public` if they are not intended to be used outside of the "factory" that returns them

mattnworb:
Some other thoughts that I think we can address outside of this PR:

1. The stuff where the `Nop` (I always thought it was spelt `noop` :wink: ) instances return `null` for every message feels like a perfect use case for `Optional`s instead

2. Would be very, very helpful to start documenting the ideas here even at this early stage, for all of
  - how to configure the master to use authentication
  - what the protocol between master and client looks like (i.e. the response header, the challenge, etc)
  - an overview of how the authentication code / modules within the master codebase are laid out (for example how the Provider gives you a Factory which gives you an Authenticator etc etc)

should be a lot easier to implement and understand all of this if we were working from an already-written markdown doc rather than documenting it after it's built

helios-services/src/main/java/com/spotify/helios/master/resources/AuthResource.java
```diff
@@ -0,0 +1,101 @@
...
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.HttpHeaders;
+import javax.ws.rs.core.Response;
+
+import static com.google.common.base.Strings.isNullOrEmpty;
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+import static javax.ws.rs.core.MediaType.TEXT_PLAIN;
+
+@Path("/_auth")
+public class AuthResource {
```
rohansingh: Talked about this on Friday, but it doesn't make sense for this to be here. It's tied too closely with crtauth. It should go into the helios-crtauth module. Either that or we should just make crtauth the only supported authentication mechanism and dump the whole modularity.

helios-authentication/src/main/java/com/spotify/helios/authentication/AuthProviderLoader.java
davidxia: We can lift this method and the one from helios-service-registration, as well as the two method above this, into a new subproject and have these two depend on that one in their poms. It's nice to not have to include any helios projects in the service provider poms though. If we didn't have to create, release, and depend on another helios subproject, this would be good to do, but it feels a bit not worth it to me?

helios-authentication/src/main/java/com/spotify/helios/authentication/NopAuthClient.java
davidxia: yup

davidxia: I'll do this for the other Nops

helios-services/src/main/java/com/spotify/helios/master/resources/AuthResource.java
davidxia: thanks for the reminder

mattnworb:
We'll also need a place to store expiration times for access tokens, or (if crtauth provides that by baking it into the token) then a method for checking if an arbitrary token (as present in the client request) is still valid.

davidxia:
Do we need to? The client can't tell from the token if it's expired since
the whole thing is hashed. The client can just always try the token if it
exists and redo the handshake to get a new one if expired?

On Tue, Oct 13, 2015 at 4:46 PM, Matt Brown <notifications@github.com>
wrote:

> We'll also need a place to store expiration times for access tokens, or
> (if crtauth provides that by baking it into the token) then a method for
> checking if an arbitrary token (as present in the client request) is still
> valid.
>
> —
> <https://github.com/spotify/helios/pull/686#issuecomment-147847747>.
>



-- 
David Xia
Software Engineer
dxia@spotify.com

helios-authentication/src/main/java/com/spotify/helios/authentication/AuthProviderLoader.java
mattnworb: I think it should be possible to refactor this so there doesn't need to be code duplicated and still make it possible for the "plugin implementation" to not have to refer to too much beyond importing the interface that it is implementing.

Another thing to keep in mind is that for authentication we might want to allow loading of more than one "provider" from ServiceLoader (rather than the first one it finds) to support different auth schemes etc.

I'll take a stab at this refactoring.

mattnworb:
@davidxia right, I just mean on the server-side we will need a way to validate that a given token from the request is still valid or not; so a method we can call in one of these interfaces like `boolean isValid(string token, string user)`.

davidxia:
@mattnworb Oh, gotcha. I think crtauth-java does that for us. https://github.com/spotify/crtauth-java/blob/0a7ca12fe1e052047b40aab6487fc9e12f4952f1/src/main/java/com/spotify/crtauth/CrtAuthServer.java#L275

davidxia:
@mattnworb I just included the helios-crtauth module.

mattnworb:
closing as #694 replaces this

PR #688, opened 2015-10-13
WIP: Authentication documentation and plan

mattnworb:
**Don't merge**

Adds a doc on the topic of authentication in Helios:
- how to enable it
- how crtauth works
- how to setup different schemes if desired
- document the protocol / flow between client and server

This is very rough and I'm really just sharing it at this point to be able to collaborate and discuss some of the details via Github.

I thought it would be useful to draft how we picture Authentication working, in all it's details, before starting to implement anything.

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -1,0 +1,77 @@
+# Authentication
+
+This document details how to configure your Helios installation to authenticate
+requests between the Helios CLI and the Helios masters/cluster.
+
+# TODOs for this document
+
+- configuration needed to setup master to require authentication
+  - make sure authentication can be enabled based on client version header, to
+    apply to versions >= some value and/or completely
+- add notion of schemes
+- link to how to setup https for helios masters
+
+## Authentication flow
+
+*Note that since all authentication requires clients to send some sort of
+credentials over the wire to the Helios masters, you should first configure
+your Helios clusters to use HTTPS*
+
+When configured to do so, the Helios master(s) will respond to any
+unauthenticated request with a `HTTP 401 Unauthorized` status code in the
+response. The master will set the `WWW-Authenticate` header in the response
+with the configured scheme for the client to use.
+
+When the client receives this response status code, it will look for a plugin
+registered for the given scheme. The Helios CLI will have support for the
+[`crtauth`][] scheme built in, but additional schemes can be added to the CLI
+by doing ... *TODO*.
+
+If the client cannot find a plugin for the scheme, it will be unable to proceed.
+
+After finding the plugin, the client activates it. The plugin is then
+responsible for performing authentication - for example in the crtauth flow (as
+explained below), it is the plugin's role to handle making the multiple HTTP
+requests for the two-step challenge-response flow.
+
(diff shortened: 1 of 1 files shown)


docs/authentication.md
```diff
@@ -0,0 +1,77 @@
...
+request to a protected resource on the service with an expired access token,
+the server will again respond with `401 Unauthorized` to signal that the client
+needs to perform the authentication flow again.
+
+### Crtauth flow
+
+To begin the crtauth flow, the client sends a request to `/_auth` on one of the
+masters to request a challenge. The client's request includes the username of
+the current user (`System.getProperty("user.name")` in Java, but this should be
+overridable via the CLI). The server will response with a challenge.
```
davidxia: yes with `--user <username`

docs/authentication.md
```diff
@@ -0,0 +1,77 @@
...
+
+The client signs the challenge using the user's configured SSH key and sends it
```
davidxia: private SSH key

docs/authentication.md
```diff
@@ -0,0 +1,77 @@
+# Authentication
+
+This document details how to configure your Helios installation to authenticate
+requests between the Helios CLI and the Helios masters/cluster.
+
+# TODOs for this document
+
+- configuration needed to setup master to require authentication
```
davidxia: How about starting the master with `--auth` and if no `--auth-plugin <path/to/jar>` is provided, it uses crtauth?

docs/authentication.md
```diff
@@ -0,0 +1,77 @@
...
+  - make sure authentication can be enabled based on client version header, to
```
davidxia: Another switch here when starting the master to make auth required for client versions greater than X? Defaults to requiring auth regardless of version if `--auth` is specified.

docs/authentication.md
mattnworb: How about something like `--requires-auth=STRING` where `STRING` can be a client version number to require auth after or just the string `yes`

```
--requires-auth=0.8.526
# or
--requires-auth=yes
```

docs/authentication.md
```diff
@@ -0,0 +1,77 @@
...
+    apply to versions >= some value and/or completely
+- add notion of schemes
+- link to how to setup https for helios masters
```
davidxia: I had docs in this PR: https://github.com/spotify/helios/pull/655/files#diff-ca6bc2e6c7f2fbcadd6202b5707a25adR108

mattnworb: yep that is exactly what I was referring to here

docs/authentication.md
davidxia: yea good idea

docs/authentication.md
```diff
@@ -0,0 +1,77 @@
+# Authentication
```
davidxia: Maybe just call this file "Security" since we can talk about encrypted comms, authentication, and later authorization together here? We can append a link like this to `README.md`:

```
* **[Security](security.md)**
  How to secure Helios
```

PR #689, opened 2015-10-13
master: Add metrics for client version for requests

rohansingh:
This will let us see what client versions are making requests to the master.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/servicescommon/statistics/MasterMetricsImpl.java
+++ b/helios-services/src/main/java/com/spotify/helios/servicescommon/statistics/MasterMetricsImpl.java
@@ -23,9 +23,9 @@ package com.spotify.helios.servicescommon.statistics;
 
 import com.google.common.collect.Maps;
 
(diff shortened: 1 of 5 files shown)


helios-services/src/main/java/com/spotify/helios/servicescommon/statistics/MasterMetricsImpl.java
```diff
@@ -75,6 +77,19 @@ private MasterRequestMetrics request(final String name) {
   }
 
   @Override
+  public void clientVersion(final String version) {
+    final String name = name(group, TYPE + "_meter_client_version", version);
+
+    Meter m = clientVersionMeters.get(name);
+    if (m == null) {
+      m = registry.meter(name);
+      clientVersionMeters.put(name, m);
```
mattnworb: since the MetricRegistry maintains a ConcurrentHashMap for all of the registered metrics, and examines it when you call `register.meter(name)`, feels like you could do away with the same map-keeping here, as it is pretty unlikely to try to request a metric with this name that isn't a meter

gimaker:
:+1:

PR #690, opened 2015-10-13
don't block when calling KafkaSender.send()

mattnworb:
Current implementation blocks for up to five seconds to be able to log
the outcome of the sending of the record to Kafka.

Since the code calling `KafkaSender` doesn't care and isn't made aware
of the results of sending to Kafka (and since any exceptions or timeouts
are swallowed in KafkaSender), the blocking isn't really necessary. In
the case where Kafka can't be reached, this causes the original
operation to wait an up to five seconds (per message sent) before
continuing on.

Changing the sending to Kafka to be truly fire-and-forget by performing
the logging of outcomes in a callback removes the blocking and the
impact of any timeouts from the calling-code.


mattnworb:

gimaker:
:+1: good catch

davidxia:
:+1:

PR #691, opened 2015-10-14
add validation of access tokens in http requests to master

mattnworb:
When the master is started with `--auth-required-versions`, `AccessTokenFilter` checks each HTTP request to the master for the presence of an access token in the http headers.

If the http header is not present, a response of 401 Unauthorized is
sent.

If the http header is present, the supplied access token is checked
against the `HttpAuthenticator` that the master is configured to use. If
the authenticator indicates that the token is not valid (or has
expired), a response of 401 Unauthorized is sent.

Otherwise the request is allowed to continue.

Adding this filter requires a few changes in the HttpAuthenticator
interface:

- add a method to expose the name of the auth scheme being used so that
  it can be sent to unauthenticated clients in the WWW-Authenticate
  header
- add a method for validating the supplied access token for the user's
  request

some TODOs:
- Not clear where the `username` to use when validating the token will
  come from in the http request
- Should probably link the new argument controlling whether
  authentication is enabled with the loading of a plugin that actually
  does the authenication - enabling authentication but using the no-op
  implementation seems pointless
- is the signature for `verifyToken(username, token)` legit for the way
  that crtauth works?

### DIFF
--- a/helios-authentication/src/main/java/com/spotify/helios/authentication/HttpAuthenticator.java
+++ b/helios-authentication/src/main/java/com/spotify/helios/authentication/HttpAuthenticator.java
@@ -23,6 +23,9 @@ package com.spotify.helios.authentication;
 
 public interface HttpAuthenticator {
 
+  /** The name of this authentication scheme */
+  String getSchemeName();
+
   String getHttpAuthHeaderKey();
 
   String createChallenge(String request) throws HeliosAuthException;
@@ -33,4 +36,6 @@ public interface HttpAuthenticator {
 
   String badAuthHeaderMsg();
 
+  // TODO (mbrown): correct semantics for this method
+  boolean verifyToken(String username, String accessToken);
 }
--- a/helios-services/src/main/java/com/spotify/helios/master/MasterConfig.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/MasterConfig.java
@@ -54,6 +54,7 @@ public class MasterConfig extends Configuration {
   private Path stateDirectory;
   private Path authPlugin;
   private String authSecret;
+  private String versionNumberRequiredForAuthentication;
 
   public String getDomain() {
     return domain;
@@ -224,6 +225,15 @@ public class MasterConfig extends Configuration {
 
   public MasterConfig setAuthSecret(final String authSecret) {
     this.authSecret = authSecret;
+    return this;
+  }
+
+  public String getVersionNumberRequiredForAuthentication() {
+    return versionNumberRequiredForAuthentication;
+  }
+
+  public MasterConfig setVersionNumberRequiredForAuthentication(String str) {
+    this.versionNumberRequiredForAuthentication = str;
     return this;
   }
 
--- a/helios-services/src/main/java/com/spotify/helios/master/MasterParser.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/MasterParser.java
@@ -21,9 +21,11 @@
 
 package com.spotify.helios.master;
 
+import com.spotify.helios.common.PomVersion;
 import com.spotify.helios.servicescommon.ServiceParser;
 
 import net.sourceforge.argparse4j.inf.Argument;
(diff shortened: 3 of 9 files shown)


helios-authentication/src/main/java/com/spotify/helios/authentication/HttpAuthenticator.java
```diff
@@ -33,4 +36,6 @@
 
   String badAuthHeaderMsg();
 
+  // TODO (mbrown): correct semantics for this method
+  boolean verifyToken(String username, String accessToken);
```
mattnworb: see AccessTokenFilter for why this is needed

helios-authentication/src/main/java/com/spotify/helios/authentication/HttpAuthenticator.java
```diff
@@ -23,6 +23,9 @@
 
 public interface HttpAuthenticator {
 
+  /** The name of this authentication scheme */
+  String getSchemeName();
```
mattnworb: see AccessTokenFilter for why this is needed

helios-services/src/main/java/com/spotify/helios/master/http/AccessTokenFilter.java
```diff
@@ -0,0 +1,108 @@
...
+    final String clientVersionHeader = httpRequest.getHeader(HELIOS_VERSION_HEADER);
+    if (clientVersionHeader == null) {
+      httpResponse.sendError(401, "Requests must include '" + HELIOS_VERSION_HEADER + "' header");
+      return;
+    }
+
+    final PomVersion clientVersion = PomVersion.parse(clientVersionHeader);
+    if (clientVersionRequiresAuth.apply(clientVersion)) {
+      final String username = null; // TODO (mbrown): where does this come from?
+      final String accessToken = httpRequest.getHeader(authenticator.getHttpAuthHeaderKey());
```
mattnworb: regarding `authenticator.getHttpAuthHeaderKey()`, feels like it would be semantic for things like access tokens to always be sent in an `Authorization` header rather than customizing the header name for each implementation. 

Is there a need for crtauth to use a non-standard header?

usually the format of the Authorization header is like `Authorization: <scheme> <token>` like `Authorization: crtauth abcdef1234`

helios-services/src/main/java/com/spotify/helios/master/MasterConfig.java
```diff
@@ -54,6 +54,7 @@
   private Path stateDirectory;
   private Path authPlugin;
   private String authSecret;
+  private String versionNumberRequiredForAuthentication;
```
davidxia: I don't think this variable name is long enough :)

helios-services/src/main/java/com/spotify/helios/master/MasterParser.java
```diff
@@ -101,4 +112,26 @@ protected void addArgs(final ArgumentParser parser) {
   public MasterConfig getMasterConfig() {
     return masterConfig;
   }
+
+  private static class VersionStringForAuthArgumentChoice implements ArgumentChoice {
+
+    @Override
+    public boolean contains(Object val) {
+      String value = String.valueOf(val);
+      if ("all".equalsIgnoreCase(value)) {
```
davidxia: Should we pick either `all` or `yes`?

mattnworb: either is fine; maybe squeezing the two things (enable or a version string) into one arg is too much

davidxia: I have no preferences. One switch seems easier for user but slightly more complex code. Two switches might be simpler code.

helios-services/src/main/java/com/spotify/helios/master/http/AccessTokenFilter.java
davidxia: You're right. basic auth, oauth also require credentials to be sent in the client's `Authorization` request header. We can remove this method.

PR #692, opened 2015-10-14
master: Add support for logback config & Kafka

rohansingh:
1. If a logback-access.xml exists in the master state directory, use that
   instead of the included logback-access.xml resource.

2. Add a logback-kafka-appender dependency so that access logs can be sent
   directly to Kafka.


mattnworb:
:+1:

PR #694, opened 2015-10-16
add authentication to helios

mattnworb:
WIP. @davidxia and I worked on this together.

Opening now for feedback and to share, will probably close and re-open again when properly tested (and the tests don't fail...)

This adds an (optional) authentication module to helios master, along with an (example) HTTP Basic implementation as well as one for crtauth.

There are some detailed notes in the commits and in [docs/authentication.md](https://github.com/mattnworb/helios/blob/authentication-new/docs/authentication.md) on how it all works so I won't repeat myself here.

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -1,0 +1,112 @@
+# Authentication
+
+This document details how to configure your Helios installation to authenticate
+requests between the Helios CLI and the Helios masters/cluster.
+
+# TODOs for this document
+
+- configuration needed to setup master to require authentication
+  - make sure authentication can be enabled based on client version header, to
+    apply to versions >= some value and/or completely
+- add notion of schemes
+- link to how to setup https for helios masters
+
+## Authentication flow
+
+*Note that since all authentication requires clients to send some sort of
+credentials over the wire to the Helios masters, you should first configure
+your Helios clusters to use HTTPS*
+
+When configured to do so, the Helios master(s) will respond to any
+unauthenticated request with a `HTTP 401 Unauthorized` status code in the
+response. The master will set the `WWW-Authenticate` header in the response
+with the configured scheme for the client to use.
+
+When the client receives this response status code, it will look for a plugin
+registered for the given scheme. The Helios CLI will have support for the
+[`crtauth`][] scheme built in, but additional schemes can be added to the CLI
+by doing ... *TODO*.
+
+If the client cannot find a plugin for the scheme, it will be unable to proceed.
+
+After finding the plugin, the client activates it. The plugin is then
+responsible for performing authentication - for example in the crtauth flow (as
+explained below), it is the plugin's role to handle making the multiple HTTP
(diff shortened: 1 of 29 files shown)


docs/authentication.md
```diff
@@ -0,0 +1,149 @@
...
+Helios uses the `java.util.ServiceLoader` facility for loading
+Authenticator instances from the classpath. The first Authenticator
+instance found that matches the configured scheme name is used. If no
+Authenticator instance is found, the master will fail to start up.
+
+### Implementing a new plugin
+
+A new plugin needs to implement three interfaces: 
+
+- com.spotify.helios.auth.ServerAuthentication, which tells the server (and
```
gimaker: Could use some minor formatting improvements. ServerAuthentication -> `ServerAuthentication`

rohansingh:
Does this replace #686 then?

helios-crtauth/src/main/java/com/spotify/helios/auth/crt/CrtAuthProvider.java
```diff
@@ -0,0 +1,127 @@
...
+          }
+        }
+      }
+
+      // if we have fallen through here, because the header is not present or is not in the
+      // expected structure, return 401 if the auth token is required.
+      if (required) {
+        throw new WebApplicationException(
+            Response.status(Status.UNAUTHORIZED)
+                // TODO (mbrown): check if this header value is correct
```
gimaker: My interpretation of http://www.ietf.org/rfc/rfc2617.txt is that `realm` must always be included, but the text isn't trivial to parse so I could be wrong.

mattnworb:
@rohansingh yep!

helios-crtauth/src/main/java/com/spotify/helios/auth/crt/CrtAuthProvider.java
mattnworb: I'm not sure if 2617 is relevant here, but this is how the protocol for crtauth is documented as far as I know: https://github.com/spotify/crtauth/blob/master/PROTOCOL.md#overview

mattnworb:
A bunch of TODOs I forgot to mention:

- checking the client version against the `--auth-minimum-version` flag is not implemented yet
- none of the client-side stuff, obviously
- I want to refactor the `ServerAuthentication` interface so that an implementer just has to return a `io.dropwizard.auth.Authenticator` instance. Right now they have to return the Jersey-specific `InjectableProvider` which is a) confusing and b) involves a lot of boilerplate that we can probably handle in our code instead.
- to actually authenticate all calls with this approach we have to add `@Auth(required = false) HeliosUser user` as a method parameter to **all** Jersey resource classes

mattnworb:
this work is now in #698

docs/authentication.md
mattnworb: I think I caught all these in #698 but let me know if you find more

PR #695, opened 2015-10-16
HeliosClient: Separate sending of requests from creation

gimaker:


gimaker:
@davidxia @rohansingh @mattnworb

helios-client/src/main/java/com/spotify/helios/client/Response.java
```diff
@@ -0,0 +1,91 @@
...
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+class Response {
+
+  private final String method;
+  private final URI uri;
+  private final int status;
+  private final byte[] payload;
+  private final Map<String, List<String>> headers;
```
mattnworb: guava's Multimap could be a good fit here

helios-client/src/main/java/com/spotify/helios/client/Response.java
```diff
@@ -0,0 +1,91 @@
...
+    return "Response{" +
+           "method='" + method + '\'' +
+           ", uri=" + uri +
+           ", status=" + status +
+           ", payload=" + Arrays.toString(payload) +
+           ", headers=" + headers +
+           '}';
+  }
+
+  private String decode(final byte[] payload) {
```
mattnworb: this method is never called in this class

helios-client/src/main/java/com/spotify/helios/client/Response.java
```diff
@@ -0,0 +1,91 @@
...
+    this.headers = headers;
+  }
+
+  @Override
+  public String toString() {
+    return "Response{" +
+           "method='" + method + '\'' +
+           ", uri=" + uri +
+           ", status=" + status +
+           ", payload=" + Arrays.toString(payload) +
```
mattnworb: If we are always dealing with character-based payloads (and maybe we aren't) then logging the decoded version of the payload in a toString() would probably be more useful than the raw byte array

helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
```diff
@@ -0,0 +1,298 @@
...
+        final URI realUri = new URI(scheme, host + ":" + port, fullpath, uri.getQuery(), null);
+        try {
+          log.debug("connecting to {}", realUri);
+          return connect0(realUri, method, entity, headers,
+                          ipToHostnameUris.get(ipEndpoint).getHost());
+        } catch (ConnectException | SocketTimeoutException | UnknownHostException e) {
+          // UnknownHostException happens if we can't resolve hostname into IP address.
+          // UnknownHostException's getMessage method returns just the hostname which is a useless
+          // message, so log the exception class name to provide more info.
+          log.debug(e.getClass().getSimpleName() + " - " + e.getMessage());
```
mattnworb: e.toString() would return the same thing

mattnworb:
+1 for moving the low-level http stuff out of HeliosClient.

With that stuff moved into a RequestDirector, seems like adding a HeliosClient unit test would be a lot easier now :wink:

helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
gimaker: almost :) e.toString will print the full class name, with the package. This has not been changed in this patch, just moved around but I can change it.

helios-client/src/main/java/com/spotify/helios/client/Response.java
gimaker: Indeed, but I'm not sure it adds much value?

helios-client/src/main/java/com/spotify/helios/client/Response.java
gimaker: AFAIK, we're always dealing with text payloads. That said I'm not sure we ever actually call `Reponse.toString()` anywhere. I changed it to use the decoded payload.

davidxia:
:+1:

PR #696, opened 2015-10-16
Use logstash logback encoder instead of Kafka

rohansingh:
The logstash folks have provided some a helpful logback extension to send
request logs directly to logstash. Use this and bypass Kafka altogether.

This also includes a change to helios-solo to support passing in a logstash
destination (useful for testing).


mattnworb:
:+1: :100:

PR #698, opened 2015-10-18
add authentication to helios masters

mattnworb:
This is a redo of #694 so I could move the branch with the changes from my fork to the official repo.

This adds an (optional) authentication module to helios master, along with an (example) HTTP Basic implementation as well as one for crtauth.

See detailed notes on how it works and the ideas behind it in [docs/authentication.md](https://github.com/helios/helios/blob/authentication-new/docs/authentication.md).

### TODOs
As this is still WIP, a list of TODOs:

- :white_medium_small_square: the client-side half of authentication
- :ballot_box_with_check: ~~checking the client version against the `--auth-minimum-version` flag is not implemented yet~~ 
- :ballot_box_with_check: ~~I want to refactor the `ServerAuthentication` interface so that an implementer just has to return a `io.dropwizard.auth.Authenticator` instance. Right now they have to return the Jersey-specific `InjectableProvider` which is a) confusing and b) involves a lot of boilerplate that we can probably handle in our code instead.~~
- :ballot_box_with_check: ~~to actually authenticate all calls with this approach we have to add `@Auth(required = false) HeliosUser user` as a method parameter to **all** Jersey resource classes~~
  - ~~may make more sense to define a new annotation of our own since we won't care about a `required` field with values potentially different at each annotated method (not our use case)~~
- :ballot_box_with_check: : ~~`AuthenticationLoader` does not yet load plugins defined in external jar files~~


helios-authentication/src/main/java/com/spotify/helios/auth/AuthenticationPluginLoader.java
```diff
@@ -0,0 +1,54 @@
...
+
+import com.google.common.collect.ImmutableList;
+
+import java.nio.file.Path;
+import java.util.List;
+import java.util.ServiceLoader;
+
+public class AuthenticationPluginLoader {
+
+  public static AuthenticationPlugin<?> load(ServerAuthenticationConfig config) {
```
gimaker: Presumably this will be used for loading plugins for the client as well - perhaps the `ServerAuthenticationConfig` name is unfortunate?

mattnworb: yea, let's refactor that when we have a better idea of what loading the client-side part will look like

mattnworb:

I cleaned up the integration between Jersey and the authentication plugin a bunch, I think it makes it simpler to understand how this all works. See notes in 34e0c0f. 

Only thing left AFAIK is being able to load the plugin from any arbitrary path. 

If you haven't taken a look already please review, this is just about my final draft.

docs/authentication.md
```diff
@@ -0,0 +1,155 @@
...
+
+After finding the plugin, the client activates it. The plugin is then
+responsible for performing authentication - for example in the crtauth flow (as
+explained below), it is the plugin's role to handle making the multiple HTTP
+requests for the two-step challenge-response flow. A simpler authentication
+scheme may just read a shared secret from a file on disk or an environment
+variable and populate the `Authorization` header with that value.
+
+The end result of the authentication process is that the Helios client will
+have a token for the `Authorization` header of it's HTTP requests that is known
```
davidxia: s/it's/its/

docs/authentication.md
```diff
@@ -0,0 +1,155 @@
...
+- [crtauth][]
+- [HTTP Basic Authentication][http-basic]
+
+## How to configure the Helios masters to require authentication
+
+Authentication is enabled in the master by adding the `--auth-scheme SCHEME`
+flag. This flag both enables authentication and configures the authentication
+scheme to be used.
+
+Authentication can be restricted to only apply to a certain client
```
davidxia: s/apply to a certain/apply to certain/

docs/authentication.md
```diff
@@ -0,0 +1,155 @@
...
+version and above, to allow for rolling out the authentication requirement
```
davidxia: s/version/versions/

docs/authentication.md
```diff
@@ -0,0 +1,155 @@
...
+slowly across a Helios installation. This is controlled with the
+`--auth-minimum-version VERSION` argument when starting the master (where
+`VERSION` is a Helios version string like `0.8.500`). If not set,
+authentication is applied to all clients and requests.
+
+To use an authentication plugin that is not shipped with Helios itself (such as
+a scheme you have authored), include the path to the JAR containing the plugin
+with the `--auth-plugin` argument to the master.
```
davidxia: Will the master fail to start if you give it an unknown scheme?

davidxia: Nvm, I see the below.

helios-authentication/src/main/java/com/spotify/helios/auth/basic/BasicAuthenticator.java
```diff
@@ -0,0 +1,88 @@
...
+            if (i > 0) {
+              final String username = decoded.substring(0, i);
+              final String password = decoded.substring(i + 1);
+              final BasicCredentials credentials = new BasicCredentials(username, password);
+              return Optional.of(credentials);
+            }
+          }
+        }
+      }
+    } catch (IllegalArgumentException e) {
```
davidxia: Do we need to catch this? I can't see what in the try would throw this.

docs/authentication.md
mattnworb: thanks, fixed all of these in 0fa40a2

helios-authentication/src/main/java/com/spotify/helios/auth/basic/BasicAuthenticator.java
mattnworb: it comes from `B64Code.decode(string, encoding)` if the string is not actually base64 encoded (for example if it contains chars outside of the base64 range)

davidxia: Ah ok

docs/authentication.md
```diff
@@ -0,0 +1,155 @@
...
+
+1. Translate HTTP headers in the request to a credentials type (which you
+   define yourself, specific to the plugin implementation). 
+2. When the credentials are present in the request, actually authenticate the
+   credentials.
+
+Helios will take care of the boilerplate and glue code necessary to hook into
+Jersey to call these methods for each HTTP request to a protected resource when
+authentication is enabled; the plugin only has to specify its unique logic
+(parsing HTTP headers, and authenticating the credentials to a user).
```
davidxia: nit: don't need this comma here?

helios-authentication/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+    <version>0.8.0-SNAPSHOT</version>
+  </parent>
+
+  <name>Helios Authentication</name>
+  <artifactId>helios-authentication</artifactId>
+  <packaging>jar</packaging>
+
+  <description>
+    SPI for authentication in Helios. Defines the framework for authentication using
+    plugins, and an example implementation providing HTTP Basic Authentication.
```
davidxia: nit: don't need this comma here either (not an independent clause).

helios-authentication/src/main/java/com/spotify/helios/auth/AuthenticationPlugin.java
```diff
@@ -0,0 +1,70 @@
...
+  ServerAuthentication<C> serverAuthentication();
+
+  ClientAuthentication<C> clientAuthentication();
+
+  interface ServerAuthentication<C> {
+
+    /**
+     * An Authenticator instance to use when authenticating HTTP requests to Helios.
+     * <p>
+     * Helio's Authentication support builds on top of the {@link io.dropwizard.auth.Authenticator
```
davidxia: s/Helio's/Helios'/

helios-authentication/src/main/java/com/spotify/helios/auth/basic/BasicServerAuthentication.java
```diff
@@ -0,0 +1,79 @@
...
+ * but demonstrations.
+ */
+public class BasicServerAuthentication implements ServerAuthentication<BasicCredentials> {
+
+  private final Map<String, String> users;
+
+  public BasicServerAuthentication() {
+    final String path = System.getenv("AUTH_BASIC_USERDB");
+    checkNotNull(path,
+        "Environment variable AUTH_BASIC_USERDB not defined, required for "
```
davidxia: Perhaps we should document this with a TODO or solicitation for contributions of different ways to look up passwords?

davidxia: Ah, I see the javadoc above. Maybe just mention that this implementation isn't very ready in the docs but that PRs are welcome?

helios-authentication/src/test/java/com/spotify/helios/auth/AuthenticationPluginLoaderTest.java
```diff
@@ -0,0 +1,66 @@
...
+import static org.hamcrest.CoreMatchers.instanceOf;
+import static org.junit.Assert.assertThat;
+
+public class AuthenticationPluginLoaderTest {
+
+  private final ServerAuthenticationConfig config = new ServerAuthenticationConfig();
+
+  @Test
+  public void canLoadPlugin() {
+    System.out.println(TestPlugin.class.getCanonicalName());
```
davidxia: Do we need this line?

helios-authentication/src/test/java/com/spotify/helios/auth/AuthenticationPluginLoaderTest.java
```diff
@@ -0,0 +1,66 @@
...
+    public ServerAuthentication<String> serverAuthentication() {
+      return null;
+    }
+
+    @Override
+    public ClientAuthentication<String> clientAuthentication() {
+      return null;
+    }
+  }
+}
```
davidxia: nitpick: newline here?

davidxia:
@mattnworb I'm loving the tests. Nice job.

helios-authentication/src/main/java/com/spotify/helios/auth/basic/BasicServerAuthentication.java
mattnworb: Yeah I struggled with how much of an "example" to make this versus being practical. 

To be useful, `BasicServiceAuthentication` needs some way to verify the supplied username/password, but if we wanted to do that by requiring the user to inject an `Authenticator` that does the verification, then it we don't have an easy way to do that given that the plugin loading infrastructure requires the class to have a no-arg constructor. 

We have the same problem in crtauth where the plugin is explicitly wiring the CrtAuthServer with the LDAP version of KeyProvider, even though any KeyProvider would work.

helios-authentication/src/test/java/com/spotify/helios/auth/AuthenticationPluginLoaderTest.java
mattnworb: nope, thanks for catching

helios-services/src/main/java/com/spotify/helios/master/MasterParser.java
```diff
@@ -79,9 +107,46 @@ protected void addArgs(final ArgumentParser parser) {
         .type(Integer.class)
         .setDefault(5802)
         .help("admin http port");
+
+    authenticationScheme = parser.addArgument("--auth-scheme")
+        .metavar("SCHEMENAME")
+        .help("Name of authentication scheme to use. "
+              + "Setting this flag enables authentication in Helios. "
+              + "'crtauth' support is built into Helios. "
```
davidxia: Should we also mention `http-basic` here or is the implementation too "basic" to be of any real use?

mattnworb: I don't think it is useful enough to mention here in the argparse help message, but I will add a note to authentication.md that it can be used (but really shouldn't, since it is so insecure) and which environment variables need to be set.

helios-crtauth/src/main/java/com/spotify/helios/auth/crt/CrtAuthenticationPlugin.java
```diff
@@ -0,0 +1,119 @@
...
+  @Override
+  public String schemeName() {
+    return "crtauth";
+  }
+
+  @Override
+  public ServerAuthentication<CrtAccessToken> serverAuthentication() {
+    // only validate the presence of environment variables when this method is called, as opposed to
+    // in the constructor, as the client-side code will not use the same environment variables
+    final String ldapUrl =  getRequiredEnv("CRTAUTH_LDAP_URL");
```
davidxia: Would requiring config file(s) be better than env vars? We could look in `/etc/default/helios-master` (which is already being written to and read from by the debian package) and then fallback to a json file in the current working directory.

helios-crtauth/src/main/java/com/spotify/helios/auth/crt/CrtAuthenticationPlugin.java
```diff
@@ -0,0 +1,119 @@
...
+    final int tokenLifetimeSecs = getOptionalEnv("CRTAUTH_TOKEN_LIFETIME_SECS", 540);
+
+    final LdapContextSource contextSource = new LdapContextSource();
+    contextSource.setUrl(ldapUrl);
+    contextSource.setAnonymousReadOnly(true);
+    contextSource.setCacheEnvironmentProperties(false);
+
+    final LdapTemplate ldapTemplate = new LdapTemplate(contextSource);
+
+    // TODO (mbrown): this should be general, support reading keys from flat files etc
```
davidxia: Should we allow combining multiple types of KeyProviders? Users might want to have LDAP for humans but flat files for CI agents because their IT department doesn't like adding non-human LDAP credentials. I'm not sure if this would make the implementation less secure though, eg timing attacks.

mattnworb: That sort of conditional logic - which classes to instantiate based on some config - is what I really wasn't sure how to do here. It's almost as if we would want to include some sort of Guice or Spring module for wiring up an user-supplied object graph of these things; I'm not sure if there demand is really there to match the complexity that would add.

mattnworb:
9c8eb21 adds support for loading the authentication plugin from an arbitrary path; which would be needed for any Helios user that wants to set up their own custom authentication scheme. 

With that I feel like this PR is mostly done; server-side authentication is available and can be enabled, and we have http-basic and crtauth implementations.

I'd like to still add more integration tests but I'm don't think those are necessary for this PR as authentication in the masters is not enabled by default (so if something is broken if we merge this now there is no impact).

Thoughts on merging @rohansingh @gimaker ?

mattnworb:
ugh I rebased on master on accident and now Github has re-ordered all the commits so the comments look out of order

mattnworb:
and now all the tests pass again in CircleCI after removing the thing where I tried to archive the logs from maven-invoker-plugin

mattnworb:
@gimaker found a nice way to modify SystemTestBase to be able to write system tests where auth is enabled, check out 32630ae

docs/authentication.md
```diff
@@ -0,0 +1,155 @@
...
+
+# TODOs for this document
+
+- link to how to setup HTTPS for Helios masters
+
+## Authentication flow
+
+*Note that since all authentication requires clients to send some sort of
+credentials over the wire to the Helios masters, you should first configure
+your Helios clusters to use HTTPS.*
```
rohansingh: I wonder, should we make it so the client just gives up and throws an exception if it's asked to authenticate over a plain HTTP connection?

mattnworb: Yeah I think that is a very good idea.

davidxia: :+1:

gimaker: Any ideas for *how* to do that?

mattnworb: AuthenticatingRequestDispatcher can check the protocol of the URI it is requesting prior to using the AuthProvider to get a token (either stored or new).

PR #699, opened 2015-10-19
CircleCI: Use increased parallelism

rohansingh:
Run DeploymentGroupTest in its own container since it takes almost five
minutes on its own.


davidxia:
How much more is the extra container?

mattnworb:
+1

rohansingh:
@davidxia We were already paying for it and not using it.

PR #700, opened 2015-10-19
client: Set request method before opening connection

rohansingh:
Fixes an issue wherein PUT's would be sent as POST's because we were opening the output stream before setting the request method.


rohansingh:
@davidxia

davidxia:
:+1:

rohansingh:
This was a red herring. The real issue is addressed by #703.

PR #701, opened 2015-10-19
Remove client retries based on IPs

davidxia:


davidxia:

The HostnameVerifier logic breaks `helios -z` when used with TLS. Resolving hostnames and retrying unique IPs would've been nice to have in the case of one hostname with multiple A records, but the added complexity and the regression don't seem worth the trouble.

rohansingh:
So this means retries won't round-robin to different hosts? I think that's the main thing that makes the retry useful (if a host is down, try another).

I think the fact that `-z` doesn't work is kind of expected. It's the same way that `curl https://my-actual-machine.something.cloud.spotify.net` will have an SSL error as opposed to `curl https://my-nice-cname.spotify.net`.

If we want to keep `-z` working, I'd prefer keeping the round-robin retry and having a flag to disable hostname verification.

mattnworb:
I agree about `-z https` not working shouldn't be a big deal.

We could add a `--insecure` flag like curl for this case, to disable verification of the certificate to the actual host / IP.

PR #703, opened 2015-10-19
client: Fix request method hack to work with HTTPS

rohansingh:
We have a horrible hack to set the request method on HttpURLConnection to
non-standard methods (like "PATCH"). It didn't work with HTTPS connections
though, due to the internals being different.

This fixes it for now, but a long-term fix would be to move to a non-stupid
HTTP library.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
+++ b/helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
@@ -280,12 +280,21 @@ class DefaultRequestDispatcher implements RequestDispatcher {
     // Nasty workaround for ancient HttpURLConnection only supporting few methods
     final Class<?> httpURLConnectionClass = connection.getClass();
     try {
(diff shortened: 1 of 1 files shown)


rohansingh:
@davidxia @gimaker @mattnworb

mattnworb:
[Apache HttpClient supports PATCH](https://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/client/methods/HttpPatch.html) :smirk:

mattnworb:
oh man: [the source code for HttpsURLConnectionImpl](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/sun/net/www/protocol/https/HttpsURLConnectionImpl.java?av=f#34)

:+1:

rohansingh:
Yeah, it's a mess. Was taking a look at Jersey source though and it looks like they do something very similar to patch around this limitation in `HttpsURLConnection`, so that makes me feel better.

PR #706, opened 2015-10-20
Implement authentication in HeliosClient and the CLI

gimaker:
Only basic auth is currently implemented.

TODO:
 - [x] Fix dependency issues
 - [x] Tests!
 - [ ] ~~CRTAuth impl~~ 
 - [ ] Make client plugins loadable from file system


gimaker:

gimaker:
Dependency hell.... :angry:

helios-authentication/src/main/java/com/spotify/helios/auth/AuthenticationPlugin.java
```diff
@@ -45,8 +45,6 @@
 
   ServerAuthentication<C> serverAuthentication();
 
-  ClientAuthentication<C> clientAuthentication();
```
gimaker: The `AuthenticationPlugin` interface doesn't really make sense anymore, with `ClientAuthentication` removed. Remove one layer here for the server-side as well? E.g. `AuthenticationPlugin` -> `ServerAuthPlugin`

helios-authentication/src/main/java/com/spotify/helios/auth/AuthenticationPluginLoader.java
```diff
@@ -70,4 +58,20 @@ private static ClassLoader pluginClassLoader(final Path plugin) {
       throw new RuntimeException("Failed to load plugin jar " + plugin, e);
     }
   }
+
+  private static ServiceLoader<AuthenticationPlugin> serviceLoaderForPath(final Path pluginPath) {
```
gimaker: Pardon the seemingly spurious changes -- this made sense one commit ago.

helios-tools/src/main/java/com/spotify/helios/auth/BasicClientAuthenticationPlugin.java
```diff
@@ -0,0 +1,50 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.helios.auth;
+
+import com.spotify.helios.client.RequestDispatcher;
+
+public class BasicClientAuthenticationPlugin implements ClientAuthenticationPlugin {
```
gimaker: Class names could be shorter here. Perhaps just call it `AuthenticationPlugin` on both the server and client and let the name be overloaded?

helios-authentication/src/main/java/com/spotify/helios/auth/AuthenticationPlugin.java
mattnworb: I think that would make sense; I initially thought it would be simplest for plugin-implementors to implement both at the same time / in the same "package".

helios-client/src/main/java/com/spotify/helios/auth/AuthProvider.java
```diff
@@ -0,0 +1,49 @@
...
+import com.spotify.helios.client.RequestDispatcher;
+
+public interface AuthProvider {
+
+  /**
+   * Get the current value to use in the "Authorization" header.
+   *
+   * @return {@code null} if the value is unavailable.
+   */
+  String currentAuthorization();
```
mattnworb: I suggest renaming this and the method below `currentAuthorizationToken` to make it clear it returns the token

helios-client/src/main/java/com/spotify/helios/auth/AuthProviderSelector.java
```diff
@@ -0,0 +1,74 @@
...
+
+  @Override
+  public String currentAuthorization() {
+    final AuthProvider authProvider = activeProvider.get();
+    return authProvider != null ? authProvider.currentAuthorization() : null;
+  }
+
+  @Override
+  public ListenableFuture<String> renewAuthorization(final String authHeader) {
+    // TODO(staffan): Support the case when multiple comma-separated auth schemes are returned in
```
mattnworb: we don't have support on the server-side either and I can't see why it would be useful so I don't think this needs to be worried about

helios-client/src/main/java/com/spotify/helios/auth/AuthenticatingRequestDispatcher.java
```diff
@@ -0,0 +1,101 @@
...
+    this.authProvider = authProvider;
+  }
+
+  @Override
+  public ListenableFuture<Response> request(final HeliosRequest request) {
+    // Include an Authorization header if it's currently available
+    final HeliosRequest req;
+    final String authHeader = authProvider.currentAuthorization();
+    if (authHeader != null) {
+      req = request.toBuilder().header(HttpHeaders.AUTHORIZATION, authHeader).build();
```
mattnworb: should break `request.toBuilder().header(HttpHeaders.AUTHORIZATION, authHeader).build();` into it's own method as you call it twice and the long chain is a little hard to read

helios-client/src/main/java/com/spotify/helios/auth/AuthenticatingRequestDispatcher.java
```diff
@@ -0,0 +1,101 @@
...
+
+  private ListenableFuture<Response> authenticateAndRetry(final HeliosRequest request,
+                                                          final Response response) {
+    final ListenableFuture<Response> f = Futures.transform(
+        authProvider.renewAuthorization(response.header(HttpHeaders.WWW_AUTHENTICATE)),
+        new AsyncFunction<String, Response>() {
+          @Override
+          public ListenableFuture<Response> apply(final String authHeader)
+              throws Exception {
+            return delegate.request(
```
mattnworb: I think that if the "new" token is the same as the previous one, we should return an error / not make a new request, assuming that the server will continue to think that value is not valid.

helios-client/src/main/java/com/spotify/helios/auth/AuthenticatingRequestDispatcher.java
```diff
@@ -0,0 +1,101 @@
...
+                request.toBuilder().header(HttpHeaders.AUTHORIZATION, authHeader).build());
+          }
+        });
+
+    // If authentication fails, return the first (unauthorized) response we got, as opposed
+    // to a failed future.
+    return Futures.withFallback(f, new FutureFallback<Response>() {
```
mattnworb: does `f` actually fail if the server returns `401 Unauthorized`? It looks to me like DefaultRequestDispatcher doesn't throw an exception on 4xx or 5xx

helios-client/src/main/java/com/spotify/helios/client/HeliosRequest.java
```diff
@@ -0,0 +1,171 @@
...
+  private final String method;
+  private final byte[] entity;
+  private final Map<String, List<String>> headers;
+
+  public HeliosRequest(final URI uri, final String method, final byte[] entity,
+                       final Map<String, List<String>> headers) {
+    this.uri = checkNotNull(uri);
+    this.method = checkNotNull(method);
+    this.entity = entity == null ? new byte[] {} : entity;
+    this.headers = checkNotNull(headers);
```
mattnworb: should probably copy the headers if you are aiming for immutability

helios-tools/src/main/java/com/spotify/helios/auth/ClientAuthenticationPluginLoader.java
```diff
@@ -0,0 +1,78 @@
...
+  public static List<ClientAuthenticationPlugin> loadAll() {
+    return loadAll(null);
+  }
+
+  public static List<ClientAuthenticationPlugin> loadAll(final Path pluginPath) {
+    final ServiceLoader<ClientAuthenticationPlugin> loader =
+        serviceLoaderForPath(pluginPath);
+    final List<ClientAuthenticationPlugin> plugins = Lists.newArrayList();
+    Iterators.addAll(plugins, loader.iterator());
+    return ImmutableList.copyOf(plugins);
```
mattnworb: I think these 3 lines could just be `return ImmutableList.copyOf(loader.iterator())`

mattnworb:
I like moving the method params for "an http request" into the HeliosRequest class

mattnworb:
Overall, looks good to me.

I couldn't follow why the `AuthProviderSelector` was necessary at first, so some comments about why it is necessary might be useful (since multiple plugins may be present?). I'm still not sure if there is a real need to use `AtomicReference` in there though.

At some point we should probably unite the 3 places in Helios that try to load things through `java.util.ServiceLoader`

mattnworb:
Possibly a style issue, but `AuthProvider.Factory` feels unnecessary to me. I think you could just move the `AuthProvider create(RequestDispatcher requestDispatcher)` method into the plugin interface and then you wouldn't need a factory.

helios-client/src/main/java/com/spotify/helios/auth/AuthProvider.java
gimaker: It doesn't return a token per se -- it returns the value to be put in the `Authorization` header. Which typically includes things like the scheme-name.

`currentAuthorizationHeader()` ?

helios-client/src/main/java/com/spotify/helios/auth/AuthenticatingRequestDispatcher.java
gimaker: Renewing the authorization can fail, which was the reason behind the fallback. The intent was to return the original response if we can't authenticate for some reason.

helios-client/src/main/java/com/spotify/helios/auth/AuthProvider.java
mattnworb: Good point. `currentAuthorizationHeader` sounds good to me

helios-client/src/main/java/com/spotify/helios/auth/AuthenticatingRequestDispatcher.java
mattnworb: Right, but my question was just if the future ever actually has an exception associated it versus returning a Future with a "failed" statuscode.

gimaker:
@mattnworb: Re `AuthProvider.Factory`, although I guess that would work technically, as the `ClientAuthenticationPlugin` is essentially a factory in it's own right. That said, I don't want to couple the plugins with the auth providers.

The `AuthProvider.Factory` is created before the `HeliosClient` is -- so we need some kind of factory.

helios-tools/src/main/java/com/spotify/helios/cli/CliMain.java
```diff
@@ -67,17 +76,37 @@ public static void main(final String... args) {
...
   public CliMain(final PrintStream out, final PrintStream err, final String... args)
       throws Exception {
+    this.authProviderFactory = configureAuthProvider();
     this.parser = new CliParser(args);
     this.out = out;
     this.err = err;
     setupLogging();
   }
 
+  private static AuthProvider.Factory configureAuthProvider() {
```
rohansingh: I kind of feel that this should actually go in `HeliosClient`. I mean, there's nothing really CLI-specific about it, is there? That way existing `HeliosClient` usages would continue to work without having to worry about doing `setAuthProviderFactory`. Thoughts?

gimaker: There was.

So, in an initial draft plugins had access to the CLI so that they could inject new parameters. In an effort to keep things smalled and focused, we dropped it for now. I still think it make sense though.

rohansingh: Meh. We could just be like, all auth params must be env vars.

helios-client/src/main/java/com/spotify/helios/auth/AuthenticatingRequestDispatcher.java
gimaker: The returned future typically returns a 401 back to the user if things fail. Which may not be ideal.

E.g. if you...
 * receive a 401
 * proceed to authenticate, but authentication fails
 * then the user will: get the first 401 back. and error message will be lost.

PR #707, opened 2015-10-21
client: Treat 502 Bad Gateway as connection error

rohansingh:
If we get a 502 Bad Gateway, we should treat it as a connection error and
retry the response with the next server (just like we do for a TCP error).
This lets retries work even when Helios is behind something like nginx.


rohansingh:
@davidxia @mattnworb @gimaker

mattnworb:
+1 that was fast

PR #708, opened 2015-10-21
kafka: Use an async non-blocking producer

rohansingh:
This will let messages be batched for better throughput, but also ensures
we never block on sending to Kafka.


mattnworb:
I think we should add `block.on.buffer.full` also: https://github.com/mattnworb/helios/commit/3e74c500745fc9ac740873a6c71a8f1c45371b26

rohansingh:
Ah yup, looks like you had it right. The configuration actually changed at some point, `block.on.buffer.full` is what the new config is supposed to be like.

rohansingh:
Fixed to use new config language. Imma merge it if there's no objections.

mattnworb:
ah cool, didn't realize `block.on.buffer.full` replaced `async`

PR #709, opened 2015-10-21
add developer guide

mattnworb:
Really just a scratchpad really for notes to help get new people
set up quickly based on my experience, for instance to
document the logging setup when running system-tests locally (which is
very smart but just not obvious until you dig into it).


gimaker:
:+1:

PR #710, opened 2015-10-21
build api-docs module after helios-testing

mattnworb:
The Jacoco setup in the api-docs module requires helios-testing to have
been build first (as it refers to it's `target/classes` directory).

Adding a `<dependency>` on helios-testing moves the API Documentation to
the last module to be built by Maven.

But in reality we are probably abusing Maven and the setup of this thing
is totally fragile.

### DIFF
--- a/helios-api-documentation/pom.xml
+++ b/helios-api-documentation/pom.xml
@@ -24,6 +24,11 @@
       <version>${project.version}</version>
     </dependency>
     <dependency>
+      <groupId>com.spotify</groupId>
+      <artifactId>helios-testing</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
       <groupId>com.google.code.findbugs</groupId>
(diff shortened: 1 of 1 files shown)


mattnworb:
this should resolve an error on our internal jenkins when `mvn -e -P sign-artifacts -P build-images -P jacoco -P build-solo clean deploy` is run. For some reason the build order was computed as

```
[INFO] Reactor Build Order:
[INFO] 
[INFO] Helios Parent
[INFO] Helios Authentication
[INFO] Helios CRTauth plugin
[INFO] Helios Service Registration
[INFO] Helios Client
[INFO] Helios Tools
[INFO] Helios Testing Common Library
[INFO] Helios Services
[INFO] Helios API Documentation
[INFO] Helios System Tests
[INFO] Helios Testing Library
[INFO] Helios Integration Tests
```

davidxia:
:-1: I wonder why it only broke on staffan's computer.

PR #711, opened 2015-10-21
Implement authentication in HeliosClient and the CLI

gimaker:
This implements on Basic authentication.

This does not yet support loading client-side authentication plugins from disk.


gimaker:

gimaker:
I'm feeling pretty "done" with this, with regards to this PR. There's more stuff to do, but it feels better to do that separately.

mattnworb:
Is the `AuthProvider.Context` really necessary? It feels to me like a way of hiding the parameters being passed

helios-authentication/src/main/java/com/spotify/helios/auth/AuthenticationPlugin.java
```diff
@@ -30,19 +30,25 @@
...
+   * The name of the scheme that this plugin provides, as used in the CLI to select this
+   * authentication plugin.
+   *
    * <p>
    * When the Helios master starts up and attempts to load all configured authentication plugins,
    * it will compare the return value of this method against the <code>--auth-scheme</code>
    * argument that it was started with.
    * </p>
    */
+  String cliSchemeName();
```
mattnworb: :+1:

mattnworb:
besides the comment above: :+1:

gimaker:
@mattnworb how does `AuthProviderSelector` know which implementation to use ahead of time?

The `Context` object enables adding more parameters later on without breaking the API, so yes - I'd say it's necessary if you don't want to break API in the future.

gimaker:
Hmm I guess if you defer creation of the `AuthProvider` until you get a 401, and pass in the auth-scheme when calling the factory...

gimaker:
rebased on master

PR #712, opened 2015-10-22
Implement client crtauth

davidxia:


davidxia:
@mattnworb @rohansingh @gimaker

helios-tools/src/main/java/com/spotify/helios/auth/CrtClientAuthenticationPlugin.java
```diff
@@ -0,0 +1,40 @@
...
+public class CrtClientAuthenticationPlugin implements ClientAuthenticationPlugin {
+
+  @Override
+  public String schemeName() {
+    return "crtauth";
+  }
+
+  @Override
+  public AuthProvider.Factory authProviderFactory() {
+    final String authServer = System.getenv("AUTH_CRT_SERVER");
```
davidxia: I need the crt auth server here somehow.

helios-client/src/main/java/com/spotify/helios/auth/CrtAuthProvider.java
```diff
@@ -0,0 +1,165 @@
...
+
+  @Override
+  public String currentAuthorizationHeader() {
+    // TODO (dxia) Should we use a lock here? It doesn't really matter if the token isn't
+    // thread-safe. We should get valid tokens each time. The lock just prevents
+    // another thread from doing extra work to get an extra token when it could just wait.
+    if (token == null) {
+      synchronized (lock) {
+        try {
+          token = renewAuthorizationHeader(null).get();
```
mattnworb: my understanding of the initial interface was that this method was for returning the token-stored-on-disk, if any

but in general, I don't think you need locking here.

helios-client/src/main/java/com/spotify/helios/auth/CrtAuthProvider.java
```diff
@@ -0,0 +1,165 @@
...
+
+      final ListenableFuture<Response> challengeFuture = dispatcher.request(request);
+
+      final ListenableFuture<Response> tokenFuture =
+          Futures.transform(challengeFuture, new AsyncFunction<Response, Response>() {
+            @Override
+            public ListenableFuture<Response> apply(@NotNull final Response response)
+                throws Exception {
+              final String challengeHeader = response.header(CRT_HEADER);
+              // TODO (dxia) Should we check status codes or just simply check headers as I've done?
```
mattnworb: should check status code, I believe the reference python implementation does (and the server will indeed send 200 OK as long as everything is OK)

helios-client/src/main/java/com/spotify/helios/auth/CrtAuthProvider.java
```diff
@@ -0,0 +1,165 @@
...
+                    "Missing HTTP X-CHAP header for crtauth challenge."));
+              }
+
+              final int i = challengeHeader.indexOf(':');
+              if (i == -1) {
+                return immediateFailedFuture(new CrtAuthException(
+                    "Invalid HTTP X-CHAP header for crtauth challenge."));
+              }
+
+              final String challenge = challengeHeader.substring(i + 1);
```
mattnworb: should also check that the part of the header value before `:` has the expected value

helios-client/src/main/java/com/spotify/helios/auth/CrtAuthProvider.java
```diff
@@ -0,0 +1,165 @@
...
+                  .method("GET")
+                  .uri(new URI(AUTH_URI))
+                  .appendHeader(CRT_HEADER, "response:" + crtResponse)
+                  .build();
+
+              return dispatcher.request(request);
+            }
+          });
+
+      return Futures.transform(tokenFuture, new AsyncFunction<Response, String>() {
```
mattnworb: minor but since every path through this AsyncFunction returns an immediate future, this can just be a Function (and you don't have to wrap the return values in Futures)

helios-client/src/main/java/com/spotify/helios/auth/CrtAuthProvider.java
```diff
@@ -0,0 +1,165 @@
...
+          if (isNullOrEmpty(tokenHeader)) {
+            return immediateFailedFuture(new CrtAuthException(
+                "Missing HTTP X-CHAP header for crtauth token."));
+          }
+
+          final int i = tokenHeader.indexOf(':');
+          if (i == -1) {
+            return immediateFailedFuture(new CrtAuthException(
+                "Invalid HTTP X-CHAP header for crtauth token."));
+          }
```
mattnworb: same comment as above; for completeness we should check the prefix of the header value

helios-client/src/test/java/com/spotify/helios/auth/BasicAuthProviderTest.java
```diff
@@ -37,8 +37,8 @@ public void test() throws Exception {
   }
 
   @Test
-  public void testCommanNotAllowedInUsername() throws Exception {
+  public void testCommandNotAllowedInUsername() throws Exception {
```
mattnworb: Command => colon?

helios-tools/src/main/java/com/spotify/helios/auth/CrtClientAuthenticationPlugin.java
mattnworb: can we call it `AUTH_CRT_SERVER_NAME` for total clarity?

helios-client/src/main/java/com/spotify/helios/auth/CrtAuthProvider.java
```diff
@@ -0,0 +1,165 @@
...
+  private final CrtAuthClient crtClient;
+  private volatile String token;
+  private final Object lock = new Object();
+
+  public CrtAuthProvider(final Context context,
+                         final String authServer) {
+    this.dispatcher = context.dispatcher();
+    this.username = context.user();
+    try {
+      this.crtClient = new CrtAuthClient(new AgentSigner(), authServer);
```
mattnworb: I would suggest having the constructor take the `CrtAuthClient` as a parameter (and wire it up in the `authProviderFactory()` method below); this way you can unit test this class with a mock and not have to do real ssh key stuff in the unit test.

helios-client/src/main/java/com/spotify/helios/auth/CrtAuthProvider.java
davidxia: Ah, If we're writing tokens to disk at a fixed path, should we make sure the writing to file part is thread-safe?

davidxia:
We need to edit https://github.com/spotify/helios/blob/master/helios-services/src/main/java/com/spotify/helios/master/MasterService.java#L263 to allow unauthed access to certain endpoints like `/_auth`.

helios-client/src/main/java/com/spotify/helios/auth/CrtAuthProvider.java
```diff
@@ -0,0 +1,172 @@
...
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.concurrent.ExecutionException;
+
+import static com.google.common.base.Strings.isNullOrEmpty;
+
+public class CrtAuthProvider implements AuthProvider {
+
+  private static final Logger log = LoggerFactory.getLogger(AuthProviderSelector.class);
```
gimaker: `AuthProviderSelector.class` --> `CrtAuthProvider.class`

helios-client/src/main/java/com/spotify/helios/auth/CrtAuthProvider.java
gimaker: If should return the current token (`token`) if there is one, otherwise it should return `null`. It should *not* block on getting a token.

helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
```diff
@@ -117,7 +117,7 @@ public void close() {
     dispatcher.close();
   }
 
-  private URI uri(final String path) {
+  URI uri(final String path) {
```
gimaker: spurious change?

helios-client/src/test/java/com/spotify/helios/auth/BasicAuthProviderTest.java
```diff
@@ -37,8 +37,8 @@ public void test() throws Exception {
   }
 
   @Test
-  public void testCommanNotAllowedInUsername() throws Exception {
+  public void testColonNotAllowedInUsername() throws Exception {
```
gimaker: :+1:

helios-tools/src/main/java/com/spotify/helios/auth/CrtClientAuthenticationPlugin.java
```diff
@@ -0,0 +1,51 @@
...
+public class CrtClientAuthenticationPlugin implements ClientAuthenticationPlugin {
+
+  @Override
+  public String schemeName() {
+    return "crtauth";
+  }
+
+  @Override
+  public AuthProvider.Factory authProviderFactory() {
+    final String authServer = System.getenv("AUTH_CRT_SERVER_NAME");
```
gimaker: Maybe default the server name to something like "heliosmaster"?

helios-client/src/main/java/com/spotify/helios/auth/CrtAuthProvider.java
gimaker: We probably need to tweak the `AuthProvider` interface to work well with the stored token, or rethink something else.

Currently, with `AuthProviderSelector`, the first request is *always* sent without authorization. Upon getting a 401 back it will select an `AuthProvider` and proceed to call `renewAuthorizationHeader()`. Now, we could tweak that to only call `renewAuthorizationHeader()` if the `currentAuthorizationHeader()` returns null on the selected provider - BUT, if we do that and the current token is expired the request will failed. Is there a way to tell if a token is expired on the client-side?

helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
davidxia: yea thanks for noticing this

helios-client/src/main/java/com/spotify/helios/auth/CrtAuthProvider.java
```diff
@@ -0,0 +1,159 @@
...
+
+  private static final String CRT_HEADER = "X-CHAP";
+  private static final String AUTH_URI = "https://helios/_auth";
+  private static final String CHALLENGE_PREFIX = "challenge";
+  private static final String TOKEN_PREFIX = "token";
+
+  private final RequestDispatcher dispatcher;
+  private final String username;
+  private final CrtAuthClient crtClient;
+  private final AtomicReference<String> tokenRef;
```
davidxia: I'm pretty much only using AtomicReference so I can later modify it inside the anonymous Function class here https://github.com/spotify/helios/pull/712/files#diff-04d67aa64d1fea0945d7a366258ab0ebR119. Is there a nicer way?

helios-client/src/main/java/com/spotify/helios/auth/CrtAuthProvider.java
davidxia: > Is there a way to tell if a token is expired on the client-side?

With crtauth no :( We can use any available cached tokens and if that gets
401, we redo the handshake to get a new one and cache that?

On Fri, Oct 23, 2015 at 10:17 AM, Staffan Gimåker <notifications@github.com>
wrote:

> In
> helios-client/src/main/java/com/spotify/helios/auth/CrtAuthProvider.java
> <https://github.com/spotify/helios/pull/712#discussion_r42869304>:
>
> > +                  final String username,
> > +                  final Signer signer) {
> > +    this.dispatcher = dispatcher;
> > +    this.username = username;
> > +    this.crtClient = new CrtAuthClient(signer, authServer);
> > +  }
> > +
> > +  @Override
> > +  public String currentAuthorizationHeader() {
> > +    // TODO (dxia) Should we use a lock here? It doesn't really matter if the token isn't
> > +    // thread-safe. We should get valid tokens each time. The lock just prevents
> > +    // another thread from doing extra work to get an extra token when it could just wait.
> > +    if (token == null) {
> > +      synchronized (lock) {
> > +        try {
> > +          token = renewAuthorizationHeader(null).get();
>
> We probably need to tweak the AuthProvider interface to work well with
> the stored token, or rethink something else.
>
> Currently, with AuthProviderSelector, the first request is *always* sent
> without authorization. Upon getting a 401 back it will select an
> AuthProvider and proceed to call renewAuthorizationHeader(). Now, we
> could tweak that to only call renewAuthorizationHeader() if the
> currentAuthorizationHeader() returns null on the selected provider - BUT,
> if we do that and the current token is expired the request will failed. Is
> there a way to tell if a token is expired on the client-side?
>
> —
> <https://github.com/spotify/helios/pull/712/files#r42869304>.
>



-- 
David Xia
Software Engineer
dxia@spotify.com

helios-client/src/main/java/com/spotify/helios/auth/CrtAuthProvider.java
mattnworb: I think you could just have a normal field of the class like `private String token` and assign it inside the anonymous Function. 

There would be some issues with thread-visibility (should it be `volatile` etc) but since this isn't code that is called very frequently I don't think it would matter ... but I also think the AtomicReference is fine too.

helios-client/src/main/java/com/spotify/helios/auth/CrtAuthProvider.java
```diff
@@ -0,0 +1,159 @@
...
+            tokenRef.set(token);
+            return token;
+          } catch (CrtAuthException e) {
+            throw Throwables.propagate(e);
+          }
+        }
+      });
+    } catch (URISyntaxException e) {
+      // This should never happen
+      throw Throwables.propagate(e);
```
mattnworb: minor: is it possible to have this try-catch wrap just whatever line is potentially throwing it?

mattnworb:
:+1: 

only comment would be (as I think I mentioned above) that if you passed in the `CrtAuthClient` as a parameter to the `CrtAuthProvider` constructor (i.e. move the instantiation of `CrtAuthClient` one layer up) then `CrtAuthProviderTest` could be a lot simpler, as you could mock out the calls to the client instead of providing actually valid ssh keys, but since you already have that test...

helios-client/src/main/java/com/spotify/helios/auth/CrtAuthProvider.java
davidxia: the problem is if it's not final, I can't reference it inside the anonymous `Function` class. And if it is final, I can't change it. So I need a reference to a String instead of a String itself.

davidxia: I could create an inner class but then I still need to pass the String by reference or somehow modify the enclosing instance of the inner class.

helios-client/src/main/java/com/spotify/helios/auth/CrtAuthProvider.java
davidxia: Yup can do.

helios-client/src/main/java/com/spotify/helios/auth/CrtAuthProvider.java
mattnworb: are you sure? it compiles for me

technically it should be possible to have an inner class (not a nested one) do this, as inner classes have an implicit reference to the parent instance

mattnworb: but I think the AtomicReference is fine too

davidxia:
@mattnworb Yea you're right. I didn't quite see it last time. thanks

PR #713, opened 2015-10-23
fix crtauth handshake bug; allow system tests to set environment variables

mattnworb:
### easier system tests

Refactor MasterService so that it is possible to pass custom environment
variables (represented as a `Map<String, String>`).

This makes it possible for system tests to start the master with custom defined
environment - previously not possible since the MasterMain is started in the
same process.

Also pass the Master's view of environment variables down to the
AuthenticationPlugin framework to simplify their loading logic.  

### allow plugins to set paths that don't require auth

This fixes a bug where the CrtHandshakeResource (at `/_auth`) requires a
token in order to get a token, which will probably never work.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/master/MasterService.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/MasterService.java
@@ -279,10 +279,29 @@ public class MasterService extends AbstractIdleService {
 
   private Predicate<HttpRequestContext> authenticationRequired(
       final ServerAuthentication<?> authentication, final ServerAuthenticationConfig config) {
-    if (config.isEnabledForAllVersions()) {
-      return Predicates.alwaysTrue();
-    }
 
+    // evaluate if the request is for the actual authentication endpoint
+    Predicate<HttpRequestContext> pathRequiresAuth = new Predicate<HttpRequestContext>() {
+      @Override
+      public boolean apply(final HttpRequestContext input) {
+        final String requestPath = input.getPath();
+        return !authentication.unauthenticatedPaths().contains(requestPath);
+      }
+    };
+
+    // if authentication is enabled for all versions then we just need to check that the request
+    // is not for an unauthenticated path.
+    // Otherwise we also check that the clientversion header is >= the minimum version that needs
+    // authentication
+
+    if (config.isEnabledForAllVersions()) {
+      return pathRequiresAuth;
+    }
+    return Predicates.and(pathRequiresAuth, clientVersionRequiresAuth(config));
+  }
+
+  private Predicate<HttpRequestContext> clientVersionRequiresAuth(
+      final ServerAuthenticationConfig config) {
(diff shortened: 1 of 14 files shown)


helios-services/src/main/java/com/spotify/helios/master/MasterService.java
```diff
@@ -270,11 +277,31 @@ private void setupAuthentication(final Environment environment, final MasterConf
...
+    // if authentication is enabled for all versions then we just need to check that the request
+    // is not for an unauthenticated path.
+    // Otherwise we also check that the clientversion header is >= the minimum version that needs
+    // authentication
+
     if (config.isEnabledForAllVersions()) {
-      return Predicates.alwaysTrue();
+      return pathRequiresAuth;
     }
+    return Predicates.and(pathRequiresAuth, clientVersionRequiresAuth(config));
```
mattnworb: predicates :metal:

mattnworb:

Ultimately it would be nicer if this "plugin" framework had a way for implementers to easily be able to wire up what they need without too much extra work. For example: crtauth needs an instance of `KeyProvider` and doesn't really care about the actual concrete class, but since we use `java.util.ServiceLoader` there is a requirement for the loaded classes to have no-arg constructors so passing "configuration" down to the plugin is a bit tricky. I want to stay away from putting too much logic around this, so we don't end up practically rewriting Guice or Spring or some other sort of DI framework, but at the same time all this stuff with environment variables feels super hacky.

helios-system-tests/src/main/java/com/spotify/helios/system/MasterAuthenticationTest.java
```diff
@@ -76,8 +92,8 @@ public String schemeName() {
     }
 
     @Override
-    public ServerAuthentication<String> serverAuthentication() {
-      return new ServerAuthentication<String>() {
+    public ServerAuthentication<String> serverAuthentication(Map<String, String> environment) {
+      return new SimpleServerAuthentication<String>() {
         @Override
         public Authenticator<String> authenticator() {
           return new Authenticator<String>() {
```
davidxia: Is the semicolon here https://github.com/spotify/helios/pull/713/files#diff-7f9fb5083fc38f934dca424ad96be781R111 required?

helios-services/src/main/java/com/spotify/helios/master/MasterService.java
davidxia: :+1:

davidxia:
Looks great. :+1:

helios-system-tests/src/main/java/com/spotify/helios/system/MasterAuthenticationTest.java
mattnworb: I tried to stage this into separate commits but in reality i wrote all the code at once; probably introduced it then

mattnworb:
added a test of HTTP Basic too, now that SystemTestBase allows for mocking out environment variables.

PR #714, opened 2015-10-26
test that users can proactively authenticate

mattnworb:
Adds a system test that enables authentication with
`--auth-minimum-version` set to a really high version (99.9.9) and
verifies that a client can supply `Authorization` headers when their 
version is less than this value and be authenticated if they really want to.

To verify this behavior, add a Jersey response filter that sets a header
`Authentication-Principal` (too enterprise-y??) containing the username of the
authenticated user. 

Since having someone else's authorization token gives you power to do anything
that user could do, finding out the username that corresponds to that token
seems like it would give no benefit to an attacker (who has already won).


mattnworb:

davidxia:
Is showing the username something we'd rather not leak when we move to authorization?

mattnworb:
@davidxia I could be missing something but I'm not sure what the "return username" endpoint would leak *if* someone needs to already have a token to access it

davidxia:
@mattnworb Yea I think you're right then.

PR #715, opened 2015-10-26
allow for eager client authentication

mattnworb:
Change the client authentication behavior so that the flag
`--force-authentication-scheme` will cause the client to immediately 
authenticate using the specified scheme.

If not set, then the previous behavior of waiting until the master
returns a `401 Unauthorized` response is kept.

To enable this, a small refactoring to the `AuthProvider.Factory`
interface: change the first parameter of the `create` method to `scheme`
instead of `authorization header value` - this moves the header parsing
up one layer.

### DIFF
--- a/helios-tools/src/main/java/com/spotify/helios/cli/CliParser.java
+++ b/helios-tools/src/main/java/com/spotify/helios/cli/CliParser.java
@@ -17,6 +17,7 @@
 
 package com.spotify.helios.cli;
 
+import com.google.common.base.Optional;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
@@ -95,6 +96,7 @@ public class CliParser {
   private final List<Target> targets;
   private final String username;
   private boolean json;
+  private final Optional<String> authenticationScheme;
 
   public CliParser(final String... args)
       throws ArgumentParserException, IOException, URISyntaxException {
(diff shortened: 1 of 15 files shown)


mattnworb:

The main part of the change is the additional Future handling in AuthenticatingRequestDispatcher.

mattnworb:
I also removed `Utils.getClient` and moved that method into the Command classes by introducing an `AbstractCliCommand` that they can all extend.

mattnworb:
It feels kludgy to me to have to add things to `CliMain` so that they can be passed from the CliParser/argparse-handling-layer over to how the HeliosClient is created, but I don't see another way

mattnworb:
also PRs to merge into other branches == :angry:

gimaker:
Is this for testing purposes only? I think most people will be too lazy to actually type out all those CLI flags. :)

This makes it impossible for `AuthProviders` act on authentication parameters (e.g. `realm`). Doesn't seem like a huge concern in practice though so maybe we're okay with that?

mattnworb:
@gimaker "testing" in that this way we could set up the CLI on some robots (Jenkins) to eagerly authenticate as a test before we roll it out to a larger user-base or use the "authentication required if client version >= some version" flag. 

I doubt anyone would want to type these in regular use, yeah, just for automation.

helios-tools/src/main/java/com/spotify/helios/cli/CliParser.java
```diff
@@ -308,6 +318,9 @@ public boolean getJson() {
       noLogSetup = addArgument("--no-log-setup")
           .action(storeTrue())
           .help(SUPPRESS);
+
+      authenticationScheme = addArgument("--force-authentication-scheme")
+        .help("Scheme to use with authentication; must be set if --force-authentication is set");
```
gimaker: The help seems wrong/outdated (`--force-authentication` doesn't exist).

gimaker:
looks alright to me

PR #716, opened 2015-10-27
refactor how Version.java is generated

mattnworb:
When changing the license header the other week it bugged me that
Version.java is generated in the `src/main/java` directory, as this
means that `mvn clean` doesn't touch it and any change in the template
requires some manual effort to remove the generated file.

So I moved it to `target/generated-sources` instead.


davidxia:
LGTM

mattnworb:
@rohansingh @gimaker

rohansingh:
:+1:

PR #717, opened 2015-10-27
client: Use ssh-agent identity for TLS client auth

rohansingh:
Use the local SSH agent identity to generate a self-signed certificate for
TLS client authentication, and sign the handshake hash with ssh-agent.

This lets the client setup a secure SSL connection and prove ownership of
its SSH public key. It is up to the server to verify the public key against
a known-good list (like in LDAP).

### DIFF
--- a/helios-client/pom.xml
+++ b/helios-client/pom.xml
@@ -57,6 +57,31 @@
       <artifactId>httpclient</artifactId>
       <version>4.5</version>
     </dependency>
+    <dependency>
+      <groupId>org.bouncycastle</groupId>
+      <artifactId>bcprov-jdk15on</artifactId>
+      <version>1.53</version>
+    </dependency>
+    <dependency>
+      <groupId>org.bouncycastle</groupId>
+      <artifactId>bcpkix-jdk15on</artifactId>
+      <version>1.53</version>
+    </dependency>
+    <dependency>
+      <groupId>com.eaio.uuid</groupId>
+      <artifactId>uuid</artifactId>
+      <version>3.2</version>
+    </dependency>
+    <dependency>
+      <groupId>com.spotify</groupId>
+      <artifactId>ssh-agent-proxy</artifactId>
+      <version>0.1.2</version>
+    </dependency>
+    <dependency>
+      <groupId>commons-codec</groupId>
+      <artifactId>commons-codec</artifactId>
+      <version>1.10</version>
+    </dependency>
 
     <!-- test deps -->
     <dependency>
--- a/helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
+++ b/helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
@@ -215,11 +215,16 @@ class DefaultRequestDispatcher implements RequestDispatcher {
         final URI realUri = new URI(scheme, host + ":" + port, fullpath, uri.getQuery(), null);
 
         AgentProxy agentProxy = null;
-        Deque<Identity> identities = null;
+        Deque<Identity> identities = Queues.newArrayDeque();
         try {
           if (scheme.equals("https")) {
             agentProxy = AgentProxies.newInstance();
-            identities = Queues.newArrayDeque(agentProxy.list());
+            for (final Identity identity : agentProxy.list()) {
+              if (identity.getPublicKey().getAlgorithm().equals("RSA")) {
+                // only RSA keys will work with our TLS implementation
+                identities.offerLast(identity);
+              }
+            }
           }
         } catch (Exception e) {
           log.warn("Couldn't get identities from ssh-agent", e);
@@ -228,9 +233,7 @@ class DefaultRequestDispatcher implements RequestDispatcher {
         try {
           boolean tryNextIdentity = false;
           do {
-            final Identity identity = (identities != null) ?
-                                      identities.poll() :
-                                      null;
+            final Identity identity = identities.poll();
             try {
               log.debug("connecting to {}", realUri);
               return connect0(realUri, method, entity, headers,
@@ -240,7 +243,7 @@ class DefaultRequestDispatcher implements RequestDispatcher {
               // there was some sort of security error. if we have any more SSH identities to try,
               // retry with the next available identity
               log.debug(e.toString());
-              tryNextIdentity = (identities != null) && !identities.isEmpty();
+              tryNextIdentity = !identities.isEmpty();
             } catch (ConnectException | SocketTimeoutException | UnknownHostException e) {
               // UnknownHostException happens if we can't resolve hostname into IP address.
               // UnknownHostException's getMessage method returns just the hostname which is a
--- a/helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
+++ b/helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
@@ -231,19 +231,26 @@ class DefaultRequestDispatcher implements RequestDispatcher {
         }
 
         try {
-          boolean tryNextIdentity = false;
           do {
             final Identity identity = identities.poll();
+
             try {
               log.debug("connecting to {}", realUri);
-              return connect0(realUri, method, entity, headers,
+
+              final HttpURLConnection connection = connect0(realUri, method, entity, headers,
                               ipToHostnameUris.get(ipEndpoint).getHost(),
                               agentProxy, identity);
-            } catch (SecurityException e) {
-              // there was some sort of security error. if we have any more SSH identities to try,
-              // retry with the next available identity
-              log.debug(e.toString());
-              tryNextIdentity = !identities.isEmpty();
+
+              final int responseCode = connection.getResponseCode();
+              if (((responseCode == HTTP_FORBIDDEN) || (responseCode == HTTP_UNAUTHORIZED))
+                  && !identities.isEmpty()) {
+                // there was some sort of security error. if we have any more SSH identities to try,
+                // retry with the next available identity
+                log.debug("retrying with next SSH identity since {} failed", identity.getComment());
+                continue;
+              }
+
+              return connection;
             } catch (ConnectException | SocketTimeoutException | UnknownHostException e) {
               // UnknownHostException happens if we can't resolve hostname into IP address.
               // UnknownHostException's getMessage method returns just the hostname which is a
@@ -252,11 +259,10 @@ class DefaultRequestDispatcher implements RequestDispatcher {
               // Connecting failed, sleep a bit to avoid hammering and then try another endpoint
               Thread.sleep(200);
             }
-          } while (tryNextIdentity);
+          } while (false);
         } finally {
           if (agentProxy != null) {
             agentProxy.close();
-            agentProxy = null;
           }
         }
       }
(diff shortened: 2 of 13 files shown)


helios-client/pom.xml
```diff
@@ -57,6 +57,31 @@
...
+    </dependency>
+    <dependency>
+      <groupId>com.eaio.uuid</groupId>
+      <artifactId>uuid</artifactId>
+      <version>3.2</version>
+    </dependency>
+    <dependency>
+      <groupId>com.spotify</groupId>
+      <artifactId>ssh-agent-proxy</artifactId>
+      <version>0.1.2-SNAPSHOT</version>
```
davidxia: 0.1.2 is released on maven central now

helios-client/src/main/java/com/spotify/helios/client/tls/SshAgentContentSigner.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  @Override
+  public OutputStream getOutputStream() {
+    return stream;
+  }
+
+  @Override
+  public byte[] getSignature() {
+    try (final DefaultAgentProxy agentProxy = new DefaultAgentProxy()) {
```
davidxia: I changed this in 0.1.2 to be `final AgentProxy agentProxy = AgentProxies.newInstance()`, ie I hid the impl class in a static factory.

helios-client/src/main/java/com/spotify/helios/client/tls/SshAgentSSLSocketFactory.java
```diff
@@ -0,0 +1,100 @@
...
+  @Override
+  public String[] getSupportedCipherSuites() {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public Socket createSocket(Socket s, String host, int port, boolean autoClose)
+      throws IOException {
+    final Identity identity;
+    try (final DefaultAgentProxy agentProxy = new DefaultAgentProxy()) {
```
davidxia: Same here `final AgentProxy agentProxy = AgentProxies.newInstance()`

helios-client/src/main/java/com/spotify/helios/client/tls/SshAgentTlsClient.java
```diff
@@ -0,0 +1,131 @@
...
+          new JcaX509CertificateConverter().setProvider("BC");
+
+      final List<X509Certificate> x509Certs = Lists.newArrayList();
+      try {
+        for (final org.bouncycastle.asn1.x509.Certificate cert :
+            serverCertificate.getCertificateList()) {
+          x509Certs.add(converter.getCertificate(new X509CertificateHolder(cert)));
+        }
+      } catch (CertificateException e) {
+        throw Throwables.propagate(e);
```
mattnworb: since the interface has `throws IOException` should probably package any exception as that type

mattnworb:
I think it would be good to have at least a few tests, for example `RecordingTlsClientProtocol` has some meaty logic in it outside of the stuff it inherits from it's superclass.

mattnworb:
Would this force anyone who wants to access their Helios installation over https into needing to have an ssh-agent process running, even if they didn't care to send client certificates in the TLS handshake?

helios-client/src/main/java/com/spotify/helios/client/tls/SshAgentContentSigner.java
```diff
@@ -0,0 +1,73 @@
...
+import org.bouncycastle.asn1.x509.AlgorithmIdentifier;
+import org.bouncycastle.operator.ContentSigner;
+import org.bouncycastle.operator.DefaultSignatureAlgorithmIdentifierFinder;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+
+/**
+ * Returns a signed SHA1 hash of any arbitrary data by sending it to the local SSH agent.
```
gimaker: Random note: I had a hard time figuring out that the `AgentProxy.sign()` method does the SHA1-hashing (as opposed to the content signer doing the hashing somewhere in the super class and then sending the hash for signing). It's not very apparent from the `AgentProxy` interface that it will hash and sign.

helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
```diff
@@ -233,9 +238,13 @@ private HttpURLConnection connect0(final URI ipUri, final String method, final b
     // We verify the TLS certificate against the original hostname since verifying against the
     // IP address will fail
     if (urlConnection instanceof HttpsURLConnection) {
+      checkNotNull(user, "username must be provided for HTTPS client auth");
```
gimaker: Maybe disable the use of client certs if `user` is `null`? It's conceivable that someone wants to use SSL without using client certificates for authentication.

helios-client/src/main/java/com/spotify/helios/client/tls/SshAgentTlsClient.java
rohansingh: Good point.

helios-client/src/main/java/com/spotify/helios/client/tls/SshAgentContentSigner.java
rohansingh: Yeah, I guess not. That's just the default behavior of ssh-agent behavior, all requests are SHA1-hashed and signed. Maybe we could change the method name though. @davidxia

helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
rohansingh: Good point.

rohansingh:
@mattnworb Good point. I'll try to refactor so that if there's an issue with instantiating the `AgentProxy`, we still continue the request, just without client-side certs.

rohansingh:
With the latest version, if there's any error with the `AgentProxy` or fetching identities, we just fallback to a normal HTTPS connection with no client certificate. Same with if a `user` is not passed to the `DefaultRequestDispatcher` constructor.

helios-client/src/main/java/com/spotify/helios/client/tls/SshAgentContentSigner.java
gimaker: Exactly how the signature is produced depends on the key type, so might be hard to be explicit unless you want to restrict the interface to specific key type.
 
"Upon receiving this request, the agent will look up the private key that
corresponds to the public key contained in key_blob. It will use this
private key to sign the "data" and produce a signature blob using the
key type-specific method described in RFC 4253 section 6.6 "Public Key
Algorithms"."

rohansingh: Touché.

gimaker: Explicitly being vague by removing the part about the SHA1 hash from the comment might actually make it clearer :)

davidxia: I updated java docs for `AgentProxy.sign()` https://github.com/spotify/ssh-agent-proxy/commit/db2f61cc6526460aa8cf0f2cc63afa8f225bd505

davidxia:
@rohansingh Let me know if another kind of exception from `AgentProxy.newInstance()` is more suitable.

helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
```diff
@@ -243,6 +297,11 @@ public boolean verify(String ip, SSLSession sslSession) {
           return new DefaultHostnameVerifier().verify(tHostname, sslSession);
         }
       });
+
+      if (!isNullOrEmpty(user) && (agentProxy != null) && (identity != null)) {
```
mattnworb: :+1:

mattnworb:
:+1: if you document this somehow (maybe update/replace authentication.md)

helios-client/src/main/java/com/spotify/helios/client/tls/SshAgentTlsSocket.java
```diff
@@ -0,0 +1,466 @@
...
+    underlying.setOOBInline(on);
+  }
+
+  @Override
+  public boolean getOOBInline() throws SocketException {
+    return underlying.getOOBInline();
+  }
+
+  @Override
+  public synchronized void setSoTimeout(int timeout) throws SocketException {
```
gimaker: `synchronized` is harmless but superfluous here

gimaker:
Looks good too me. Too much boilerplate and too few tests, but I can understand why :)

rohansingh:
Definitely needs more tests. Working on that.

rohansingh:
So this builds and it works, and there is some documentation now. Could use more tests though.

But also I'd like to merge it before it gets too far off master again, and I'm out of office until mid-November. Thoughts on merging?

mattnworb:
Nice job on the doc. Do you think you could also add an example of the nginx config to have it pass the certificate to an arbitrary endpoint (or point to the nginx doc on that topic)?

For using existing x509 certificates, could the client just check for a file in a well-known location before it contacts the ssh-agent? Or is there more hooks we have to add in besides just that?

rohansingh:
> Do you think you could also add an example of the nginx config to have it pass the certificate to an arbitrary endpoint (or point to the nginx doc on that topic)?

The doc has a link to the nginx SSL module documentation that shows the option to turn on client certificate support. I'll add an example config for validating the cert.

> For using existing x509 certificates, could the client just check for a file in a well-known location before it contacts the ssh-agent?

Yup, that's exactly what I was thinking. Figured we can implement that later though.

mattnworb:
> Yup, that's exactly what I was thinking. Figured we can implement that later though.

Right, I was asking more to see if it was a simple one or three line thing that could be added now or if it needs to be plugged into a lot more places with the bouncycastle stuff.

Assuming the answer is the latter, merging sounds good to me though.

davidxia:
I'm OK with merging. Awesome work all

On Thursday, October 29, 2015, Matt Brown <notifications@github.com> wrote:

> Yup, that's exactly what I was thinking. Figured we can implement that
> later though.
>
> Right, I was asking more to see if it was a simple one or three line thing
> that could be added now or if it needs to be plugged into a lot more places
> with the bouncycastle stuff.
>
> Assuming the answer is the latter, merging sounds good to me though.
>
> —
> <https://github.com/spotify/helios/pull/717#issuecomment-152265067>.
>


-- 
David Xia
Software Engineer
dxia@spotify.com

rohansingh:
@mattnworb It should be fairly simple. I actually hacked it for testing at some point. The actual work is only a few lines. Probably what'll be more of a pain is all the plumbing for configuring the cert location, reading & parsing it from disk, etc.

rohansingh:
I'll go ahead and merge this for now, and start work on an end-to-end integration test and certificate file support separately.

helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
```diff
@@ -196,17 +213,57 @@ private HttpURLConnection connect(final URI uri, final String method, final byte
...
-          return connect0(realUri, method, entity, headers,
-                          ipToHostnameUris.get(ipEndpoint).getHost());
-        } catch (ConnectException | SocketTimeoutException | UnknownHostException e) {
-          // UnknownHostException happens if we can't resolve hostname into IP address.
-          // UnknownHostException's getMessage method returns just the hostname which is a useless
-          // message, so log the exception class name to provide more info.
-          log.debug(e.toString());
-          // Connecting failed, sleep a bit to avoid hammering and then try another endpoint
-          Thread.sleep(200);
+          if (scheme.equals("https")) {
```
davidxia: Any reason we don't do this once outside the for loop?

davidxia:
Do we want to switch to Apache HTTP client from HttpUrlConnection at any point? DefaultRequestDispatcher seems pretty complex still. I'm wondering if it's worth it to clean it up.

PR #718, opened 2015-10-27
add more logging to ZooKeeperTestingServerManager

mattnworb:
Trying to test a theory that sometimes system tests fail in CircleCI
because two tests happen to run in parallel that conflict with each
other.


mattnworb:
well this didn't help diagnose the test problem but the additional logging would be good to have anyway

gimaker:
:+1: maybe log when it's stopped too?

PR #719, opened 2015-10-27
fix broken helios-authentication integration tests

mattnworb:
This broke in `4ff70b4` and was not noticed because the "integration" tests
within helios-authentication are not being run at all by CircleCI
(oops).


davidxia:
:+1:

PR #720, opened 2015-10-28
shade/relocate dns dependencies of helios-client

mattnworb:
For anyone adding recent version of `helios-testing` to their projects
*and* using `com.spotify:dns:2.x`, there is a dependency conflict
waiting to happen as there were changes in 3.x that are backwards
incompatible with 2.x.

This will relocate the DNS-related classes in `helios-client` (where the
dependency in `helios-testing` comes from) to prevent these headaches
for users.

Someone attempting to use the problematic versions of helios-testing
with version 2.x of `com.spotify:dns` will get errors like:

```
java.lang.NoSuchMethodError: com.spotify.dns.DnsSrvResolvers$DnsSrvResolverBuilder.metered(Z)Lcom/spotify/dns/DnsSrvResolvers$DnsSrvResolverBuilder;
```


mattnworb:
had to make the dependency on `dnsjava` in `helios-tools` explicit as it is no longer silently/transitively inherited via `helios-client`

mattnworb:

should fix a problem @alanw had this morning

gimaker:
does helios-tools use dnsjava? I couldn't any references to it. Although there does seem to be some dependencies on the xbill stuff.

mattnworb:
@gimaker it's in com.spotify.helios.cli.Output and cli.command.HostResolver

rohansingh:
To be honest I am a little bit scared of shading due to experiences with it in docker-client. Kind of wonder if it's better to just force people to upgrade.

mattnworb:
For reference the problematic project is getting `com.spotify:dns:2.x` from our Cool Awesome Internal RPC Framework v0.6 and upgrading to `com.spotify:dns:3.x` would mean moving to 1.x of Cool Awesome Internal RPC Framework.

mattnworb:
rebased on master to give the tests a chance to go green

rohansingh:
:+1:

PR #723, opened 2015-10-28
Fix tests that fail on docker 1.8.x

davidxia:
For some reason, docker throws an exception now when trying
to kill a container that's already not running.
So we now ignore this error.

### DIFF
--- a/helios-system-tests/src/main/java/com/spotify/helios/system/SystemTestBase.java
+++ b/helios-system-tests/src/main/java/com/spotify/helios/system/SystemTestBase.java
@@ -1107,6 +1107,15 @@ public abstract class SystemTestBase {
       public Object call() throws Exception {
         try {
           dockerClient.killContainer(containerId);
+        } catch (DockerRequestException e) {
+          if (e.message().contains("is not running")) {
+            // Container already isn't running. So we continue.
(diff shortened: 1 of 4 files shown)


davidxia:
@mattnworb @rohansingh @gimaker

helios-system-tests/src/main/java/com/spotify/helios/system/SystemTestBase.java
```diff
@@ -1107,6 +1107,17 @@ protected static void removeContainer(final DockerClient dockerClient, final Str
       public Object call() throws Exception {
         try {
           dockerClient.killContainer(containerId);
+        } catch (DockerException e) {
+          //noinspection StatementWithEmptyBody
+          if ((e instanceof DockerRequestException) &&
```
mattnworb: super minor: if you just `catch (DockerRequestException)` you can skip half this block

davidxia: oh yea duh

rohansingh:
+:1234::100:

davidxia:
:+1:

mattnworb:
finally :clap:

PR #727, opened 2015-11-01
client: use next ssh key on HTTP_FORBIDDEN

davidxia:


mattnworb:
does #728 replace this?

davidxia:
@mattnworb No the two are unrelated. I havent looked into why tests fail on this one yet.

davidxia:
@rohansingh

helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
```diff
@@ -243,7 +243,7 @@ private HttpURLConnection connect(final URI uri, final String method, final byte
 
               final int responseCode = connection.getResponseCode();
               if (((responseCode == HTTP_FORBIDDEN) || (responseCode == HTTP_UNAUTHORIZED))
-                  && !identities.isEmpty()) {
+                  && connection.getHeaderField("WWW-Authenticate") != null) {
```
davidxia: @rohansingh We need some way to differentiate between app-level and TLS-level unauthed responses. Would a header set by nginx work?

davidxia:
Replaced by https://github.com/spotify/helios/pull/749

PR #728, opened 2015-11-01
Move URI + IP logic from RequestDispatcher to Endpoint

davidxia:
The Endpoint object associates a URI with its resolved IP addresses.
Moving hostname-to-IP resolution logic out of DefaultRequestDispatcher
and into the static factory class Endpoints simplifies DefaultRequestDispatcher
and makes this logic more testable. DefaultRequestDispatcher is just
responsible for sending requests given a list of Endpoints.


davidxia:
@gimaker Let me know what you think of this approach.

davidxia:
TODO: move more endpoint validation and selection logic out of DefaultRequestDispatcher.

davidxia:
@rohansingh @mattnworb

helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
```diff
@@ -77,20 +73,16 @@
...
   private static final Logger log = LoggerFactory.getLogger(DefaultRequestDispatcher.class);
 
-  private static final long RETRY_TIMEOUT_MILLIS = SECONDS.toMillis(60);
   private static final long HTTP_TIMEOUT_MILLIS = SECONDS.toMillis(10);
-  private static final List<String> VALID_PROTOCOLS = ImmutableList.of("http", "https");
-  private static final String VALID_PROTOCOLS_STR =
-      String.format("[%s]", Joiner.on("|").join(VALID_PROTOCOLS));
 
-  private final Supplier<List<URI>> endpointSupplier;
+  private final Iterator<Endpoint> endpointIterator;
```
mattnworb: You can only iterate over the items in an Iterator once. Does this mean a new instance of DefaultRequestDispatcher has to be made for each request?

mattnworb: Ah I see that the `EndpointIterator` will cycle over the elements. In that case I'd declare this field as `EndpointIterator` to make that behavior clear, as this class depends on that behavior as opposed to the Iterator behavior. 

(It'd be even nicer if there was some interface to use in the declaration like `CyclingIterator`)

helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
```diff
@@ -77,20 +73,16 @@
...
   private final String user;
 
-  public DefaultRequestDispatcher(final Supplier<List<URI>> endpointSupplier,
-                                  final String user,
-                                  final ListeningExecutorService executorService) {
-    this.endpointSupplier = endpointSupplier;
+  DefaultRequestDispatcher(final Supplier<List<Endpoint>> endpointSupplier,
+                           final String user,
+                           final ListeningExecutorService executorService) {
+    endpointIterator = EndpointIterator.of(endpointSupplier.get());
```
mattnworb: if we only ever call `Supplier.get()` once, I think you could just simplify and have the constructor parameter be the actual list / iterable

helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
```diff
@@ -169,107 +161,67 @@ private HttpURLConnection connect(final URI uri, final String method, final byte
...
+      final String fullpath = endpointUri.getPath() + uri.getPath();
+
+      final String uriScheme = endpointUri.getScheme();
+
+      final URI ipUri = new URI(
+          uriScheme, endpointUri.getUserInfo(), endpoint.getIp().getHostAddress(),
+          endpointUri.getPort(), fullpath, uri.getQuery(), null);
+
+      AgentProxy agentProxy = null;
+      Deque<Identity> identities = Queues.newArrayDeque();
```
mattnworb: I think this could just be a regular Queue / LinkedList, we only insert at the end and remove from the head

helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
```diff
@@ -169,107 +161,67 @@ private HttpURLConnection connect(final URI uri, final String method, final byte
...
+      } finally {
+        if (agentProxy != null) {
+          agentProxy.close();
         }
       }
-      log.warn("Failed to connect, retrying in 5 seconds.");
-      Thread.sleep(5000);
-    }
-    throw new TimeoutException("Timed out connecting to master");
+    throw new HeliosException("foo");
```
mattnworb: :flushed:

helios-client/src/main/java/com/spotify/helios/client/EndpointIterator.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+class EndpointIterator implements Iterator<Endpoint> {
+
+  private final List<Endpoint> endpoints;
+  private final int size;
+  private int cursor;
+
+  private EndpointIterator(final List<Endpoint> endpoints) {
+    this.endpoints = checkNotNull(endpoints);
+    size = endpoints.size();
```
mattnworb: constructor should copy the list since it stores the size here, so that any modifications to the passed-in list don't have weird side effects

helios-client/src/main/java/com/spotify/helios/client/EndpointIterator.java
```diff
@@ -0,0 +1,74 @@
...
+  @Override
+  public Endpoint next() {
+    if (size == 0) {
+      throw new NoSuchElementException();
+    }
+
+    return cursor < size ? endpoints.get(cursor++) : endpoints.get(cursor = 0);
+  }
+
+  private int positive(final int value) {
```
mattnworb: method is unused

helios-client/src/main/java/com/spotify/helios/client/Endpoints.java
```diff
@@ -0,0 +1,170 @@
...
+    DefaultEndpoint(final URI uri, final InetAddress ip) {
+      this.uri = checkNotNull(uri);
+      this.ip = checkNotNull(ip);
+
+      final String scheme = this.uri.getScheme();
+      final String host = this.uri.getHost();
+      final int port = this.uri.getPort();
+      if (!VALID_PROTOCOLS.contains(scheme) || host == null || port == -1) {
+        throw new IllegalArgumentException(String.format(
+            "Master endpoints must be of the form \"%s://heliosmaster.domain.net:<port>\"",
```
mattnworb: if we don't actually care what format the hostname has (which this error message kind of suggests we do), I'd make this message be `Endpoints must be of the form %s://<host>:<port>`

helios-client/src/main/java/com/spotify/helios/client/RetryingRequestDispatcher.java
```diff
@@ -0,0 +1,159 @@
...
+  public void close() {
+    delegate.close();
+  }
+
+  private void startRetry(final SettableFuture<Response> future,
+                          final Supplier<ListenableFuture<Response>> code,
+                          final long deadline,
+                          final long delay,
+                          final TimeUnit timeUnit,
+                          final Predicate<Response> retryCondition) {
```
mattnworb: is `retryCondition` always `Predicates.alwaysTrue()`?

helios-client/src/main/java/com/spotify/helios/client/RetryingRequestDispatcher.java
```diff
@@ -0,0 +1,159 @@
...
+      codeFuture = code.get();
+    } catch (Exception e) {
+      handleFailure(future, code, deadline, delay, timeUnit, retryCondition, e);
+      return;
+    }
+
+    Futures.addCallback(codeFuture, new FutureCallback<Response>() {
+      @Override
+      public void onSuccess(Response result) {
+        if (retryCondition.apply(result)) {
```
mattnworb: the name here is confusing, because it seems to do the opposite:

```
if retry:
   resolve future with the result
else:
   call handleFailure and potentially retry
```

mattnworb:
`Clock` and `SystemClock` exist in helios-services already, which depends on `helios-tools` which depends on `helios-client`; so if you add a `Clock` and `SystemClock` to `helios-client` you could also remove the duplicates from `helios-services`.

I think the intent here makes sense, just small things to clean up above.

helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
davidxia: I could've used guava's `Iterators.cycle()` which will create an `Iterator` that cycles. But this doesn't let me randomize the starting point. I can declare `endpointIterator` as an `EndpointIterator` or make a `CyclingIterator` interface that extends `Iterator`. I noticed guava doesn't do this though. Their `Iterators.cycle()` just returns `Iterator<T>` and they note the cyclic nature in the javadoc.

helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
davidxia: Yea I didn't change it at first to keep the code changes minimal.

helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
davidxia: @rohansingh wrote this part and yea I was wondering why he used `Deque`. I'm going to leave this up to him though as it's not related to this PR.

helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
davidxia: oops

helios-client/src/main/java/com/spotify/helios/client/EndpointIterator.java
davidxia: thanks I always forget to make defensive copies.

helios-client/src/main/java/com/spotify/helios/client/RetryingRequestDispatcher.java
davidxia: It is. I lifted this code from github.com/spotify/futures-extra. I'll remove the predicate.

helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
mattnworb: I think that declaring this field as `EndpointIterator` would be best, since it communicates that the code only works with an instance of `EndpointIterator` rather than any `Iterator<Endpoint>`.

mattnworb:
:+1:

helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
rohansingh: Deques are the shit.

rohansingh:
:+1:

PR #729, opened 2015-11-01
Use `this` for all Objects.toStringHelper()

davidxia:
For consistency


davidxia:
@gimaker @rohansingh @mattnworb

davidxia:
Half the time we use `toStringHelper(getClass())` and the other half we use `toStringHelper(this)`. I just picked one.

gimaker:
sure :+1:

PR #731, opened 2015-11-04
Make PortMapping protocol default to "tcp"

### LINKED ISSUE #602: Not specifying protocol in job's port in job file throws NPE
If you `helios create -f <job_config_file.json>` and a port doesn't have the "protocol" key, NPE is thrown. This shouldn't happen as tcp is said to be the default in `helios create -h`.

```
  "ports" : {
    "http" : {
      "externalPort" : 8080,
      "internalPort" : 8080,
      "protocol" : "tcp"
    },
    "http-admin" : {
      "externalPort" : 8081,
      "internalPort" : 8081,
      "protocol" : "tcp"
    }
  },
```

davidxia:
The docs state that the protocol for a port mapping is optional and
defaults to "tcp". But if you try to do this, JobValidator will throw
NPE. Default to "tcp" in PortMapping to fix this.

Fixes #602

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/PortMapping.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/PortMapping.java
@@ -18,6 +18,7 @@
 package com.spotify.helios.common.descriptors;
 
 import com.google.common.base.Objects;
+import com.google.common.base.Optional;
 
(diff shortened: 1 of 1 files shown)


davidxia:
@gimaker @mattnworb

mattnworb:
:+1:

PR #733, opened 2015-11-05
Consolidate shared deps + plugins in parent pom

davidxia:
Consolidate shared dependency info in parent pom's
dependencyManagement section.


davidxia:
@mattnworb @gimaker

davidxia:
This was motivated by me trying to upgrade the docker-client dep in helios-services. That caused maven-enforcer to complain about a ton of upper-bound requirements. Fixing them was tedious as the dependency versions were duplicated everywhere.

mattnworb:
:+1: have been meaning to do this myself

PR #737, opened 2015-11-09
Update Readme

ReadmeCritic:
Created with https://github.com/dkhamsing/frankenstein


mattnworb:
these look like useful link updates actually

PR #742, opened 2015-11-12
Fix ZooKeeperMasterModel.getPreviousJobStates() again...

davidxia:


davidxia:
@negz

negz:
LGTM

negz:
Or wait, were you right all along?

https://docs.oracle.com/javase/8/docs/api/java/util/List.html#subList-int-int-

It says the high index is exclusive? So if you want to include the entire list you need to specify one index higher than what actually exists.

PR #743, opened 2015-11-12
Trying to fix shaded helios-tools jar

davidxia:


davidxia:
@mattnworb @negz @rohansingh

davidxia:
I couldn't find a better way that worked. This roughly doubles the size of the helios-tools jar. But I haven't measured the difference in speed which is more important here.

mattnworb:
If it's not obvious which classes need to be included from the jnr stuff to prevent any errors, and we also can't really do away with the jnr library for unix socket support, this makes sense to me - +1

PR #749, opened 2015-11-17
Fix logic for retrying with next SSH identity

rohansingh:


mattnworb:
easier to follow than ` do .. while (false)` :+1:

davidxia:
ha marginally?

PR #750, opened 2015-11-18
Track and expose Docker container state error.

negz:
@rohansingh @gimaker @mattnworb @davidxia 

http://docs.docker.com/engine/reference/api/docker_remote_api_v1.21/#inspect-a-container

This error (State.Error in the above API) reflects the last error encountered while attempting to start a container.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/TaskStatus.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/TaskStatus.java
@@ -56,7 +56,8 @@ import static java.util.Collections.emptyMap;
  *     }
  *   },
  *   "state" : "RUNNING",
- *   "throttled" : "NO"
+ *   "throttled" : "NO",
+ *   "containerError": "Something broke starting the container!"
  * },
  * </pre>
  */
@@ -85,6 +86,7 @@ public class TaskStatus extends Descriptor {
   private final ThrottleState throttled;
   private final Map<String, PortMapping> ports;
   private final Map<String, String> env;
+  private final String containerError;
 
   /**
    * @param job The job the task is running.
@@ -94,6 +96,7 @@ public class TaskStatus extends Descriptor {
    * @param throttled The throttle state of the task.
    * @param ports The ports actually assigned to the task.
    * @param env The environment passed to the container.
+   * @param containerError The last Docker error encountered while starting the container.
    */
   public TaskStatus(@JsonProperty("job") final Job job,
                     @Nullable @JsonProperty("goal") final Goal goal,
@@ -101,7 +104,8 @@ public class TaskStatus extends Descriptor {
                     @Nullable @JsonProperty("containerId") final String containerId,
                     @JsonProperty("throttled") final ThrottleState throttled,
                     @JsonProperty("ports") final Map<String, PortMapping> ports,
-                    @Nullable @JsonProperty("env") final Map<String, String> env) {
+                    @Nullable @JsonProperty("env") final Map<String, String> env,
+                    @Nullable @JsonProperty("containerError") final String containerError) {
     this.job = checkNotNull(job, "job");
     this.goal = goal; // TODO (dano): add null check when all masters are upgraded
     this.state = checkNotNull(state, "state");
@@ -111,17 +115,19 @@ public class TaskStatus extends Descriptor {
     this.throttled = Optional.fromNullable(throttled).or(ThrottleState.NO);
     this.ports = Optional.fromNullable(ports).or(EMPTY_PORTS);
     this.env = Optional.fromNullable(env).or(Maps.<String, String>newHashMap());
+    this.containerError = Optional.fromNullable(containerError).or("");
   }
 
   public Builder asBuilder() {
     return newBuilder()
-        .setJob(job)
-        .setGoal(goal)
-        .setState(state)
-        .setContainerId(containerId)
-        .setThrottled(throttled)
-        .setPorts(ports)
-        .setEnv(env);
+            .setJob(job)
+            .setGoal(goal)
+            .setState(state)
+            .setContainerId(containerId)
+            .setThrottled(throttled)
+            .setPorts(ports)
+            .setEnv(env)
+            .setContainerError(containerError);
   }
 
   private TaskStatus(final Builder builder) {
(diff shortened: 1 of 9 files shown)


negz:
I'd like to add a test or two to ensure this actually, you know, works. I figured I'd get the PR out early to ensure I'm on a reasonable path.

helios-client/src/main/java/com/spotify/helios/common/descriptors/TaskStatus.java
```diff
@@ -111,17 +115,19 @@ public TaskStatus(@JsonProperty("job") final Job job,
...
   public Builder asBuilder() {
     return newBuilder()
-        .setJob(job)
-        .setGoal(goal)
-        .setState(state)
-        .setContainerId(containerId)
-        .setThrottled(throttled)
-        .setPorts(ports)
-        .setEnv(env);
+            .setJob(job)
```
negz: I've configured IntelliJ to use the Google style guide and it *refuses* to let me have these at the indent level I found them.

helios-services/src/main/java/com/spotify/helios/agent/TaskMonitor.java
```diff
@@ -242,6 +243,20 @@ private void updateState(final TaskStatus.State state) {
...
+   * Propagate the container error by setting it and committing the update.
+   */
+  private void updateContainerError(final String containerError) {
+    statusUpdater.setContainerError(containerError);
+    // Commit and push a new status
+    try {
+      statusUpdater.update();
+    } catch (InterruptedException e) {
+      // TODO: propagate interrupt instead?
+      Thread.currentThread().interrupt();
```
mattnworb: this line and the comment above this line seem to contradict

helios-services/src/main/java/com/spotify/helios/agent/TaskRunner.java
```diff
@@ -161,6 +176,7 @@ private int run0() throws InterruptedException, DockerException {
             .build().newScheduler();
 
         while (!healthChecker.get().check(containerId)) {
+          // TODO(negz): Why aren't we using getContainerInfo() here?
```
mattnworb: fair point, seems like it should

mattnworb:
LGTM

helios-client/src/main/java/com/spotify/helios/common/descriptors/TaskStatus.java
```diff
@@ -210,6 +222,10 @@ public boolean equals(Object o) {
     if (throttled != that.throttled) {
       return false;
     }
+    if (containerError != null ? !containerError.equals(that.containerError) :
+            that.containerError != null) {
+      return false;
+    }
```
gimaker: `hashCode()` should be regenerated to look at `containerError` as well

negz: /me learns how to do that.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -653,6 +653,13 @@ private RollingUpdateOp rollingUpdateTimedoutError(final RollingUpdateOpFactory
               RollingUpdateError.IMAGE_PULL_FAILED,
               metadata);
     }
+    if (!taskStatus.getContainerError().equals("")) {
```
davidxia: This is probably a matter of personal preference, but guava has a `isNullOrEmpty()` that this code base seems to like a lot.

helios-services/src/main/java/com/spotify/helios/agent/TaskMonitor.java
negz: Yeah - this comment was ganked from the same `updateState` method I ganked most of it from. I could get rid of both?

As an aside, I'm also open to implementations that avoid me having to call `update()` the `statusUpdater` twice.

helios-services/src/main/java/com/spotify/helios/agent/TaskRunner.java
negz: I'll take care of that in a separate PR. :)

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: In `TaskStatus.java` containerError is created like:
```java
this.containerError = Optional.fromNullable(builder.containerError).or("")
```

Is it safe for me to assume that means `taskStatus.getContainerError()` should never return null? If so, is there still value in using `isNullOrEmpty()`? I notice there's also an `isEmpty()` method available on `taskStatus.getContainer()` already.

helios-services/src/main/java/com/spotify/helios/agent/TaskMonitor.java
davidxia: It seems like the updateState() method is a convenience method so the caller doesn't need to remember to call `StatusUpdater.update()` and catch an `InterruptedException`. Maybe instead of

```
    updateState(FAILED);
    updateContainerError(containerError);
 ```
you could do

```
    statusUpdater.setState(FAILED);
    statusUpdater.setContainerError(containerError);
    statusUpdater.update();
```

File: helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java (hunk 5 above)
Comment (davidxia) [inline helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java]: It should never return null. I guess `isNullOrEmpty()` looks nice and protects against NPEs. Otherwise, you could just use `isEmpty()`.

File: helios-client/src/main/java/com/spotify/helios/common/descriptors/TaskStatus.java [hunk 6]
```diff
@@ -165,53 +172,41 @@ public Job getJob() {
...
-      return false;
-    }
-    if (goal != that.goal) {
-      return false;
-    }
-    if (job != null ? !job.equals(that.job) : that.job != null) {
+    if (job != null ? !job.equals(that.job) : that.job != null) return false;
+    if (goal != that.goal) return false;
+    if (state != that.state) return false;
+    if (containerId != null ? !containerId.equals(that.containerId) : that.containerId != null)
```
davidxia: :( checkstyle fail

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: Done!

davidxia:
:+1:

PR #752, opened 2015-11-19
solo: Update Debian depends to allow docker-engine

rohansingh:
There is a new Debian package of Docker that's named "docker-engine".


rohansingh:
@gimaker @davidxia

davidxia:
:+1:

gimaker:
:100:

PR #754, opened 2015-11-19
Add a unit test for tracking Docker container state.

negz:
@mattnworb @gimaker @davidxia @rohansingh 

Functionality implemented in https://github.com/spotify/helios/pull/750

### DIFF
--- a/helios-services/src/test/java/com/spotify/helios/agent/TaskRunnerTest.java
+++ b/helios-services/src/test/java/com/spotify/helios/agent/TaskRunnerTest.java
@@ -23,6 +23,11 @@ import com.spotify.docker.client.DockerClient;
 import com.spotify.docker.client.DockerTimeoutException;
 import com.spotify.docker.client.ImageNotFoundException;
 import com.spotify.docker.client.ImagePullFailedException;
+import com.spotify.docker.client.messages.ContainerConfig;
+import com.spotify.docker.client.messages.ContainerCreation;
+import com.spotify.docker.client.messages.ContainerInfo;
+import com.spotify.docker.client.messages.ContainerState;
+import com.spotify.docker.client.messages.ImageInfo;
 import com.spotify.helios.common.Clock;
 import com.spotify.helios.common.HeliosRuntimeException;
 import com.spotify.helios.common.descriptors.Job;
@@ -39,7 +44,12 @@ import static java.util.Arrays.asList;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyString;
 import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
 @RunWith(MockitoJUnitRunner.class)
 public class TaskRunnerTest {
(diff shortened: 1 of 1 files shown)


helios-services/src/test/java/com/spotify/helios/agent/TaskRunnerTest.java
```diff
@@ -113,4 +129,49 @@ public void testPullTimeoutVariation() throws Throwable {
...
+            .docker(mockDocker)
+            .listener(mockListener)
+            .healthChecker(mockHealthChecker)
+            .build();
+
+    tr.run();
+
+    try {
+      tr.resultFuture().get();
+      fail("this should throw");
```
gimaker: Using the `ExpectedException` rule makes this a bit nicer in my opinion. Look here for more info: http://junit.org/apidocs/org/junit/rules/ExpectedException.html

helios-services/src/test/java/com/spotify/helios/agent/TaskRunnerTest.java
```diff
@@ -58,6 +69,11 @@
   @Mock private Clock clock;
   @Mock private ContainerDecorator containerDecorator;
 
+  @Mock private TaskRunner.NopListener mockListener;
```
gimaker: I guess this is a matter of opinion on style, but.... If these mocks are only used in a single test consider declaring them locally in that test.

E.g. instead of having members just do `final ImageInfo mockImageInfo = mock(ImageInfo.class)` etc in the test method.

helios-services/src/test/java/com/spotify/helios/agent/TaskRunnerTest.java
gimaker: Hmm, I guess that might not necessarily work if you use the future, via throwing of the exception, as a synchronization point which it kind of looks like you do? :)

mattnworb: +1

helios-services/src/test/java/com/spotify/helios/agent/TaskRunnerTest.java
```diff
@@ -113,4 +129,49 @@ public void testPullTimeoutVariation() throws Throwable {
...
+      @Override
+      public ContainerState state() {
+        final ContainerState state = mock(ContainerState.class);
+        when(state.running()).thenReturn(false);
+        when(state.error()).thenReturn("container is a potato");
+        return state;
+      }
+    };
+
+    doReturn("potato").when(mockCreation).id();
```
mattnworb: any particular reason to use this style instead of `when(mockCreation.id()).thenReturn("potato")` ?

helios-services/src/test/java/com/spotify/helios/agent/TaskRunnerTest.java
negz: +1 - I tried to do something similar using the @Mock annotation, but was told off by IntelliJ.

helios-services/src/test/java/com/spotify/helios/agent/TaskRunnerTest.java
negz: https://github.com/spotify/helios/commit/ab133a06749986d1e5707ca6913aaea9ccda8e13

I've really just copied the pattern from Drew's implementation in testPullTimeoutVariation. I'm open to switching ExpectedException, but I'm not clear as to how I'd inspect the ExecutionException to discover the real RuntimeException within if I did so.

gimaker:
Would it make more sense to write an integration test that tries to run a broken image and inspect the output of that?

IIUC this test mostly tests that `failed()` is called correctly on the listener when an exception is thrown by `run0()` ?

helios-services/src/test/java/com/spotify/helios/agent/TaskRunnerTest.java
mattnworb: Yet Another Opinion on Style but I like just declaring the fields (when they need to be fields) using the expression @gimaker mentioned, not super crazy about `@Mock` as I feel like it saves you just a few keystrokes

gimaker: @negz: Yeah you can't use the `@Mock` annotation for that. It's simple to do nonetheless.

Just do `final MyClass myMock = Mockito.mock(MyClass.class);` (bonus points for doing a static import of that function to make it a bit more terse`.

This is basically what "@Mock" does under the hoods.

helios-services/src/test/java/com/spotify/helios/agent/TaskRunnerTest.java
negz: I saw `doThrow` in the test above and used tab completion to test whether there was a similar `doReturn`. I then noticed the `when().thenReturn()` pattern in another test file. So not really. Is there a reason to prefer one or the other?

mattnworb: Not really, just wondering if you knew something I didn't know :smile: There are some cases where `doThrow()` can express things that the reverse style can't; but it's not a huge deal.

negz: @gimaker mentioned that `when().thenReturn()` is the more common pattern in the codebase, so I'm going to switch it over.

helios-services/src/test/java/com/spotify/helios/agent/TaskRunnerTest.java
negz: So I think I'll leave it this way if it's not a huge deal. `run()` is explicitly calling `result.setException()` so it seems nice to ensure that's being propagated, and it matches the pattern of the test above.

negz:
@gimaker I think an integration test makes a lot of sense, but I don't think that lessens the value of this test.

negz:
Anyone against me merging this?

mattnworb:
LGTM. 

Since you asked above, if you were to use ExpectedException you can [use matchers against the inner exception](http://junit.org/apidocs/org/junit/rules/ExpectedException.html#expectCause(org.hamcrest.Matcher)) or message or etc

negz:
Thanks!

PR #755, opened 2015-11-19
Improve error message when ssh-agent not configured

davidxia:


davidxia:
@rohansingh @mattnworb

mattnworb:
This will output once per connection, but some commands connect to more than one master (or connect more than once) so it might be very redundant.

I think it would be best if we could hold off on outputting anything, or at least a multi-line message, unless the server is returning 401 Unauthorized  / Forbidden / whatever

davidxia:
ah good point

davidxia:
How about if we output it once so that users can be warned?

helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
```diff
@@ -183,7 +183,15 @@ private HttpURLConnection connect(final URI uri, final String method, final byte
         }
       }
     } catch (Exception e) {
-      log.warn("Couldn't get identities from ssh-agent", e);
+      log.debug("Couldn't get identities from ssh-agent", e);
+      log.warn(
+          "Your Helios cluster is setup to work over HTTPS based on the Helios DNS records "
```
gimaker: HTTPS and the client-side certificates are two different things though, right? Having HTTPS-configured without using client-certs seems like a valid setup.

davidxia: Yes. I should be clearer about why using HTTPS, which is good in and of itself for encryption, causes the CLI to try to get your ssh-agent identities, which is needed for authentication which is hopefully done over an encrypted channel.

gimaker: By design authentication cannot happen over an unencrypted channel :)

davidxia:
Replaced with https://github.com/spotify/helios/pull/757

PR #756, opened 2015-11-19
Make ssh-agent error message debug instead of warn

davidxia:


davidxia:
For now. @rohansingh @mattnworb @gimaker

PR #757, opened 2015-11-20
Optimize DefaultRequestDispatcher and improve ssh-agent warning message

davidxia:
and only get them if an endpoint is HTTPS.
We were asking the ssh-agent for the list of identities
for every new connection to an endpoint.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
+++ b/helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
@@ -91,6 +91,14 @@ class DefaultRequestDispatcher implements RequestDispatcher {
         identities = getSshIdentities();
       } catch (Exception e) {
         log.debug("Couldn't get identities from ssh-agent", e);
+        log.warn(
+            "Your Helios cluster is setup to work over HTTPS based on the Helios DNS records "
+            + "found by this Helios CLI. This will provide client and server-side authentication. "
+            + "The team that operates your Helios cluster may choose to make this authentication "
+            + "optional for now in which case this is just a warning.\nWhen authentication is "
+            + "required, you'll need to run ssh-agent and set the SSH_AUTH_SOCK environment "
+            + "variable whenever you invoke this CLI."
+        );
       }
     }
   }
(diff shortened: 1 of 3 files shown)


mattnworb:
Is the `Identity` instance usable after the `AgentProxy` has been closed?

mattnworb:
Looks like it is usable, so nevermind.

mattnworb:
:+1: to the change itself.

I think this class would be a lot more testable though (it has no test today) if instead you passed either the `List<Identity>` or `AgentProxy` instance in the constructor, so that the class that is making http requests didn't also have to be concerned with how to get the ssh-agent identities or how to get an AgentProxy etc.

davidxia:
Yea, I haven't had the chance to add a test yet. Good idea about passing in
the list of identities. But it seems like if you don't create a new
instance of an AgentProxy on every connection, the TLS handshake is messed
up somehow. I wasn't able to drill down into why though...

On Fri, Nov 20, 2015 at 1:09 PM, Matt Brown <notifications@github.com>
wrote:

> [image: :+1:] to the change itself.
>
> I think this class would be a lot more testable though (it has no test
> today) if instead you passed either the List<Identity> or AgentProxy
> instance in the constructor, so that the class that is making http requests
> didn't also have to be concerned with how to get the ssh-agent identities
> or how to get an AgentProxy etc.
>
> —
> <https://github.com/spotify/helios/pull/757#issuecomment-158478315>.
>



-- 
David Xia
Software Engineer
dxia@spotify.com

helios-client/src/main/java/com/spotify/helios/client/DefaultRequestDispatcher.java
```diff
@@ -83,6 +85,22 @@
     endpointIterator = EndpointIterator.of(endpoints);
     this.executorService = executorService;
     this.user = user;
+
+    if (endpointIterator.hasHttps()) {
+      try {
+        identities = getSshIdentities();
+      } catch (Exception e) {
+        log.debug("Couldn't get identities from ssh-agent", e);
+        log.warn(
```
davidxia: @gimaker @mattnworb This should now happen only once. Open to suggestions on how to improve this message.

davidxia:
I'm having trouble thinking of a sane way to make `DefaultRequestDispatcher` more testable. How could I mock out the URL connection part?...

PR #760, opened 2015-11-23
Allow specifying a job JSON for a temporary job

rohansingh:
Currently folks store their job configurations for their real jobs in JSON
files, but configure their temporary testing jobs programmatically. This
has the downside of making the job that's run in integration tests
different from the job that actually gets deployed.

To address this, allow creating a TemporaryJob from a job config JSON.
Ideally this should be the same JSON as the real job. For example:

    temporaryJobs.jobWithConfig(".helios/job_config.json").deploy()

### DIFF
--- a/helios-testing/src/test/java/com/spotify/helios/testing/JobWithConfigTest.java
+++ b/helios-testing/src/test/java/com/spotify/helios/testing/JobWithConfigTest.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright (c) 2015 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.helios.testing;
+
+import com.google.common.io.Resources;
+import com.google.common.util.concurrent.ListenableFuture;
+
+import com.spotify.helios.client.HeliosClient;
+import com.spotify.helios.common.descriptors.Job;
+import com.spotify.helios.common.descriptors.JobId;
+
+import org.junit.Before;
+import org.junit.Rule;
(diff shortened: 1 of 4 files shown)


rohansingh:
@staffan @mattnworb @davidxia

gimaker:
:+1:

helios-testing/src/test/java/com/spotify/helios/testing/JobWithConfigTest.java
```diff
@@ -0,0 +1,101 @@
...
+  @Mock
+  private static Deployer deployer;
+
+  @Before
+  public void setup() {
+    final ListenableFuture<Map<JobId, Job>> future =
+        immediateFuture((Map<JobId, Job>) new HashMap<JobId, Job>());
+
+    // Return an empty job list to skip trying to remove old jobs
+    when(client.jobs()).thenReturn(future);
```
mattnworb: since the `client` is static (and why is that?) this adds a new mock expectations for each `@Test`. 

If the mock is static then this should probably be `@BeforeClass` or more simply they should both be non-static

helios-testing/src/test/java/com/spotify/helios/testing/JobWithConfigTest.java
```diff
@@ -0,0 +1,101 @@
...
+      final TemporaryJobBuilder builder = temporaryJobs.jobWithConfig(configFile)
+          .port("https", 443);
+      builder.deploy("test-host");
+
+      final ArgumentCaptor<Job> captor = ArgumentCaptor.forClass(Job.class);
+      verify(deployer).deploy(captor.capture(), anyListOf(String.class), anySetOf(String.class),
+                              any(Prober.class), any(TemporaryJobReports.ReportWriter.class));
+
+      assertEquals(80, captor.getValue().getPorts().get("http").getInternalPort());
+      assertEquals(443, captor.getValue().getPorts().get("https").getInternalPort());
```
mattnworb: 443 but the json file has 456?

rohansingh: Yeah that's on purpose, to see if the override on line 85 is successful.

helios-testing/src/test/java/com/spotify/helios/testing/JobWithConfigTest.java
rohansingh: The members are static so that they can be accessed from with `JobWithConfigTestImpl`. The `setup` is not `@BeforeClass` because if it is, the mocks aren't ready at that point. In practice though the setup only runs once since there is only one `test()` method that runs all the tests in the inner class.

PR #763, opened 2015-11-24
solo: Fix get_helios_version check in helios-use

rohansingh:
Due to a change in the formatting of the output from `helios --version`,
get_cli_version was returning a newline-separated string containing both
the CLI version and the target Docker version.

Fix it to only return the CLI version.


rohansingh:
@gimaker

davidxia:
:+1:

PR #764, opened 2015-11-24
Make master in tests bind to 0.0.0.0

davidxia:
instead of localhost.

Our previous build server was a VM and
InetAddress.getAllByName("localhost") returned only 127.0.0.1.
For some reason, on physical machines
InetAddress.getAllByName("localhost") also returns
::1 which is the IPV6 loopback address.

For some other mysterious reason, when maven runs tests that
start a master process like ProberTest, the master binds to
only 127.0.0.1. And tests that try to connect the local master via IPv6
will fail. But when you run the master process with `java`,
the master will bind to ::1 as well (assuming you don't set
`-Djava.net.preferIPv4Stack=true`).

And for some reason, setting SystemTestBase.setupDefaultMaster()
to start the master to bind to the same host as specified
in MasterParser's default of 0.0.0.0 makes tests like ProberTest
pass on bare metal machines.

Specifying 0.0.0.0 instead of localhost will accept external
connections, but this shouldn't be a big deal.

`¯\_(ツ)_/¯`

### DIFF
--- a/helios-system-tests/src/main/java/com/spotify/helios/system/SystemTestBase.java
+++ b/helios-system-tests/src/main/java/com/spotify/helios/system/SystemTestBase.java
@@ -543,7 +543,7 @@ public abstract class SystemTestBase {
     final List<String> argsList = Lists.newArrayList(
         "-vvvv",
         "--no-log-setup",
-        "--http", "http://localhost:" + (masterPort() + offset),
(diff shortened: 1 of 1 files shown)


davidxia:
@mattnworb

mattnworb:
:+1:

PR #765, opened 2015-11-25
Remove deployment group ZK nodes in transaction

davidxia:
If data in ZooKeeper for a deployment group gets
denormalized, make sure we still delete
all relevant nodes for a deployment group
when the remove-deployment-group command is run.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
@@ -340,7 +340,7 @@ public class ZooKeeperMasterModel implements MasterModel {
             create(Paths.configDeploymentGroup(deploymentGroup.getName()), deploymentGroup),
             create(Paths.statusDeploymentGroup(deploymentGroup.getName())),
             create(Paths.statusDeploymentGroupHosts(deploymentGroup.getName()))
-          );
+        );
       } catch (final NodeExistsException e) {
         throw new DeploymentGroupExistsException(deploymentGroup.getName());
       }
@@ -375,13 +375,12 @@ public class ZooKeeperMasterModel implements MasterModel {
     final ZooKeeperClient client = provider.get("removeDeploymentGroup");
     try {
       client.ensurePath(Paths.configDeploymentGroups());
-      client.delete(Paths.configDeploymentGroup(name));
-      if (client.exists(Paths.statusDeploymentGroupHosts(name)) != null) {
-        client.delete(Paths.statusDeploymentGroupHosts(name));
-      }
-      if (client.exists(Paths.statusDeploymentGroup(name)) != null) {
-        client.delete(Paths.statusDeploymentGroup(name));
-      }
+      client.ensurePath(Paths.statusDeploymentGroups());
+      client.transaction(
+          delete(Paths.configDeploymentGroup(name)),
+          delete(Paths.statusDeploymentGroupHosts(name)),
+          delete(Paths.statusDeploymentGroup(name))
+      );
     } catch (final NoNodeException e) {
       throw new DeploymentGroupDoesNotExistException(name);
     } catch (final KeeperException e) {


davidxia:
@mattnworb

davidxia:
One place to log would be here https://github.com/spotify/helios/blob/master/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java#L345, but I don't know if this kind of denormalized data happens enough to have any logging add more signal than noise.

I'd like to add a test, but it seems hard to simulate cases like this in ZooKeeper.

mattnworb:
> If data in ZooKeeper for a deployment group gets denormalized

any idea how the data gets in that state in the first place?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -375,13 +375,15 @@ public void removeDeploymentGroup(final String name) throws DeploymentGroupDoesN
...
-      client.delete(Paths.configDeploymentGroup(name));
+
       if (client.exists(Paths.statusDeploymentGroupHosts(name)) != null) {
         client.delete(Paths.statusDeploymentGroupHosts(name));
       }
       if (client.exists(Paths.statusDeploymentGroup(name)) != null) {
         client.delete(Paths.statusDeploymentGroup(name));
       }
+
+      client.delete(Paths.configDeploymentGroup(name));
```
mattnworb: if the issue was that this path didn't exist (not clear to me if that was the case), why not wrap it in an `if (client.exists(path))` block like the other deletes?

davidxia: Yea, that was my first attempt, but that broke `ZooKeeperMasterModelIntegrationTest. testRemoveNonExistingDeploymentGroup()` which expected `DeploymentGroupDoesNotExistException`: https://circleci.com/gh/spotify/helios/3297

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -375,13 +375,15 @@ public void removeDeploymentGroup(final String name) throws DeploymentGroupDoesN
     final ZooKeeperClient client = provider.get("removeDeploymentGroup");
     try {
       client.ensurePath(Paths.configDeploymentGroups());
-      client.delete(Paths.configDeploymentGroup(name));
+
```
gimaker: I am very confused. It just looks like you're reordering a line. What problem does this fix and how?

davidxia:
@mattnworb I'm not sure. I don't think this can result from a race between a create and a remove DG command because the create command creates `/config/deployment-group/<name>`, `/status/deployment-group/<name>`, and `/status/deployment-group/hosts/<name>` in a transaction. And the remove DG command wouldn't have tried to delete the second two nodes if the first wasn't there. And if the first node was there, that means the other two are as well since they're created in a single transaction.

Maybe a master was restarted during the removal of the radio-spaces DG?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
davidxia: The problem mbetter had was that `/config/deployment-group/<name>` didn't exist but `/status/deployment-group/<name>` and `/status/deployment-group/hosts/<name>`. He couldn't create or remove his DG.

This change will remove all relevant DG ZK nodes even if the `/config/deployment-group/<name>` path doesn't exist.

gimaker:
:-1: I'd rather understand how we get into this state in the first place and do a proper fix. Preferably with tests to verify.

gimaker:
Seems like the config/status/hosts should always exist (provided that the DG does), right? Checking that status & hosts exists seems superfluous. Seems like we should be able to put deletes for all 3 paths in a transaction.

davidxia:
Aye that works

On Wednesday, November 25, 2015, Staffan Gimåker <notifications@github.com>
wrote:

> Seems like the config/status/hosts should always exist (provided that the
> DG does), right? Checking that status & hosts exists seems superfluous.
> Seems like we should be able to put deletes for all 3 paths in a
> transaction.
>
> —
> <https://github.com/spotify/helios/pull/765#issuecomment-159673974>.
>


-- 
David Xia
Software Engineer
dxia@spotify.com

gimaker:
:+1:

mattnworb:
:+1:

mattnworb:
Had a similar issue pop-up today, where a user creating a deployment group failed with 500 errors.

- `/config/deployment-groups/foobar` did not exist
- `/status/deployment-groups/foobar` exists, is empty
- `/status/deployment-group-tasks/foobar` exists, is empty

After removing the last two, creating the DG succeeds.

Makes me wonder if perhaps the issue isn't a race between creating the DG and removing it but if something else caused the data in ZK to get in this state.

PR #767, opened 2015-11-30
fix type error in KAFKA_QUOROM_PARAMETER

mattnworb:
The Kafka `ProducerConfig` expects the value for `ACKS_CONFIG` to be a `String` and not an `int`, even though the value of the String looks to almost always be numeric (seems like it is a String since one possible value is `all` in addition to `1` or `0` or `-1`).

With this configuration error, the agent or master is never actually able to instantiate a KafkaProducer and not able to send any events to Kafka. This was introduced in ae50b2f.

Also added tests to KafkaClientProducer to verify it can produce a Kafka client; refactored the `getProvider()` method a bit as it is always called with the same exact arguments.


mattnworb:
@gimaker @rohansingh

gimaker:
:+1:

gimaker:
checkstyle errors though

mattnworb:
checkstyle errors have been vanquished

PR #769, opened 2015-12-03
Automatically deploy helios-solo for use with helios-testing

negz:
Todo in future PRs:
* Actually plumb the `HeliosDeployment` up up to `TemporaryJobs`. Currently a new instance of `TemporaryJobs` is created per test. We need a way to deploy one `HeliosDeployment` for an entire test run (i.e. multiple classes).
* Propagate DNS for machines using NetworkManager (Java DBUS interface?).

### DIFF
--- a/helios-testing/src/main/java/com/spotify/helios/testing/DockerHost.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/DockerHost.java
@@ -1,0 +1,155 @@
+/*
+ * Copyright (c) 2014 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.helios.servicescommon;
+
+import com.google.common.net.HostAndPort;
+
+import java.net.URI;
+
+import static com.google.common.base.Optional.fromNullable;
+import static com.google.common.base.Strings.isNullOrEmpty;
+import static java.lang.System.getenv;
+
+/**
+ * Represents a dockerd endpoint. A codified DOCKER_HOST.
+ */
+public class DockerHost {
+
+  public static final int DEFAULT_PORT = 2375;
+  public static final String DEFAULT_HOST = "localhost";
+  public static final String DEFAULT_UNIX_ENDPOINT = "unix:///var/run/docker.sock";
+
+  private final String host;
+  private final URI uri;
+  private final String address;
+  private final int port;
+  private final String dockerCertPath;
+
+  private DockerHost(final String endpoint, final String dockerCertPath) {
+    if (endpoint.startsWith("unix://")) {
+      this.port = 0;
+      this.address = "localhost";
+      this.host = endpoint;
+      this.uri = URI.create(endpoint);
+    } else {
+      final String stripped = endpoint.replaceAll(".*://", "");
+      final HostAndPort hostAndPort = HostAndPort.fromString(stripped);
+      final String hostText = hostAndPort.getHostText();
+      final String scheme = isNullOrEmpty(dockerCertPath) ? "http" : "https";
+
+      this.port = hostAndPort.getPortOrDefault(defaultPort());
+      this.address = isNullOrEmpty(hostText) ? DEFAULT_HOST : hostText;
+      this.host = address + ":" + port;
+      this.uri = URI.create(scheme + "://" + address + ":" + port);
+    }
+
+    this.dockerCertPath = dockerCertPath;
+  }
+
+  /**
+   * Get a docker endpoint usable for instantiating a new DockerHost with DockerHost.from(endpoint).
+   *
+   * @return The hostname.
+   */
+  public String host() {
+    return host;
+  }
+
+  /**
+   * Get the docker rest uri.
+   *
+   * @return The uri of the host.
+   */
+  public URI uri() {
+    return uri;
+  }
+
+  /**
+   * Get the docker endpoint port.
+   *
+   * @return The port.
+   */
+  public int port() {
+    return port;
+  }
+
+  /**
+   * Get the docker ip address or hostname.
+   *
+   * @return The ip address or hostname.
+   */
+  public String address() {
+    return address;
+  }
+
+  /**
+   * Get the path to certificate and key for connecting to Docker via HTTPS.
+   *
+   * @return The path to the certificate.
+   */
+  public String dockerCertPath() {
+    return dockerCertPath;
+  }
+
+  /**
+   * Create a {@link DockerHost} from DOCKER_HOST and DOCKER_PORT env vars.
+   *
+   * @return The DockerHost object.
+   */
+  public static DockerHost fromEnv() {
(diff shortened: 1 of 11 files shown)


gimaker:
Can you elaborate a bit on "Propagate DNS for machines using NetworkManager (Java DBUS interface?)"?

mattnworb:
I think you're missing some words here

> Actually plumb the HeliosDeployment up up to `.

negz:
@mattnworb Thanks, done!

@gimaker We should replicate the functionality of https://github.com/spotify/helios/blob/master/solo/helios-up#L56. @rohansingh can likely explain better, but I believe on some machines (Ubuntu desktop?) `/etc/resolv.conf` just points to dnsmasq on localhost, and thus we currently shell out to `nmcli` and pass its DNS servers through to Docker. We should be able to do the same from Java via NetworkManager's DBUS API. What's not clear to me is why we need this workaround for dnsmasq, when we don't work around a similar paradigm for unbound in production.

helios-testing/src/main/java/com/spotify/helios/testing/DockerHost.java
```diff
@@ -0,0 +1,167 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// TODO(negz): Dedupe with com.spotify.helios.servicescommon, move into docker-client?
```
mattnworb: I think docker-client would be a nice destination

helios-testing/src/main/java/com/spotify/helios/testing/HeliosDeployment.java
```diff
@@ -0,0 +1,26 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.helios.testing;
+
+import com.spotify.helios.client.HeliosClient;
+
+public interface HeliosDeployment extends AutoCloseable {
```
mattnworb: probably sounds like a nit, but I think even a one-sentence javadoc summary of the class would be useful

helios-testing/src/main/java/com/spotify/helios/testing/HeliosDeployment.java
```diff
@@ -0,0 +1,26 @@
...
+  HeliosClient client();
```
mattnworb: same here - e.g. I assume this is meant to provide a HeliosClient that points at the "deployment"

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -0,0 +1,390 @@
...
+  final DockerClient dockerClient;
+  final DockerHost dockerHost;
+  final String namespace;
+  final List<String> env;
+  final List<String> binds;
+  final String heliosContainerId;
+  final HeliosClient heliosClient;
+
+  HeliosSoloDeployment(final Builder builder) {
+    final String heliosHost;
```
mattnworb: another nit but prefer to declare vars as close to their point of use as possible, so in this case next to the assignment

mattnworb:
I like the concept of a `HeliosDeployment` which (if my interpretation is correct) represents a deployed cluster somewhere and `HeliosSoloDeployment` points to a deployment of solo.

It seems like `HeliosSoloDeployment` does all of it's work in the constructor; for testing and re-use (such as the addition of other methods to the interface in the future) it might be nice if this was separated out.

Do you have in mind what the use of `HeliosSoloDeployment` by `TemporaryJobs` (i.e. the API to set one up) would look like? Because I think thinking about that can help answer the previous point.

For instance, maybe something like 

``` java
HeliosDeployment solo = new HeliosSoloDeployment();
solo.start();
```

would be a simple way to separate the instantiation of the object from the "now do some work" part of it.

mattnworb:
Ah I missed that the way to build and start a `HeliosSoloDeployment` is `HeliosSoloDeployment.fromEnv().build()` which negates my previous comment a bit.

I think it would be good to have a unit test of HeliosSoloDeployment separate from the IT.

negz:
https://ghe.spotify.net/gist/negz/460c004fdf5816a6ce76

It occurred to me that we could just put the onus on the caller to deploy Helios solo from their tests and pass the resulting client into their TemporaryJobs instance, rather than trying to deploy it magically. Principle of least surprise and all that. Thoughts?

mattnworb:
re the gist - is there teardown work that has to happen to clean up the helios solo container? is the `@Rule` on TemporaryJobs taking care of that?

I think the idea of declaratively using a self-container solo container is intriguing

helios-testing/src/main/java/com/spotify/helios/testing/DockerHost.java
```diff
@@ -0,0 +1,167 @@
...
+    return dockerCertPath;
+  }
+
+  /**
+   * Create a {@link DockerHost} from DOCKER_HOST and DOCKER_PORT env vars.
+   *
+   * @return The DockerHost object.
+   */
+  public static DockerHost fromEnv() {
+    String defaultEndpoint;
```
rohansingh: Should be `final`.

negz: It was like that when I copy pasta'd it from services-common. I'll add a TODO for now and fix it when we copy the implementation into the docker-client in a future PR.

helios-testing/src/main/java/com/spotify/helios/testing/HeliosDeploymentResource.java
```diff
@@ -0,0 +1,52 @@
...
+import com.spotify.helios.client.HeliosClient;
+
+import org.junit.rules.ExternalResource;
+
+/**
+ * A HeliosDeploymentResource makes the supplied {@link HeliosDeployment} available to a JUnit
+ * test, and guarantees to tear it down afterward.
+ */
+public class HeliosDeploymentResource extends ExternalResource {
+  HeliosDeployment deployment;
```
mattnworb: should be `private final`

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -0,0 +1,480 @@
...
+    this.env = containerEnv();
+    this.binds = containerBinds();
+
+    final String heliosHost;
+    final String heliosPort;
+    //TODO(negz): Determine and propagate NetworkManager DNS servers?
+    try {
+      assertDockerReachableFromContainer();
+      if (dockerHost.address().equals("localhost") || dockerHost.address().equals("127.0.0.1")) {
+          heliosHost = containerGateway();
```
mattnworb: super nit: this line is indented different than one below it

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -0,0 +1,480 @@
...
+  private Boolean isBoot2Docker(final Info dockerInfo) {
+    return dockerInfo.operatingSystem().contains(BOOT2DOCKER_SIGNATURE);
+  }
+
+  private Info dockerInfo() {
+    try {
+      return this.dockerClient.info();
+    } catch (DockerException | InterruptedException e) {
+      // There's not a lot we can do if Docker is unreachable.
+      throw new AssertionError(e);
```
mattnworb: `throw RuntimeException(e)` or `throw Throwables.propogate(e)` would be more idiomatic

negz:
Soooo, shall I merge?

mattnworb:
thanks for addressing my comments. :+1:

gimaker:
:+1:

PR #770, opened 2015-12-03
Allow JUnit parallelism for Jenkins builds

rohansingh:
We removed <junit.parallel.threads>1</> for all other build profiles, not
sure why we left it in for Jenkins.


rohansingh:
@davidxia

mattnworb:
:ok:

davidxia:
:+1: If it passes on our jenki

rohansingh:
It didn't actually make things faster.

PR #771, opened 2015-12-03
Fix race conditions that might lead to inconsistent DG state

gimaker:
There were two related race conditions that could cause a DG get
inconsistent state, both triggered by removing the DG. The issue
manifests itself with a DG that has no config node but does have
the status and/or tasks paths.

This could happen if a remove happened at the same time as a
rolling-update was issued or if the host-update thread updated
the hosts for that DG as the DG was removed. The latter is much
more likely to happen.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
@@ -397,11 +397,16 @@ public class ZooKeeperMasterModel implements MasterModel {
       Optional<Integer> version = Optional.absent();
       List<String> curHosts;
       try {
+        // addDeploymentGroup creates Paths.statusDeploymentGroupHosts(name) so it should always
+        // exist. If it doesn't, then the DG was (likely) deleted.
         final Node node = client.getNode(Paths.statusDeploymentGroupHosts(name));
         version = Optional.of(node.getStat().getVersion());
         curHosts = Json.read(node.getBytes(), new TypeReference<List<String>>() {});
-      } catch (NoNodeException | JsonMappingException e) {
+      } catch (JsonMappingException e) {
         curHosts = Collections.emptyList();
+      } catch (NoNodeException e) {
+        // DG was deleted -- abort.
+        return;
       }
 
       if (!version.isPresent() || !hosts.equals(curHosts)) {
@@ -409,8 +414,6 @@ public class ZooKeeperMasterModel implements MasterModel {
         final List<ZooKeeperOperation> ops = Lists.newArrayList();
         ops.add(set(Paths.statusDeploymentGroupHosts(name), Json.asBytes(hosts)));
 
-        client.ensurePath(Paths.statusDeploymentGroup(name));
-        client.ensurePath(Paths.statusDeploymentGroupTasks(name));
         final DeploymentGroup deploymentGroup = getDeploymentGroup(name);
         ImmutableList<Map<String, Object>> events = ImmutableList.of();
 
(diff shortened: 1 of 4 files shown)


gimaker:
@mattnworb @rohansingh @davidxia

helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdateOpFactory.java
```diff
@@ -53,13 +58,24 @@ public RollingUpdateOpFactory(final DeploymentGroupTasks tasks,
...
     events.add(eventFactory.rollingUpdateStarted(deploymentGroup, reason));
 
+    final Stat tasksStat = client.exists(
+        Paths.statusDeploymentGroupTasks(deploymentGroup.getName()));
+    if (tasksStat == null) {
+      // Create the tasks path if it doesn't already exist. The following operations (delete or set)
+      // assume the node already exists. If the tasks path is created/deleted before the transaction
+      // is committed it will fail. This will on occasion generate a user-visible error but is
+      // better than having inconsistent state.
+      ops.add(create(Paths.statusDeploymentGroupTasks(deploymentGroup.getName())));
```
mattnworb: This is a nice fix and makes sense to me. But I'm missing why you had to pass the ZookeeperClient to this class to accomplish that?

helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdateOpFactory.java
```diff
@@ -79,6 +95,9 @@ public RollingUpdateOp start(final DeploymentGroup deploymentGroup,
       ops.add(set(Paths.statusDeploymentGroupTasks(deploymentGroup.getName()), tasks));
     }
 
+    // NOTE: If the DG was removed this set() cause the transaction to fail, because removing
+    // the DG removes this node. It's *important* there's there's an operation that causes the
+    // transaction to fail if the DG was removed or we'll end up with inconsistent state.
```
mattnworb: :+1: for commenting

helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdateOpFactory.java
gimaker: To me having to pass in the client is real ugly, but I couldn't really figure out a better way to do it. We need to pass it in so that we can do a `exists()` to check if the node exists or not. AFAIK there's no way to do a "create-if-it-doesn't-exist" operation in ZK.

gimaker: An alternative way would be to try without the create first and then retry with the create operation if the first transaction fails, but that also seems ugly.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -397,11 +397,16 @@ public void updateDeploymentGroupHosts(final String name, final List<String> hos
       Optional<Integer> version = Optional.absent();
       List<String> curHosts;
       try {
+        // addDeploymentGropup creates Paths.statusDeploymentGroupHosts(name) so it should always
```
davidxia: s/Gropup/Group/

helios-services/src/test/java/com/spotify/helios/rollingupdate/RollingUpdateOpFactoryTest.java
```diff
@@ -64,7 +67,43 @@ public void testStartManualNoHosts() {
...
+        ImmutableSet.copyOf(op.operations()));
+
+    // Two events should return: rollingUpdateStarted and rollingUpdateDone
+    assertEquals(2, op.events().size());
+    verify(eventFactory).rollingUpdateStarted(DEPLOYMENT_GROUP, MANUAL);
+    verify(eventFactory).rollingUpdateDone(DEPLOYMENT_GROUP);
+  }
+
+  @Test
+  public void testStartManualNoHostsTasksAlredyExist() throws Exception {
```
davidxia: s/Alredy/Already/

davidxia:
:100: other than some misspelled words :)

PR #772, opened 2015-12-03
simplify usage of KafkaSender

mattnworb:
- KafkaSender is never null in non-test code, so save some indents by
  not checking for null
- same loop for sending a list of events appears a few times

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
@@ -21,6 +21,7 @@ import com.google.common.base.Function;
 import com.google.common.base.Joiner;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Optional;
+import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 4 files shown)


mattnworb:
This was bugging me.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -623,6 +598,14 @@ public void rollingUpdateStep() {
     }
   }
 
+  private void emitEvents(final List<Map<String, Object>> events) {
```
gimaker: Maybe rename to `emitDeploymentGroupEvents`? It's using a rather specific topic after all.

helios-services/src/test/java/com/spotify/helios/ZooKeeperMasterModelIntegrationTest.java
```diff
@@ -103,8 +106,13 @@ public void setup() throws Exception {
     client.ensurePath(Paths.statusMasters());
     client.ensurePath(Paths.historyJobs());
 
-    model = new ZooKeeperMasterModel(
-        new ZooKeeperClientProvider(client, ZooKeeperModelReporter.noop()));
+    final ZooKeeperClientProvider zkProvider =
+        new ZooKeeperClientProvider(client, ZooKeeperModelReporter.noop());
+
+    final KafkaSender kafkaSender =
```
gimaker: Could just pass in a mocked `KafkaSender` here

PR #773, opened 2015-12-03
simplify usage of KafkaSender

mattnworb:
- KafkaSender is never null in non-test code, so save some indents by
  not checking for null
- same loop for sending a list of events appears a few times


mattnworb:
thanks in advance @gimaker

mattnworb:
@rohansingh @davidxia

gimaker:
:+1: lgtm

PR #775, opened 2015-12-04
Don't return NOT_FOUND when removing a DG fails to due a race

gimaker:
Also remove some ugliness in how the removal was implemented

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
@@ -72,6 +72,7 @@ import org.apache.zookeeper.KeeperException;
 import org.apache.zookeeper.KeeperException.NoNodeException;
 import org.apache.zookeeper.KeeperException.NodeExistsException;
 import org.apache.zookeeper.KeeperException.NotEmptyException;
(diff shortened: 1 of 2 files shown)


mattnworb:
Can test be added to verify the behavior change?

gimaker:
@mattnworb: It's tricky since it's triggered by a race condition, but I'll give it some more thought and see if I can figure out a way to do it. If you have any ideas please share.

mattnworb:
Nothing off of the top of my head, unless you can simply test that the operations you expect to be made are the ones that are made?

I think in general it would be really nice to have better testing and testability of all the ZK code but that shouldn't hold this fix up. :+1:

gimaker:
+:100: for better testing of the ZK code.

gimaker:
@mattnworb Tests added. By using `TestingServer` I think they turned out pretty decent.

mattnworb:
@gimaker nice! :+1: 

The theory framework seems pretty lame if you can only set up `ints` values for the parameters. I've used `org.junit.runners.Parameterized` before which is a bit more extensible.

gimaker:
@mattnworb You can use things other than ints with theories, but it requires extra scaffolding and it didn't seem worth the effort. The nice thing about theories is that it generates all the parameter permutations for you, which AFAIK Parameterized doesn't support.

PR #776, opened 2015-12-04
Ensure that DG tasks are removed when removing a DG

gimaker:
This fixes state inconsistency issues where tasks for a deleted DG
could be left behind. This can cause some bad and unexpected things:
 * Rolling-updates will become slower (proportional to the number of
   stale tasks)
 * We'll spam logs with errors
 * Re-creating the DG will cause things to be deployed even though
   no rolling-update was issued. This is pretty bad.


mattnworb:
:+1:

PR #778, opened 2015-12-08
add debugger support to bin/helios wrapper

mattnworb:
When the `JDWPPORT` environment variable is set, pass arguments to
`java` to tell it to suspend until a debugger is attached.

Often when developing locally it's useful to attach a debugger to the CLI. The helios-agent and helios-master wrappers have this built in so I wanted to add it here too.

### DIFF
--- a/bin/helios
+++ b/bin/helios
@@ -1,4 +1,12 @@
 #!/bin/bash -e
+#
+# Wrapper for running the Helios CLI which will use the locally built `helios`
+# if run from the checked-out source code's root directory.
+#
+# This script will start java with the necessary JDWP arguments to suspend the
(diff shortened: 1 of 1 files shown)


mattnworb:

bin/helios
```diff
@@ -1,4 +1,12 @@
 #!/bin/bash -e
+#
+# Wrapper for running the Helios CLI which will usually the locally built
```
davidxia: are you missing a word in the comment here?

davidxia:
:+1:

bin/helios
mattnworb: Yes, thanks

PR #779, opened 2015-12-08
allow SSL hostname verification to be disabled

mattnworb:
Adds a CLI switch like curl's `-k`/`--insecure` to disable hostname
verification when using HTTPS to communicate with the Helios master.

This can be useful when attempting to connect to a single master instance
using `-z` where the certificate presented by the master has a subject that
doesn't match it's hostname (for example, if the certificate's subject is
`helios.spotify.net` but the server's hostname is
`server1.foo.spotify.net`).

To make the HostnameVerifier logic testable, extracted it's use in
DefaultRequestDispatcher to it's own class.

### DIFF
--- a/helios-tools/src/main/java/com/spotify/helios/cli/Utils.java
+++ b/helios-tools/src/main/java/com/spotify/helios/cli/Utils.java
@@ -54,7 +54,7 @@ public class Utils {
   }
 
   public static HeliosClient getClient(final Target target, final PrintStream err,
-                                       final String username) {
+                                       final String username, final Namespace options) {
 
     List<URI> endpoints = Collections.emptyList();
     try {
@@ -69,6 +69,7 @@ public class Utils {
 
     return HeliosClient.newBuilder()
         .setEndpointSupplier(Endpoints.of(target.getEndpointSupplier()))
+        .setSslHostnameVerification(!options.getBoolean("insecure"))
         .setUser(username)
         .build();
   }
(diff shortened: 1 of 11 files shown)


helios-tools/src/main/java/com/spotify/helios/cli/Utils.java
```diff
@@ -54,7 +54,7 @@
   }
 
   public static HeliosClient getClient(final Target target, final PrintStream err,
-                                       final String username) {
```
mattnworb: allows the CLI options to be passed down to this method for use in constructing the HeliosClient

kind of ugly but I didn't want to add `hostnameVerificationEnabled` as a field to the existing `Target` parameter as it didn't really seem to fit there.

mattnworb:

helios-tools/src/main/java/com/spotify/helios/cli/command/ControlCommand.java
```diff
@@ -118,7 +118,7 @@ private boolean run(final Namespace options, final Target target, final PrintStr
                       final BufferedReader stdin)
       throws InterruptedException, IOException {
 
-    final HeliosClient client = Utils.getClient(target, err, username);
+    final HeliosClient client = Utils.getClient(target, err, username, options);
```
gimaker: I care little about this, but... seems like the current pattern is to pass global options as parameters to the `run()` method (e.g. `username`, `json`). why not just follow that convention? or change it around so everything is accessed via `options`?

mattnworb: Just reluctance to add an 8th parameter to that method, especially if it was only passed to the Command to be passed to `Utils.getClient()`.

mattnworb: I do agree though the Commands could be refactored to have easier access to all these params w/o passing them in every single `run()` though

rohansingh:
:+1:

PR #782, opened 2015-12-09
Ensure we're recording the TLS handshake correctly

rohansingh:
By default, BouncyCastle keeps a rolling hash of TLS handshake messages and
then signs that hash as part of the handshake. Unfortunately, that doesn't
work for us when using ssh-agent, since ssh-agent does the hashing itself.

So instead, we keep a complete record of all TLS handshake messages, and
send that to ssh-agent for hashing and signing.

Since we are keeping this record ourselves, and since only certain types of
messages should be recorded, there's potential for arcane bugs. To prevent
head-scratching, we can compare the SHA1 hash of our record to the rolling
hash that BouncyCastle has kept. The two should match, and if they don't,
then we know our recording is being done incorrectly.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/client/tls/SshAgentTlsClient.java
+++ b/helios-client/src/main/java/com/spotify/helios/client/tls/SshAgentTlsClient.java
@@ -37,7 +37,10 @@ import org.bouncycastle.crypto.tls.TlsSignerCredentials;
 import java.io.IOException;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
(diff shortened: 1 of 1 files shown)


rohansingh:
@mattnworb @gimaker @davidxia

mattnworb:
:+1:

gimaker:
:+1:

rohansingh:
I'm not sure how I managed to increase code coverage with this, but I'll take it.

davidxia:
:+1: @rohansingh ha codecov is probably counting the coverage in a way that doesn't take into account comments?

PR #783, opened 2015-12-09
Use BouncyCastle JAR with debug symbols

rohansingh:
For some reason the default BouncyCastle JAR does not have debug symbols in
it like every other JAR in the world does.


rohansingh:
@mattnworb

mattnworb:
:+1: thank you debug symbols

PR #786, opened 2015-12-10
Log the key ID of generated X509 certificates

rohansingh:


gimaker:
:+1:

helios-client/src/main/java/com/spotify/helios/client/tls/X509CertificateFactory.java
```diff
@@ -83,8 +89,12 @@ public static Certificate get(final AgentProxy agentProxy, final Identity identi
           .get(new AlgorithmIdentifier(OIWObjectIdentifiers.idSHA1));
       final X509ExtensionUtils utils = new X509ExtensionUtils(digestCalculator);
```
mattnworb: can this and the digestCalculator be created just once as a field of the class?

helios-client/src/main/java/com/spotify/helios/client/tls/X509CertificateFactory.java
```diff
@@ -83,8 +89,12 @@ public static Certificate get(final AgentProxy agentProxy, final Identity identi
...
 
-      builder.addExtension(Extension.subjectKeyIdentifier, false,
-                           utils.createSubjectKeyIdentifier(subjectPublicKeyInfo));
+      final SubjectKeyIdentifier keyId = utils.createSubjectKeyIdentifier(subjectPublicKeyInfo);
+      final String keyIdHex = BaseEncoding.base16().upperCase().withSeparator(":", 2)
```
mattnworb: `BaseEncoding.base16().upperCase().withSeparator(":", 2)` could be a field too

mattnworb:
nice find :+1:

helios-client/src/main/java/com/spotify/helios/client/tls/X509CertificateFactory.java
rohansingh: Unfortunately no since the creation of the `DigestCalculator` can throw an exception, and that doesn't play nice with static constructors.

PR #787, opened 2015-12-10
Add support for X509 client cert & key files

rohansingh:
As an alternative to generating an X509 certificate from the public keys in
ssh-agent, we now support supplying an X509 client certificate and key for
TLS authentication.


rohansingh:
@mattnworb @gimaker @davidxia @negz

gimaker:
:+1:

gimaker:
Fix the build, though (checkstyle errors)

rohansingh:
Guess I'm just not stylish enough.

mattnworb:
Does the stuff in `CertificateFileHttpsHandler` with loading a Keystore instance load the user's keystore from disk? Why is the password hardcoded? How come we have to add something to the keystore to load the certificate?

helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
```diff
@@ -502,10 +505,22 @@ public static Builder newBuilder() {
...
+    private Path clientCertificatePath;
+    private Path clientKeyPath;
     private Supplier<List<Endpoint>> endpointSupplier;
     private boolean sslHostnameVerification = true;
 
+    private Builder() {
+      final String heliosCertPath = System.getenv(HELIOS_CERT_PATH);
+      if (!isNullOrEmpty(heliosCertPath)) {
+        this.clientCertificatePath = Paths.get(heliosCertPath, "cert.pem");
+        this.clientKeyPath = Paths.get(heliosCertPath, "key.pem");
```
mattnworb: are these standard names for these files? You could also just set one env var for each path.

rohansingh:
@mattnworb Nope, it doesn't load a keystore from disk. That would be the Java(tm) way to do it, but nobody keeps their certs and keys in Java keystores so it would be a pain in the ass.

Instead, we create a keystore in memory and put the certificate and key from disk into that keystore. The keystore needs a password when we put the key into it, even though it's only going to exist for the lifetime of the process. So we just have some random password that we use. Most of this code is lifted from spotify/docker-client.

helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
rohansingh: I stuck to these since it's how `DOCKER_CERT_PATH` works.

mattnworb:
@rohansingh ah cool. I think you could stick that whole second paragraph in the source as comments and it'd be :ok_hand:

helios-client/src/main/java/com/spotify/helios/client/AuthenticatingHttpConnector.java
```diff
@@ -63,19 +66,31 @@
...
 
   @VisibleForTesting
   AuthenticatingHttpConnector(final String user,
                               final Optional<AgentProxy> agentProxyOpt,
+                              final Optional<Path> clientCertificatePath,
+                              final Optional<Path> clientKeyPath,
                               final EndpointIterator endpointIterator,
                               final DefaultHttpConnector delegate,
                               final List<Identity> identities) {
+    if (clientCertificatePath.isPresent() ^ clientKeyPath.isPresent()) {
```
mattnworb: u fancy

PR #792, opened 2015-12-15
fix infinite loop if masters return 502 Bad Gateway

mattnworb:
AuthenticatingHttpConnector handles a 502 Bad Gateway response by just
trying another iteration of the `while (true)`.

If for some reason the endpoints always return 502 (for example, the
service behind nginx is down) then this means the
AuthenticatingHttpConnector will loop forever (as long as the endpoints
are returning 502) with no way to interrupt.

This undos a change in 27d3c25 where a 502 caused the http client to
just try again with a new endpoint (and which was carried over to the
HttpConnector stuff). This is not needed anymore since
RetryingRequestDispatcher was introduced in 1c2c451.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/client/AuthenticatingHttpConnector.java
+++ b/helios-client/src/main/java/com/spotify/helios/client/AuthenticatingHttpConnector.java
@@ -42,7 +42,6 @@ import java.util.Deque;
 import java.util.List;
 import java.util.Map;
 
-import static java.net.HttpURLConnection.HTTP_BAD_GATEWAY;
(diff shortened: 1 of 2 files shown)


mattnworb:
The potential infinite loop is because when `continue`ing the loop the next iterations don't fail if the `identities.poll()` returns null because no more identities are left.

mattnworb:

mattnworb:
hmm RetryingRequestDispatcher doesn't retry in this case, this needs some work.

gimaker:
@mattnworb that's because it get's a response. it doesn't inspect the http code and retry based on those

PR #793, opened 2015-12-16
fix infinite loop if masters return 502 Bad Gateway

mattnworb:
AuthenticatingHttpConnector handles a 502 Bad Gateway response by just
trying another iteration of the `while (true)`.

If for some reason the endpoints always return 502 (for example, the
service behind nginx is down) then this means the
AuthenticatingHttpConnector will loop forever (as long as the endpoints are
returning 502) with no way to interrupt.

This check for 502 Bad Gateway is better made in the DefaultHttpConnector
as it has nothing to do with authentication. A response with this status
code will have an exception thrown so that it is caught by
RetryingRequestDispatcher (and a new request is scheduled).

Also: broke out the separate "authentication flows" in
AuthenticatingHttpConnector to make it easier to follow the various flows for
different authentication types and avoid the hairy `while true` loop.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/client/AuthenticatingHttpConnector.java
+++ b/helios-client/src/main/java/com/spotify/helios/client/AuthenticatingHttpConnector.java
@@ -20,8 +20,8 @@ package com.spotify.helios.client;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Optional;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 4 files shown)


mattnworb:
This is a better version of #792

mattnworb:
...and the integration tests finally pass after retrying the build 3 times in Circle.

gimaker:
:+1:

rohansingh:
Yeah this makes a lot more sense.

:+1:

PR #794, opened 2015-12-16
enhancements for HeliosSoloDeployment and tests

mattnworb:
1) wait for Solo to start before running test

Add a wait in the HeliosSoloDeployment TestRule to make sure that it can
connect to the HeliosSolo instance before continuing with the test that the
TestRule is used in.

This prevents a bunch of error messages in the test itself coming from the
HeliosClient that it could not connect.

2) log more details about HeliosSoloDeployment to help diagnosing test failures


mattnworb:
@rohansingh cc @negz

rohansingh:
:+1:

PR #799, opened 2015-12-17
Use signed certificates for ssh-agent auth

rohansingh:
Previously, we were overriding a bunch of things to let the ssh-agent sign
the TLS handshake on-the-fly. While this works, it's a lot of code and also
required us to duplicate/override parts of the TLS internals.

Instead of doing that, we now just generate a new key pair and certificate
and have the ssh-agent sign that certificate. This is much less code, much
simpler, and is actually compatible with our existing infrastructure.

It also consolidates a lot of the code for the ssh-agent and certificate
file options, since both just generate a java.security.cert.Certificate and
a java.security.PrivateKey which is handed off to the standard
SSLSocketFactory implementation.


rohansingh:
@mattnworb @gimaker @davidxia

helios-client/src/main/java/com/spotify/helios/client/AuthenticatingHttpConnector.java
```diff
@@ -216,7 +215,7 @@ private URI toIpUri(Endpoint endpoint, URI uri) throws URISyntaxException {
       } catch (Exception e) {
         // We catch everything because right now the masters do not require authentication.
         // So delay reporting errors to the user until the servers return 401 Unauthorized.
-        log.debug("Couldn't get identities from ssh-agent", e);
+        log.debug("Couldn't get identities from tls-agent", e);
```
mattnworb: does anyone call the process `tls-agent` instead of `ssh-agent`?

mattnworb:
Nice :+1: :100:

mattnworb:
I think this is the largest bump I've seen from codecov messages so far

> Merging #799 into master will increase coverage by +0.76% as of 9c0e48d

helios-client/src/main/java/com/spotify/helios/client/AuthenticatingHttpConnector.java
```diff
@@ -112,12 +112,11 @@ public HttpURLConnection connect(final URI uri, final String method, final byte[
...
 
-      // prioritize using the certificate file if set
       if (clientCertificatePath.isPresent() && clientKeyPath.isPresent()) {
+        // prioritize using the certificate file if set
         return connectWithCertificateFile(ipUri, method, entity, headers);
-      }
-      // ssh-agent based authentication
-      else if (agentProxy.isPresent() && !identities.isEmpty()) {
+      } else if (agentProxy.isPresent() && !identities.isEmpty()) {
+        // tls-agent based authentication
```
gimaker: what is tls-agent?

rohansingh: Ha. Damn IntelliJ smart refactoring.

helios-client/src/main/java/com/spotify/helios/client/AuthenticatingHttpConnector.java
rohansingh: Nope, just a run-away IntelliJ refactor.

davidxia:
image
:+1:

PR #800, opened 2015-12-22
Make system test logs in CircleCI easier to read

mattnworb:
... by moving the logs from system tests to a file instead.

Since the tests are run in parallel inside the container, the log output is interleaved so it's very hard or impossible to determine which log lines came from which test.

By moving these to a file and copying them to the artifact/test-report directory at at the end, the log files for these tests are now available under the "Artifacts" tab. 

image

We can also link to the log files directly, [for example](https://circle-artifacts.com/gh/spotify/helios/3483/artifacts/1/tmp/circle-junit.oRvwP5m/20151222T205423.903-com.spotify.helios.system.DeploymentGroupTest_testRollingUpdateWithOverlapAndParallelism-10893.log)


mattnworb:

davidxia:
Seems pretty useful to me :+1:

Issue #762, opened 2015-11-24
Confusing error message when deploying to incorrectly formatted hostname


mattnworb:
If the hostname passed to `helios deploy` looks like a URL (or simply contains a slash) when the `--json` switch is present, a confusing error message `Conversion = 'F'` is printed with no other output.

```
# creating the job
$ bin/helios -z http://localhost:5801 create testjob:1 some-image
Creating job: {"command":[],"created":null,"creatingUser":null,"env":{},"expires":null,"gracePeriod":null,"healthCheck":null,"hostname":null,"id":"testjob:1","image":"some-image","metadata":{},"networkMode":null,"ports":{},"registration":{},"registrationDomain":"","resources":null,"securityOpt":[],"token":"","volumes":{}}
Done.
testjob:1:2178b49ac982e92b87442b179b869b4414f770db

# regular deploy to a non-existent host
$ bin/helios --json -z http://localhost:5801 deploy testjob:1 some-host
{"host":"some-host","job":"testjob:1:2178b49ac982e92b87442b179b869b4414f770db","status":"HOST_NOT_FOUND"}

# deploy to host "foo/bar"
$ bin/helios --json -z http://localhost:5801 deploy testjob:1 foo/bar
Conversion = 'F'
```

To reduce any possible user confusion it would be nice if the `HOST_NOT_FOUND` was still output.

The cause is that the slash is encoded as `%2F` in the JSON response from the server and then this string is passed through `printf` / `java.util.Formatter`:

```
$ bin/helios -vvv --json -z http://localhost:5801 deploy testjob:1 foo/bar
...
10:11:25.987 TRACE DefaultRequestDispatcher: rep: PUT http://127.0.0.1:5801/hosts/foo%252Fbar/jobs/testjob:1:2178b49ac982e92b87442b179b869b4414f770db?token=&user=mattbrown 400 105 {
  "host" : "foo%2Fbar",
  "job" : "testjob:1:2178b49ac982e92b87442b179b869b4414f770db",
  "status" : "HOST_NOT_FOUND"
} gzip:false
java.util.UnknownFormatConversionException: Conversion = 'F'
	at java.util.Formatter$FormatSpecifier.conversion(Formatter.java:2691)
	at java.util.Formatter$FormatSpecifier.<init>(Formatter.java:2720)
	at java.util.Formatter.parse(Formatter.java:2560)
	at java.util.Formatter.format(Formatter.java:2501)
	at java.io.PrintStream.format(PrintStream.java:970)
[...]

PR #801, opened 2015-12-26
Do not use printf to avoid formatting errors

davidxia:
We don't need to use printf() when we don't have any args.


davidxia:
This helps with #762. Hostnames like `foo/bar` show up in the JSON as `foo%2Fbar` but at least the command doesn't have a strange error message. Perhaps we should replace `printf()` with `print()` in other places where there are no args just to be safe?

gimaker:
:+1:

davidxia:
I've converted more unecessary `printf` to `print`

gimaker:
:+1:

mattnworb:
+1 for not using `printf` when the String argument is known not to contain any string formatting patterns.

Why would `foo/bar` be output as `foo%2Fbar` though?

davidxia:
The url encoding happens in HeliosClient

PR #803, opened 2016-01-05
Add functionality to use ZooKeeper ACLs. Disabled by default

gimaker:
This stops (or at least impedes) an attacker with network access to ZK
from injecting data into ZK, e.g. injecting data causing a malicious job
to be deployed on some or all agents.

Everyone, including unauthenticated users, has read access to all data.
For futher permissions credentials are needed. Agents and masters use
different credentials, limiting the consequences if the agent credentials
are compromised.


gimaker:
@rohansingh @mattnworb @davidxia 

This is still WIP. There stuff left to do:
 * Documentation (?)
 * Tests to make sure it all works
 * Ideally, we also set it up so that at least some integration tests run with ZK ACLs disabled, to verify that that still works as it should
 * ???

helios-services/src/main/java/com/spotify/helios/agent/AgentParser.java
```diff
@@ -217,6 +230,15 @@ protected void addArgs(final ArgumentParser parser) {
               + "registration ID of this one can automatically deregister this one and register "
               + "itself. This is useful when the agent loses its registration ID and you don't "
               + "want to waste time debugging why the master lists your agent as constantly DOWN.");
+
+    zkMasterDigest = parser.addArgument("--zk-master-digest")
+        .type(String.class);
```
mattnworb: should add a `help(..)` here too

helios-services/src/main/java/com/spotify/helios/agent/AgentService.java
```diff
@@ -320,13 +326,39 @@ public AgentService(final AgentConfig config, final Environment environment)
...
+      if (agentPassword == null) {
+        throw new HeliosRuntimeException("ZooKeeper ACLs enabled but agent password not set");
+      }
+
+      if (masterDigest == null) {
+        throw new HeliosRuntimeException("ZooKeeper ACLs enabled but master digest not set");
+      }
+
+      final String agentDigest = Base64.toBase64String(Hash.sha1digest(
+          String.format("%s:%s", ZooKeeperAclProvider.AGENT_USER, agentPassword).getBytes()));
```
mattnworb: why perform the calculation for the "agent digest" here, but have the master digest passed in as an argument? Why not simply pass both?

helios-services/src/main/java/com/spotify/helios/master/MasterService.java
```diff
@@ -308,13 +316,40 @@ private void logBanner() {
    * @return A zookeeper client.
    */
   private ZooKeeperClient setupZookeeperClient(final MasterConfig config) {
+    ACLProvider aclProvider = null;
+    List<AuthInfo> authorization = null;
+
+    if (config.isZooKeeperEnableAcls()) {
```
mattnworb: this block is repeated in AgentService and MasterService, just with a reversal of the labels, maybe it can be moved into the CuratorClientFactory?

helios-services/src/main/java/com/spotify/helios/servicescommon/ZooKeeperAclProvider.java
```diff
@@ -0,0 +1,161 @@
...
+    this.defaultAcl = ImmutableList.of(new ACL(READ, ZooDefs.Ids.ANYONE_ID_UNSAFE));
+
+    // Master has CRWD permissions on all paths
+    rule(".*", CREATE | READ | WRITE | DELETE, masterId);
+    // Agent has READ permission on all paths
+    rule(".*", READ, agentId);
+    // Grant everyone READ-only access to make troubleshooting easier
+    rule(".*", READ, ZooDefs.Ids.ANYONE_ID_UNSAFE);
+
+    // The agent needs to be create the /config/hosts/<host> path and nodes below it. It also needs
```
mattnworb: typo: "needs to ~~be~~ create"

helios-services/src/main/java/com/spotify/helios/servicescommon/ZooKeeperAclProvider.java
```diff
@@ -0,0 +1,161 @@
...
+ *   <li>"/foo/bar", CREATE, user1</li>
+ * </ol>
+ * Then user1 can create and delete nodes on the /foo/bar node, but he can not delete nodes under
+ * the /foo node.
+ * <p>
+ * The provider uses ZooKeeper's "digest" ACL scheme.
+ */
+public class ZooKeeperAclProvider implements ACLProvider {
+
+  private static final String WILDCARD = "[^/]+";
```
mattnworb: maybe call this something like PATH_COMPONENT_WILDCARD since it's not a true wildcard like `.*`

mattnworb:
the stuff with the list of Rules in ZookeeperACLProvider is nice :ok_hand: 

but it feels weird to me to hardcode the list of rules in the constructor as oppose to passing them in - makes the other functionality of that class (scanning the list of rules and OR-ing things together) harder to unit test since you can't test it with different variations of rules. I would suggest moving the hard-coding up a level so you can test the class.

helios-services/src/main/java/com/spotify/helios/agent/AgentService.java
gimaker: Pass master-digest, agent-digest and the agent password? You could, but it's redundant since the agent-digest can be computed from the agent password.

gimaker:
@mattnworb yeah that [moving the list of rules out of the ctor] probably makes sense. it bugged me too -- but I wasn't  sure the extra boilerplate needed would be worth it.

(If you just want to be able to test with different rules, a quick and dirty approach is to create an auxiliary `getAclForPath(String path, List<Rule> rules)` method and make it package-local)

helios-services/src/main/java/com/spotify/helios/servicescommon/ZooKeeperAclProvider.java
gimaker: That is so frigging long though :disappointed:

mattnworb: Yeah. A comment mentioning that it is meant to match a single path segment would probably suffice.

helios-services/src/main/java/com/spotify/helios/agent/AgentService.java
mattnworb: hm I guess it wasn't clear to me at first read that the agent password is used in both the digest and authentication, and same for the master. I bet that would be cleared up with a nice new `docs/security.md` file though :wink:

mattnworb:
@gimaker or just add an overloaded constructor like 

```
public ZooKeeperAclProvider(final String masterDigest, final String agentDigest) {
  this(masterDigest, agentDigest, DEFAULT_RULES);
}
public ZooKeeperAclProvider(final String masterDigest, final String agentDigest, List<Rule> rules) {
   ...
}

**Comment by gimaker** [comment]
@mattnworb Yeah, supplying the rules in the ctor is what I'm thinking of doing. But it requires extra scaffolding/refactoring since the rules contain the digests as it is now (in the `Id` object).

File: helios-services/src/main/java/com/spotify/helios/agent/AgentParser.java [hunk 6]
```diff
@@ -97,6 +101,12 @@ public AgentParser(final String... args) throws ArgumentParserException {
       throw new IllegalArgumentException("Bad port range: " + portRangeString);
     }
 
+
+    String agentPassword = System.getenv(ZK_AGENT_PASSWORD_ENVVAR);
+    if (agentPassword == null) {
```
rohansingh: `isNullOrEmpty` ?

helios-services/src/main/java/com/spotify/helios/agent/AgentService.java
```diff
@@ -320,13 +326,39 @@ public AgentService(final AgentConfig config, final Environment environment)
...
   private ZooKeeperClient setupZookeeperClient(final AgentConfig config, final String id,
                                                final CountDownLatch zkRegistrationSignal) {
+    ACLProvider aclProvider = null;
+    List<AuthInfo> authorization = null;
+
+    if (config.isZooKeeperEnableAcls()) {
+      final String agentPassword = config.getZooKeeperAgentPassword();
+      final String masterDigest = config.getZooKeeperMasterDigest();
+
+      if (agentPassword == null) {
```
rohansingh: `isNullOrEmpty`

helios-services/src/main/java/com/spotify/helios/master/MasterService.java
gimaker: If you have a good idea of how to do it, please share. I can move it to another place to reduce clutter, but I can't really figure out a great way to reduce duplication.

helios-services/src/main/java/com/spotify/helios/agent/AgentParser.java
gimaker: An empty password is valid, no? Stupid, but valid.

helios-services/src/main/java/com/spotify/helios/agent/AgentService.java
rohansingh: I'd replace this with `org.apache.zookeeper.server.auth.DigestAuthenticationProvider.generateDigest()`.

helios-services/src/main/java/com/spotify/helios/master/MasterParser.java
```diff
@@ -42,6 +46,11 @@ public MasterParser(final String... args) throws ArgumentParserException {
     final Namespace options = getNamespace();
     final InetSocketAddress httpAddress = parseSocketAddress(options.getString(httpArg.getDest()));
 
+    String masterPassword = System.getenv(ZK_MASTER_PASSWORD_ENVVAR);
+    if (masterPassword == null) {
```
rohansingh: `isNullOrEmpty`

helios-services/src/main/java/com/spotify/helios/master/MasterService.java
```diff
@@ -308,13 +316,40 @@ private void logBanner() {
...
+
+      if (masterPassword == null) {
+        throw new HeliosRuntimeException("ZooKeeper ACLs enabled but master password not set");
+      }
+
+      if (agentDigest == null) {
+        throw new HeliosRuntimeException("ZooKeeper ACLs enabled but agent digest not set");
+      }
+
+      final String masterDigest = Base64.toBase64String(Hash.sha1digest(
```
rohansingh: Same here on using `DigestAuthenticationProvider.generateDigest` instead of doing it ourselves.

gimaker: It's an annoying function -- it throws a [checked] `java.security.NoSuchAlgorithmException` exception that has to be dealt with. I can use it if you feel strongly about it, but if so I'll probably wrap it in a function that re-throws that exception as a runtime exception.

davidxia:
womp womp czechstyle

rohansingh:
This is looking good to me at this point.

@davidxia @mattnworb Could you please review?
@gimaker Was there anything more you wanted to add or do?

gimaker:
@rohansingh nope, not really. Ideally it'd be good to have some tests that run without ACLs to make sure things still work when they're disabled, but maybe that's not needed? Other than that, the only thing that comes to mind is if we want to give agents more permissions on everything under status - e.g. give the CRWD on everything under `/status/hosts/<host>`, to avoid problems if we ever need more permissions.

docs/zk_security.md
```diff
@@ -0,0 +1,56 @@
+# ZooKeeper security
+
+To prevent accidental or malicious tampering with the data that Helios stores in ZooKeeper Helios
```
mattnworb: grammar nit: should be a comma in `... with the data that Helios stores in Zookeeper, Helios...`

docs/zk_security.md
```diff
@@ -0,0 +1,56 @@
...
+supports using ZooKeeper's ACL functionality to lock it down (to an extent).  This functionality is
+disabled by default.
+
+When enabled, credentials are required to access data in ZooKeeper (we use ZooKeeper's `digest`
+authentication scheme). There are two sets of credentials: one set of the masters, and one set for
```
mattnworb: should be "one set _for_ the masters"

docs/zk_security.md
```diff
@@ -0,0 +1,56 @@
...
+even read access). Agents have read access to all data but limited permissions to mutate data. They
+only have mutate permissions where needed. This limits the impact of the agent credentials being
+compromised (e.g. if an agent is compromised). While the agent credentials being compromised is
+obviously not good the lower privileges they provide prevent an attacker from deploying malicious
+jobs to other agents (in most cases).
+
+Masters are granted all permissions except ADMIN on all nodes.
+
+**Note that credentials are sent to ZooKeeper in plain-text** (we currently don't support using
+encrypted connections to ZooKeeper) - meaing that ACLs are only effective if you can trust the
```
mattnworb: "meaing" => "meaning"

docs/zk_security.md
```diff
@@ -0,0 +1,56 @@
...
+the agents. With ACLs enabled unauthenticated users have no access to the data in ZooKeeper (not
+even read access). Agents have read access to all data but limited permissions to mutate data. They
+only have mutate permissions where needed. This limits the impact of the agent credentials being
+compromised (e.g. if an agent is compromised). While the agent credentials being compromised is
+obviously not good the lower privileges they provide prevent an attacker from deploying malicious
+jobs to other agents (in most cases).
+
+Masters are granted all permissions except ADMIN on all nodes.
+
+**Note that credentials are sent to ZooKeeper in plain-text** (we currently don't support using
```
mattnworb: Is this something that a Helios cluster admin could configure separately? For example, configure the zookeeper connection string to use SSL or whatever they call it. Are there code changes necessary in Helios to support encrypted connections to Zookeeper?

gimaker:
Another thing that might be prudent to do is to run all tests and examine the logs for `NoAuth` exceptions (that may be problematic even though they don't cause tests to fail?).

docs/zk_security.md
```diff
@@ -0,0 +1,56 @@
...
+
+    $ echo -n user:password | openssl dgst -sha1 -binary | base64
+    tpUq/4Pn5A64fVZyQ0gOJ8ZWqkY=
+
+## Migrating an existing cluster to using ACLs
+
+ACLs are applied only to new nodes as they are created. Enabling ACL support on an existing cluster
+will not break anything but ACLs will not be retroactively applied to already existing ZK nodes.
+
+TODO
```
mattnworb: add a link here to `bin/helios-initialize-acl` ?

docs/zk_security.md
gimaker: Not 100% sure, but it looks like it's possible without code changes: https://cwiki.apache.org/confluence/display/ZOOKEEPER/ZooKeeper+SSL+User+Guide

helios-services/src/main/java/com/spotify/helios/servicescommon/ZooKeeperAclInitializer.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.curator.retry.ExponentialBackoffRetry;
+import org.apache.zookeeper.KeeperException;
+
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import static com.spotify.helios.servicescommon.ZooKeeperAclProviders.digest;
+import static com.spotify.helios.servicescommon.ZooKeeperAclProviders.heliosAclProvider;
+
+public class ZooKeeperAclInitializer {
```
mattnworb: can we call this (and the shell script) `Migrator` instead of `Initializer`?

helios-services/src/main/java/com/spotify/helios/master/MasterService.java
```diff
@@ -308,19 +318,68 @@ private void logBanner() {
...
+      // a new ZK cluster.
+      // Note that this is slightly racey -- if two masters start at the same time both might
+      // attempt to update the ACLs but only one will succeed. That said, it's unlikely and the
+      // effects are limited to a spurious log line.
+      try {
+        final List<ACL> curAcls = client.getAcl("/");
+        final List<ACL> wantedAcls = aclProvider.getAclForPath("/");
+        if (!Sets.newHashSet(curAcls).equals(Sets.newHashSet(wantedAcls))) {
+          log.info(
+              "Current ACLs on the zookeeper root node differ from the desired ACLs, updating");
```
mattnworb: would be good to include what those different values are in the log message

helios-services/src/main/java/com/spotify/helios/servicescommon/RuleBasedZooKeeperAclProvider.java
```diff
@@ -0,0 +1,131 @@
...
+      acls.add(new ACL(e.getValue(), e.getKey()));
+    }
+
+    return acls;
+  }
+
+  private static class Rule {
+
+    private final Pattern pattern;
+    private Id id;
```
mattnworb: `id` should be final too

helios-services/src/main/java/com/spotify/helios/servicescommon/ZooKeeperAclInitializer.java
mattnworb: because then you wouldn't need to document what the class / script does, as `migrator` is pretty self explanatory, and I was initially confused what being `initialized` meant :smile:

helios-services/src/main/java/com/spotify/helios/servicescommon/ZooKeeperAclInitializer.java
```diff
@@ -0,0 +1,78 @@
...
+
+  public static void main(final String[] args) throws Exception {
+    if (args.length == 6) {
+      initializeAcl(args[0], args[1], args[2], args[3], args[4], args[5]);
+    } else {
+      System.out.println("usage: <ZK connect string> <ZK cluster ID> " +
+                         "<master user> <master password> <agent user> <agent password>");
```
mattnworb: be a good citizen and call System.exit too :wink:

helios-services/src/main/java/com/spotify/helios/servicescommon/ZooKeeperAclInitializer.java
```diff
@@ -0,0 +1,78 @@
...
+        (int) TimeUnit.SECONDS.toMillis(60),
+        (int) TimeUnit.SECONDS.toMillis(15),
+        zooKeeperRetryPolicy,
+        aclProvider,
+        authorization);
+
+    final ZooKeeperClient client = new DefaultZooKeeperClient(curator, zooKeeperClusterId);
+    client.start();
+
+    client.initializeAclRecursive("/", aclProvider);
```
mattnworb: shouldn't you close the client when it is done?

```
try (ZooKeeperClient client = new DefaultZooKeeperClient(curator, zooKeeperClusterId)) {
 client.start();
 client.initializeAclRecursive("/", aclProvider);
}
```

helios-services/src/main/java/com/spotify/helios/servicescommon/ZooKeeperAclProviders.java
```diff
@@ -0,0 +1,99 @@
...
+import org.apache.zookeeper.data.Id;
+import org.bouncycastle.util.encoders.Base64;
+
+import static org.apache.zookeeper.ZooDefs.Perms.CREATE;
+import static org.apache.zookeeper.ZooDefs.Perms.READ;
+import static org.apache.zookeeper.ZooDefs.Perms.WRITE;
+import static org.apache.zookeeper.ZooDefs.Perms.DELETE;
+
+/**
+ * Contains a single factory method for creating ACLProvider used by Helios (agents and masters),
```
mattnworb: the "single" part of this comment is no longer true

helios-services/src/main/java/com/spotify/helios/servicescommon/ZooKeeperAclProviders.java
```diff
@@ -0,0 +1,99 @@
...
+ * that controls the ACLs set on nodes when they are created.
+ * <p>
+ * If not authentication is provided a client has no access (not even read access). There are two
```
mattnworb: "If _no_ authentication is provided"

helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/CuratorClientFactory.java
```diff
@@ -18,14 +18,19 @@
...
-    public CuratorFramework newClient(String connectString,
-                                      int sessionTimeoutMs,
-                                      int connectionTimeoutMs,
-                                      RetryPolicy retryPolicy,
-                                      String namespace);
+  public CuratorFramework newClient(String connectString,
+                                    int sessionTimeoutMs,
+                                    int connectionTimeoutMs,
+                                    RetryPolicy retryPolicy,
+                                    ACLProvider aclProvider,
```
mattnworb: how about `Optional<ACLProvider>` ?

helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/DefaultZooKeeperClient.java
```diff
@@ -108,23 +112,25 @@ public CuratorFramework getCuratorFramework() {
...
-    if (excludingLast) {
-      ensurePath = ensurePath.excludingLast();
-    }
-    try {
-      ensurePath.ensure(client.getZookeeperClient());
-    } catch (Exception e) {
-      propagateIfInstanceOf(e, KeeperException.class);
-      throw propagate(e);
+
+    final String[] parts = path.substring(1).split(Pattern.quote("/"));
```
mattnworb: curious: why do you need to use `.split(Pattern.quote("/"))` instead of `.split('/')`?

helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/DefaultZooKeeperClient.java
```diff
@@ -159,7 +165,9 @@ public Stat exists(final String path) throws KeeperException {
 
   @Override
   public void start() {
-    client.start();
+    if (client.getState() != CuratorFrameworkState.STARTED) {
```
mattnworb: is `start()` really called more than once? 

If it is - shouldn't the lines below for adding connectionStateListener etc also be in this block?

helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/DefaultZooKeeperClient.java
```diff
@@ -389,6 +395,51 @@ public Stat stat(final String path) throws KeeperException {
...
+
+  @Override
+  public void initializeAclRecursive(final String path, final ACLProvider aclProvider)
+      throws KeeperException {
+    assertClusterIdFlagTrue();
+    try {
+      final List<ACL> expected = aclProvider.getAclForPath(path);
+      final List<ACL> actual = getAcl(path);
+
+      if ((expected.size() == actual.size()) && (expected.containsAll(actual))) {
```
mattnworb: elsewhere in this diff two `List<ACL>`s are compared by converting both to sets and calling `.equals()`, should do the same here for consistency

helios-services/src/test/java/com/spotify/helios/servicescommon/RuleBasedZooKeeperAclProviderTest.java
```diff
@@ -0,0 +1,109 @@
...
+    final Id id2 = new Id("some_scheme", "id2");
+    final RuleBasedZooKeeperAclProvider aclProvider = RuleBasedZooKeeperAclProvider.builder()
+        .rule("/foo/baz", DELETE, id1)
+        .rule("/foo/bar", CREATE, id1)
+        .rule("/foo/qux", READ | WRITE, id2)
+        .build();
+
+    assertEquals(
+        Arrays.asList(new ACL(DELETE, id1)),
+        aclProvider.getAclForPath("/foo/baz"));
```
mattnworb: pretty minor but you can also use `assertThat(aclProvider.getAclForPath("/foo/baz"), contains(new ACL(DELETE, id1)))` might be less verbose

mattnworb: in case it's not clear (and this always confuses me) but hamcrest's `contains(..)` method tests not just that the item is in the collection but that the item contains the arguments in that order when iterated over

helios-services/src/test/java/com/spotify/helios/servicescommon/RuleBasedZooKeeperAclProviderTest.java
```diff
@@ -0,0 +1,109 @@
...
+    final RuleBasedZooKeeperAclProvider aclProvider = RuleBasedZooKeeperAclProvider.builder()
+        .rule("/foo.*", DELETE, id1)
+        .rule("/foo/bar", CREATE, id1)
+        .rule(".*", READ, id2)
+        .rule("/foo/bar/baz", WRITE, id2)
+        .build();
+
+    assertEquals(
+        Sets.newHashSet(new ACL(CREATE | DELETE, id1), new ACL(READ, id2)),
+        Sets.newHashSet(aclProvider.getAclForPath("/foo/bar")));
```
mattnworb: same comment - can also be

    assertThat(aclProvider.getAclForPath("/foo/bar"), containsInAnyOrder(...))

docs/zk_security.md
mattnworb: Cool, so I think it would be good to link to that page here and change the doc to say that when not using SSL the credentials are sent in plaintext, and if you want to use SSL read this link

mattnworb:
:+1: from me, nice work

helios-services/src/main/java/com/spotify/helios/servicescommon/ZooKeeperAclInitializer.java
rohansingh: I went with `Initializer` because then the bin script gets them name `helios-initialize-acl`. If we go with migrate, then the script ends up being `helios-migrate-acl` or `helios-acl-migrate` or something which I think is less clear.

But, I'm also kind of indifferent. What do you think?

mattnworb: Ha, I guess it is subjective because I found `migrate` more clear than `initialize`. In either case I think it is now clear to all involved :+1:

helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/CuratorClientFactory.java
rohansingh: Why? We'd still have to check if the optional is null and then check if it contains a value. Right now we just check if the object is null.

helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/DefaultZooKeeperClient.java
rohansingh: You can't split on chars, only on regular expressions.

mattnworb: the single-quote there was a mistake - my question was why (or is it?) is it necessary to quote a `/`?

rohansingh: It really isn't. I just always do `split(quote("string"))` out of an abundance of caution so that nobody comes back and tries to edit the string without realizing that it's a regex.

mattnworb: Gotcha. thanks

helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/CuratorClientFactory.java
mattnworb: I just meant since Optional is used a lot elsewhere in the codebase, but since this is totally an internal interface/class then sounds fine

helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/DefaultZooKeeperClient.java
```diff
@@ -389,6 +397,51 @@ public Stat stat(final String path) throws KeeperException {
...
+    try {
+      return client.getACL().forPath(path);
+    } catch (Exception e) {
+      propagateIfInstanceOf(e, KeeperException.class);
+      throw propagate(e);
+    }
+  }
+
+  @Override
+  public void initializeAclRecursive(final String path, final ACLProvider aclProvider)
```
gimaker: `initializeAclRecursive()` seems rather specific, and only has one caller -- why not keep this code in the `ZooKeeperAclInitializer`?

(Thought I already added a comment about this yesterday but can't find it anywhere today...)

helios-services/src/test/java/com/spotify/helios/servicescommon/RuleBasedZooKeeperAclProviderTest.java
gimaker: `contains()` != equals. It uses equals for a reason. If the expected value is `[acl1]` we don't want the test to pass if the actual value is `[acl1, acl2]`.

gimaker:
:+1:

rohansingh:
@gimaker I think for now let's stick with the permissions we have. If giving agents more permissions becomes necessary in the future, we can do it then. Folks will have to rerun `helios-initialize-acl`, but that shouldn't be too big of a deal.

gimaker:
@rohansingh WFM. If we want to change permission in the future, `helios-initialize-acl` will need to be re-run with "super" authentication, as neither the client nor the master user has `ADMIN` permissions (and thus cannot change ACLs on existing nodes).

PR #805, opened 2016-01-08
Remove old and ignored integration tests

gimaker:
This lets us remove some arcane conditional logic in SystemTestBase
which results in different behaviour depending on the name of the
class.


gimaker:
@rculbertson @rohansingh @mattnworb @davidxia

helios-system-tests/src/main/java/com/spotify/helios/system/SystemTestBase.java
```diff
@@ -211,18 +207,8 @@ public void baseSetup() throws Exception {
     masterAdminPort = temporaryPorts.localPort("helios master admin");
 
     String className = getClass().getName();
```
davidxia: `className` is now not needed

davidxia:
We can also remove from SystemTestBase `masterAdminEndpoint()`. `startJob()`, `deregisterHost()`, and `readLogFully()` are also unused.

rculbertson:
+1 to @davidxia's comments, otherwise :+1:

helios-system-tests/src/main/java/com/spotify/helios/system/SystemTestBase.java
gimaker: Done.

gimaker:
I'll leave the removal of the unrelated but unused `masterAdminEndpoint()`. `startJob()`, `deregisterHost()`, and `readLogFully()` to another PR.

davidxia:
:+1:

mattnworb:
Is there a short answer as to why all these tests were disabled since 084adaaa8de1c428f312347ba81335b0784a1b0e ?

mattnworb:
:+1: to cleanup though - it's too bad codecov doesn't count reduced complexity in the tests themselves

PR #806, opened 2016-01-08
Reduce some code in HeliosSoloDeployment

davidxia:
containerGateway() and assertDockerReachableFromContainer() were
doing similar things. Merge them.

### DIFF
--- a/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
@@ -53,7 +53,7 @@ import static java.util.Collections.singletonList;
 /**
  * A HeliosSoloDeployment represents a deployment of Helios Solo, which is to say one Helios
  * master and one Helios agent deployed in Docker. Helios Solo uses the Docker instance it is
- * deployed to to run its jobs.
+ * deployed on to run its jobs.
  */
 public class HeliosSoloDeployment implements HeliosDeployment {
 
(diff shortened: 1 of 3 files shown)


davidxia:
I'm just getting familiar with @negz' code for helios solo. I did one of his TODOs. I don't feel strongly about this PR and would feel fine leaving this alone.

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -349,9 +319,7 @@ private String getHostPort(final String containerId, final int containerPort)
       final NetworkSettings settings = dockerClient.inspectContainer(containerId).networkSettings();
       for (Map.Entry<String, List<PortBinding>> entry : settings.ports().entrySet()) {
         if (entry.getKey().equals(heliosPort)) {
-          for (PortBinding portBinding : entry.getValue()) {
-            return portBinding.hostPort();
-          }
+          return entry.getValue().get(0).hostPort();
```
negz: Out of curiosity, is this purely stylistic or are there other advantages in Java? For whatever reason I generally use the loop, but I don't feel strongly about it.

negz:
LGTM

In retrospect this was way more straightforward than I had expected it to be when I wrote the TODO.

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
mattnworb: Just style. I've seen `for (..) { return ..}` far less than the other. It can be a little confusing to see a for loop that iterates exactly once.

You can also do `return entry.getValue().iterator().next() ...` which is what the `for (..) { return }` variant basically does anyway.

davidxia:
Cool, if there are no objections, I'll merge in a bit.

gimaker:
:+1:

mattnworb:
:+1:

PR #809, opened 2016-01-11
less-scary error message when SSH_AUTH_SOCK is not set

mattnworb:
Although we only output the previous message at DEBUG, this log message
can be confusing for users if they have failed tests or another problem
with their client talking to Helios that leads them to think that the
SSH_AUTH_SOCK is the real issue.

So try to wordsmith the message a little bit to indicate it probably
isn't a real issue in all cases.


mattnworb:
@davidxia

gimaker:
:+1:

PR #811, opened 2016-01-12
HeliosDeploymentResource ctor needs to be public

mattnworb:
Since this class (from the testing library) is meant to be used by other
projects in their tests, the constructor really needs to not be
package-private.


mattnworb:
@rohansingh @davidxia

mattnworb:
@negz

davidxia:
:+1:

PR #812, opened 2016-01-12
Fix noop assignment in HeliosSoloDeployment

davidxia:


davidxia:
@rohansingh @mattnworb

mattnworb:
:+1:

PR #813, opened 2016-01-12
override DOCKER_HOST of localhost/127.0.0.1 with unix sockets

mattnworb:
...when using HeliosSoloDeployment.

If DOCKER_HOST is 127.0.0.1 or localhost, then the test in
`checkDockerAndGateway()` to make sure that the docker process is reachable
from a container will never work (nor will the actual helios-solo process),
so try to at least replace the endpoint used to talk to docker with the
default unix socket endpoint.

### DIFF
--- a/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
@@ -81,8 +81,7 @@ public class HeliosSoloDeployment implements HeliosDeployment {
 
     this.dockerClient = checkNotNull(builder.dockerClient, "dockerClient");
     this.dockerHost = Optional.fromNullable(builder.dockerHost).or(DockerHost.fromEnv());
-    this.containerDockerHost = Optional.fromNullable(builder.containerDockerHost)
-            .or(containerDockerHostFromEnv());
+    this.containerDockerHost = containerDockerHost(builder);
     this.namespace = Optional.fromNullable(builder.namespace).or(randomString());
     this.env = containerEnv();
     this.binds = containerBinds();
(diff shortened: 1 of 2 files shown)


mattnworb:
@rohansingh @davidxia @negz

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -112,20 +111,36 @@
...
-    } else {
-      return DockerHost.fromEnv();
     }
+
+    // otherwise construct a DockerHost from environment variables, *unless* DOCKER_HOST is set to
+    // localhost or 127.0.0.1 - which will never work inside a container. For those cases, we
+    // override the settings and use the unix socket instead.
+    final DockerHost dockerHost = DockerHost.fromEnv();
+    if (dockerHost.address().equals("localhost") || dockerHost.address().equals("127.0.0.1")) {
+      final String endpoint = DockerHost.DEFAULT_UNIX_ENDPOINT;
```
rohansingh: I would only do this if the current OS is Linux, since otherwise I think there will be an exception down the road when trying to use the Unix sockets wrapper. If it's not Linux, we should just fail early.

rohansingh:
:+1:

mattnworb:
Added a test of this, not sure why I forgot that before.

I think I caught a bug in doing this - the bind argument used when creating the helios-solo container had a value of

```
///var/run/docker.sock:///var/run/docker.sock
```

The extra `//` in front of the path seems wrong.

igorbernstein:
Sweet! thanks for fixing this...it prevented me from using helios testing last time around. I'll try to reintegrate helios-testing next time I have a free moment

PR #814, opened 2016-01-12
document how to use HeliosSoloDeployment in tests

mattnworb:
This is just a start.

### DIFF
--- a/docs/integration_tests_with_helios_solo.md
+++ b/docs/integration_tests_with_helios_solo.md
@@ -1,0 +1,53 @@
+Since Helios 0.8.662, [helios-solo][] can also be run directly from within an
+integration test using a JUnit `@ClassRule`/`@Rule`.
+
+This allows integration tests to run against Helios on the local machine
+executing the tests without having to have `helios-solo` already installed
+and/or running.
+
+In other words, integration tests can be written that will automatically
+install (and clean up) helios-solo for you without having to install any extra
+packages beforehand on your local machine / CI server / etc. The only
+requirement is that the machine running the integration test have Docker
+installed.
+
+# How to use it
+
+If you have a test that is already using [TemporaryJobs][], the test can run
+within a self-contained helios-solo instance with a few small modifications.
+
+Add a `@ClassRule` that for helios-solo to install itself on your docker instance:
+
+```java
+@ClassRule
+public static HeliosDeploymentResource soloResource =
+    new HeliosDeploymentResource(HeliosSoloDeployment.fromEnv().build());
+```
+
+`HeliosDeploymentResource` wraps a `HeliosDeployment` instance which describes
+where the Helios instance is installed. `HeliosDeploymentResource` provides a
+client that talks to the `HeliosDeployment` and cleans it up after the test is
+finished.
+
+Then modify the `TemporaryJobs` declaration to use the client that points to
+your `HeliosSoloDeployment`:
+
+```java
+@Rule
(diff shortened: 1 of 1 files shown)


mattnworb:
@negz @davidxia @gimaker @rohansingh

docs/integration_tests_with_helios_solo.md
```diff
@@ -0,0 +1,53 @@
+Since Helios 0.8.662, [helios-solo][] can also be run directly from within an
```
mattnworb: should bump this to whatever the next release is after the few in-progress improvements/PRs are merged

docs/integration_tests_with_helios_solo.md
```diff
@@ -0,0 +1,53 @@
...
+client that talks to the `HeliosDeployment` and cleans it up after the test is
+finished.
+
+Then modify the `TemporaryJobs` declaration to use the client that points to
+your `HeliosSoloDeployment`:
+
+```java
+@Rule
+public TemporaryJobs temporaryJobs = TemporaryJobs.builder()
+  .client(soloResource.client())
```
davidxia: Either change `soloResource` to `solo` or the `solo` on line 23 to `soloResource` to be consistent.

docs/integration_tests_with_helios_solo.md
```diff
@@ -0,0 +1,53 @@
...
+
+#### ClassRule vs Rule
+The example above will set up `HeliosSoloDeployment` once for all of the tests
+in your `Test` class. To instead set up (and tear down) a
+`HeliosSoloDeployment` for *each individual* test in the class, change the
+`@ClassRule` to a `@Rule` and remove the `static` modifier. 
+
+Since there is a bit of set up work in creating a `HeliosSoloDeployment`, this
+will add a lot of overhead to each test. We recommend using
+`HeliosSoloDeployment` as a `@ClassRule`.
```
davidxia: I'm trying to figure out how to create a test suite rule.

docs/integration_tests_with_helios_solo.md
mattnworb: good catch, thanks

mattnworb:
Fixed some awkwardness in the wording and added a note about the `localhost`/`127.0.0.1` problem. Please review!

mattnworb:
@davidxia I have a feeling that won't be possible without a lot of junit hacking.

Assuming you are looking to avoid the time-wait of starting up helios-solo for each test class, I wonder if revamping how HeliosSoloDeployment works could also work:
- HeliosSoloDeployment uses a fixed name for the container it starts (e.g. `helios-solo-auto`)
- at startup, if the `helios-solo-auto` container is running, just use it
- otherwise create and start that container
- after 10 minutes (or some other value) of non-use, the container kills itself

davidxia:
I think the solo container should be a unique name per test run, otherwise other projects' ITs running on the same host will end up using the same solo container.

davidxia:
:+1:

davidxia:
@mattnworb Should the docs say to be aware of `HELIOS_HOST_FILTER`? https://github.com/spotify/helios/blob/master/helios-testing/src/test/java/com/spotify/helios/testing/HeliosSoloDeploymentTest.java#L66

mattnworb:
@davidxia good point, they should. And/or we could hack TemporaryJobs to ignore filters when using HeliosSoloDeployment as the two concepts don't make sense together.

davidxia:
The latter makes more sense.

PR #815, opened 2016-01-13
Ensure cleanup of helios-solo jobs with a watchdog process

davidxia:


helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -41,21 +40,25 @@
...
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Strings.isNullOrEmpty;
 import static java.util.Collections.singletonList;
 
-/**
- * A HeliosSoloDeployment represents a deployment of Helios Solo, which is to say one Helios
- * master and one Helios agent deployed in Docker. Helios Solo uses the Docker instance it is
- * deployed on to run its jobs.
- */
```
mattnworb: why remove this?

mattnworb:
did you base this off of an old branch?

davidxia:
yea I'll rebase

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -249,28 +336,30 @@ private String deploySolo(final String heliosHost) throws HeliosDeploymentExcept
...
+        .image(HELIOS_IMAGE)
+        .build();
 
     log.info("starting container for helios-solo with image={}", HELIOS_IMAGE);
 
     final ContainerCreation creation;
     try {
-      dockerClient.pull(HELIOS_IMAGE);
-      final String containerName = HELIOS_CONTAINER_PREFIX + this.namespace;
+      //dockerClient.pull(HELIOS_IMAGE);
```
gimaker: I commented this line out in my branch for some reason that I forgot, but I doubt it was meant to stay commented out...

gimaker: IIRC, I had to comment it out because doing a `pull()` would override the `latest` tag but I could be remembering that incorrectly.

davidxia: It needs to be commented out for now because it'll pull down the spotify/helios-solo:latest image from docker hub which doesn't have the watchdog. That's also the reason I had to modify circle.sh to build and tag a spotify/helios-solo:latest image. We can put these back later I think.

davidxia:
@gimaker @mattnworb @rohansingh

gimaker:
Seems like this, if used as it is, would still create one helios-solo deployment per test-class?

gimaker:
Oops

davidxia:
Ah yea I haven't gotten to that part yet.

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -344,20 +435,24 @@ private String randomString() {
...
-    return this.heliosClient;
+    return heliosClient;
   }
 
   /**
    * Undeploy (shut down) this HeliosSoloDeployment.
    */
-  public void close() {
-    log.info("shutting ourselves down");
+  @Override
```
gimaker: The comment above is now on the wrong method

gimaker:
All of this would make a bit more sense if we only run one helios-solo deployment per test run. Now there's two ways of cleaning up the helios-solo deployment (either via the watchdog, or being explicitly shutdown via `HeliosSoloDeployment.close()` -- and they work differently)

solo/docker/start.sh
```diff
@@ -78,6 +78,34 @@ com.spotify.helios.master.MasterMain \
...
 
-set +x
-# Sleep or execute command line
-while :; do sleep 1; done
+if [ "$HELIOS_SOLO_SUICIDE" -eq 1 ]; then
+    HELIOS_CLI_CMD="java -cp /*.jar -Xmx128m -Djava.net.preferIPv4Stack=true com.spotify.helios.cli.CliMain -z http://localhost:5801"
+
+    python "/watchdog.py"
+
+    echo "Cleaning up jobs..."
```
rohansingh: Very clever.

davidxia:
Ideally, the watchdog would start 1 helios-solo at the beginning, we'd tell all the test classes which helios-solo container name to use, and then the watchdog would tear down helios-solo at the end. But I can't figure out have all the test classes share 1 helios-solo container namespace.

We can hold off on this and pick it up if necessary after using helios-solo in the wild a bit.

gimaker:
@davidxia There was code doing what you describe (running one shared helios-solo) in my PoC branch - why doesn't that approach work?

davidxia:
closing for now. We can reopen later if we need it

PR #818, opened 2016-01-14
import new DockerHost

mattnworb:
in between #813 being created and getting merged, the DockerHost class
was removed from the helios source tree and moved to docker-client.
Since the branch for #813 was not rebased this wasn't picked up earlier.


mattnworb:
@rohansingh @davidxia

PR #819, opened 2016-01-14
Fix test failure when port 20000 is occupied

rohansingh:
1. Check that the allocated port is in the correct range, not that it is
   specifically port 20000.

2. Remove one test that is just not possible to do safely.


rohansingh:
@mattnworb @davidxia

davidxia:
:+1:

mattnworb:
lgtm

PR #822, opened 2016-01-15
HttpsHandlers: Fix file descriptor leak

rohansingh:
We weren't closing the certificate or key files, leading to the number of
open file descriptors building up over time.


rohansingh:
@gimaker @davidxia @mattnworb

mattnworb:
:+1: 

wow I missed before that `PEMParser.parseObject` returns you a plain old Object

davidxia:
:+1:

PR #825, opened 2016-01-20
Decrease Kafka metadata fetch timeout

davidxia:
Decrease from 5 to 1 second. Even though we don't block,
the Helios client will timeout in less than 5 seconds.


davidxia:
@mattnworb @rohansingh @gimaker @alan

davidxia:
alanw

davidxia:
@alanw

mattnworb:
change makes sense to me but it would be nicer if these values weren't hardcoded to begin with IMHO

davidxia:
Would a command line arg be better?

gimaker:
:+1:

PR #826, opened 2016-01-20
Fix job registration domain during job creation

davidxia:
Do not overwrite config value in file with default value
of empty string from CLI args.


davidxia:
@rohansingh @gimaker @mattnworb

davidxia:
@spdage

mattnworb:
:+1:

spdage:
+1

PR #827, opened 2016-01-20
Agent re-registration now preserves jobs deployed on the replaced host

gimaker:
I.e. if agent A has job J deployed to it, and is then replaced by agent B
under the same name the B will also automatically get J deployed to it.

Previously, the automatic re-registration would fail if there were any
deployed jobs on the host being replaced -- which was the case almost all
of the time. With this in place, the automatic re-registration should
actually work. (History of any existing jobs will also be preserved.)

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/master/DeadAgentReaper.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/DeadAgentReaper.java
@@ -17,15 +17,13 @@
 
 package com.spotify.helios.master;
 
-import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.Lists;
-
 import com.spotify.helios.agent.InterruptingScheduledService;
 import com.spotify.helios.common.Clock;
 import com.spotify.helios.common.SystemClock;
 import com.spotify.helios.common.descriptors.AgentInfo;
 import com.spotify.helios.common.descriptors.HostStatus;
 
+import org.apache.commons.lang.time.DurationFormatUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -33,6 +31,8 @@ import java.util.List;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+
+import static com.google.common.base.Preconditions.checkArgument;
 
 /**
  * De-registers dead agents, where an agent that has been DOWN for more than X time is considered
@@ -48,6 +48,13 @@ public class DeadAgentReaper extends InterruptingScheduledService {
 
   private final MasterModel masterModel;
   private final long timeoutMillis;
+  private final Clock clock;
+
+  public DeadAgentReaper(final MasterModel masterModel,
+                         final long timeout,
+                         final TimeUnit timeUnit) {
+    this(masterModel, timeout, timeUnit, SYSTEM_CLOCK);
+  }
 
   /**
    * We check for dead agents every 30 minutes, which means the actual timeout the specified timeout
@@ -55,38 +62,23 @@ public class DeadAgentReaper extends InterruptingScheduledService {
    */
   public DeadAgentReaper(final MasterModel masterModel,
                          final long timeout,
-                         final TimeUnit timeUnit) {
+                         final TimeUnit timeUnit,
+                         final Clock clock) {
     this.masterModel = masterModel;
+    checkArgument(timeout > 0);
     this.timeoutMillis = timeUnit.toMillis(timeout);
+    this.clock = clock;
   }
 
   @Override
   protected void runOneIteration() {
     log.debug("Reaping agents");
-    final List<String> deadAgents = getDeadAgents(masterModel, timeoutMillis, SYSTEM_CLOCK);
-    log.debug("Reaping {} dead agents: {}", deadAgents.size(), deadAgents);
-
-    for (final String agent : deadAgents) {
-      log.info("Reaping dead agent {}", agent);
-      try {
-        masterModel.deregisterHost(agent);
-      } catch (Exception e) {
-        log.error("Failed to reap dead agent {}", agent, e);
-      }
-    }
-  }
-
-  @VisibleForTesting
-  static List<String> getDeadAgents(final MasterModel masterModel, final long timeoutMillis,
-                                    final Clock clock) {
-    final List<String> deadAgents = Lists.newArrayList();
     final List<String> agents = masterModel.listHosts();
-
     for (final String agent : agents) {
       try {
         final HostStatus hostStatus = masterModel.getHostStatus(agent);
         if (hostStatus == null || hostStatus.getStatus() != HostStatus.Status.DOWN) {
-          // Hot not found or host not DOWN -- nothing to do, move on to the next host
+          // Host not found or host not DOWN -- nothing to do, move on to the next host
           continue;
         }
 
@@ -99,14 +91,18 @@ public class DeadAgentReaper extends InterruptingScheduledService {
         final long downDurationMillis = clock.now().getMillis() - downSince;
 
         if (downDurationMillis >= timeoutMillis) {
-          deadAgents.add(agent);
+          try {
+            log.info("Reaping dead agent '{}' (DOWN for {} hours)",
+                     DurationFormatUtils.formatDurationHMS(downDurationMillis));
+            masterModel.deregisterHost(agent);
+          } catch (Exception e) {
+            log.warn("Failed to reap agent '{}'", agent, e);
+          }
         }
       } catch (Exception e) {
         log.warn("Failed to determine if agent '{}' should be reaped", agent, e);
       }
     }
-
-    return deadAgents;
   }
 
   @Override
--- a/helios-services/src/main/java/com/spotify/helios/master/DeadAgentReaper.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/DeadAgentReaper.java
@@ -35,7 +35,7 @@ import java.util.concurrent.TimeUnit;
 import static com.google.common.base.Preconditions.checkArgument;
 
 /**
- * De-registers dead agents, where an agent that has been DOWN for more than X time is considered
+ * De-registers dead agents, where an agent that has been DOWN for more than X hours is considered
  * dead.
  */
 public class DeadAgentReaper extends InterruptingScheduledService {
@@ -51,22 +51,16 @@ public class DeadAgentReaper extends InterruptingScheduledService {
   private final Clock clock;
 
   public DeadAgentReaper(final MasterModel masterModel,
-                         final long timeout,
-                         final TimeUnit timeUnit) {
-    this(masterModel, timeout, timeUnit, SYSTEM_CLOCK);
+                         final long timeoutHours) {
+    this(masterModel, timeoutHours, SYSTEM_CLOCK);
   }
 
-  /**
-   * We check for dead agents every 30 minutes, which means the actual timeout the specified timeout
-   * plus up to 30 minutes. I.e. very low timeouts will not work in practice.
-   */
   public DeadAgentReaper(final MasterModel masterModel,
-                         final long timeout,
-                         final TimeUnit timeUnit,
+                         final long timeoutHours,
                          final Clock clock) {
     this.masterModel = masterModel;
-    checkArgument(timeout > 0);
-    this.timeoutMillis = timeUnit.toMillis(timeout);
+    checkArgument(timeoutHours > 0);
+    this.timeoutMillis = TimeUnit.HOURS.toMillis(timeoutHours);
     this.clock = clock;
   }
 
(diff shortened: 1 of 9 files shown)


gimaker:
@davidxia @mattnworb @rohansingh

mattnworb:
I have a few questions.

1. This works ok with the agent's restricted ACLs?

2. What happens if agent A goes offline for a long period of time (for instance one month) and is replaced by B with the same name? Is it safe that a potentially "old" version of a job would be deployed?

3. What if A is in a deployment group - when A is offline, it's job isn't updated any longer, is it? Would job:oldversion or job:current-dg-version be deployed to B?

gimaker:
1. Yes.
2. ¯\_(ツ)_/¯
3. `job:oldversion`, then when the master detects that the hosts has changed it will in most cases initiate a rolling-update and deployed `job:current-dg-version` shortly after.

gimaker:
Alternatively, we could just delete everything under `/config/hosts/<host>`, including jobs.

gimaker:
Actually, the latter (deleting existing jobs when re-registering) will not work with the current ACLs, as agents do not have any permissions except READ on the `/config/hosts/<host>/jobs` tree.

mattnworb:
Ignoring the ACL or implementation I think that having the "new" agent come online with no jobs deployed to it is less of a surprise, especially if it's deployment group (if the agent is in one) would roll out a possibly-newer job anyway.

gimaker:
(Technically, agents are never in deployment groups - but jobs deployed to them might be.)

I'm inclined to agree, but I'm not sure how to best overcome the hurdles posed by ACLs... Giving agents more permissions on `/config/hosts/<host>/jobs` allows any compromised agent to have free reign on other agents (e.g. deploy rouge jobs, undeploy jobs, etc.)

gimaker:
What other options do we have?

 * Periodically garbage collected DOWN'ed hosts from the masters? This ought to be work, but with a delay. On the plus side, it'd also get rid of those "DOWN since 96 days" hosts :)

mattnworb:
Another option would be to have an API endpoint on the master that the agent could call to handle the deregistration server-side.  

It's too bad we didn't add this PR _before_ ACLs

gimaker:
It's not too late to pull the breaks on rolling out the ACLs. Soon it is, but not yet.

gimaker:
@mattnworb Any thoughts on how that would work in practice? I assume the agent would need to provide some kind of authentication to the master, to stop unauthenticated users from exploiting that endpoint.

gimaker:
Okay, here's an idea: do this plus garbage collect DOWN hosts (when they're down for, say, more than 3 days or so). That way, a stale job might be deployed but it will *at most* be 3 days or so stale.

mattnworb:
I like that idea 👍🏻

gimaker:
Tests for the reaping of dead agents are pretty shoddy -- not completely sure how to test that functionality properly, ideas welcome!

mattnworb:
the reRegister test seems like it would be better handled in a unit test where you could start up the ZK testing server, call the reregister function, and assert that certain nodes have had their values/ctime/mtime changed (from `client.stat(path)`) and that other nodes have *not* been changed.

helios-services/src/main/java/com/spotify/helios/master/DeadAgentReaper.java
```diff
@@ -0,0 +1,117 @@
...
+        masterModel.deregisterHost(agent);
+      } catch (Exception e) {
+        log.error("Failed to reap dead agent {}", agent, e);
+      }
+    }
+  }
+
+  @VisibleForTesting
+  static List<String> getDeadAgents(final MasterModel masterModel, final long timeoutMillis,
+                                    final Clock clock) {
```
mattnworb: I don't think this would need to be `static` if you just made the `Clock` a constructor parameter, which makes a lot more sense to me personally

mattnworb: or `VisibleForTesting` either

helios-services/src/main/java/com/spotify/helios/master/DeadAgentReaper.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  @Override
+  protected void runOneIteration() {
+    log.debug("Reaping agents");
+    final List<String> deadAgents = getDeadAgents(masterModel, timeoutMillis, SYSTEM_CLOCK);
+    log.debug("Reaping {} dead agents: {}", deadAgents.size(), deadAgents);
+
+    for (final String agent : deadAgents) {
+      log.info("Reaping dead agent {}", agent);
```
mattnworb: I think it would be helpful to log the amount of downtime here too

helios-services/src/main/java/com/spotify/helios/master/DeadAgentReaper.java
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * De-registers dead agents, where an agent that has been DOWN for more than X time is considered
+ * dead.
+ */
+public class DeadAgentReaper extends InterruptingScheduledService {
+
+  private static final Clock SYSTEM_CLOCK = new SystemClock();
+  private static final long INTERVAL = 30;
+  private static final TimeUnit INTERVAL_TIME_UNIT = TimeUnit.MINUTES;
```
mattnworb: I think these should all be constructor parameters too (and maybe CLI args)

gimaker: Why?

helios-services/src/main/java/com/spotify/helios/master/MasterService.java
```diff
@@ -201,6 +202,9 @@ public MasterService(final MasterConfig config,
     final ReactorFactory reactorFactory = new ReactorFactory();
     this.rollingUpdateService = new RollingUpdateService(model, reactorFactory);
 
+    // Set up agent reaper (de-registering hosts that have been DOWN for more than X time)
+    this.agentReaper = new DeadAgentReaper(model, config.getAgentReapingTimeout(), TimeUnit.HOURS);
```
mattnworb: do you need to start this service somewhere?

helios-services/src/test/java/com/spotify/helios/master/DeadAgentReaperTest.java
```diff
@@ -0,0 +1,122 @@
...
+    when(clock.now()).thenReturn(new Instant(2000));
+
+    final List<Datapoint> datapoints = Lists.newArrayList(
+        new Datapoint("host1", 0, TIMEOUT_MILLIS - 1, HostStatus.Status.DOWN, true),
+        new Datapoint("host2", 0, TIMEOUT_MILLIS + 1, HostStatus.Status.DOWN, false),
+        new Datapoint("host3", 1000, 1000, HostStatus.Status.UP, false),
+        new Datapoint("host4", 500, 300, HostStatus.Status.DOWN, true),
+        new Datapoint("host5", 5000, 0, HostStatus.Status.DOWN, false),
+        new Datapoint("host6", 0, 0, HostStatus.Status.UP, false)
+    );
```
mattnworb: looks like a good fit for a parameterized test :smile:

mattnworb:
By "shoddy" do you mean DeadAgentReaperTest is flaky or not as clean as you would like? I think it would read a lot easier if it was a ParameterizedTest (you could do `if (expectReaped)` instead of the filter stuff), and if the test invoked `runOneIteration` and did `verify(mastermodel).deregister(the-host, ..)` instead of invoking the `getDeadAgents` method.

helios-services/src/test/java/com/spotify/helios/master/DeadAgentReaperTest.java
```diff
@@ -0,0 +1,122 @@
...
+    final Clock clock = mock(Clock.class);
+    // Current time is 2000 (ms)
+    when(clock.now()).thenReturn(new Instant(2000));
+
+    final List<Datapoint> datapoints = Lists.newArrayList(
+        new Datapoint("host1", 0, TIMEOUT_MILLIS - 1, HostStatus.Status.DOWN, true),
+        new Datapoint("host2", 0, TIMEOUT_MILLIS + 1, HostStatus.Status.DOWN, false),
+        new Datapoint("host3", 1000, 1000, HostStatus.Status.UP, false),
+        new Datapoint("host4", 500, 300, HostStatus.Status.DOWN, true),
+        new Datapoint("host5", 5000, 0, HostStatus.Status.DOWN, false),
```
mattnworb: the start time (5000) is after the current time (2000)?

helios-services/src/main/java/com/spotify/helios/master/DeadAgentReaper.java
mattnworb: so that you can change the schedule without rebuilding the code

mattnworb: and/or disable the "reaping" altogether if the interval is `0` or some value

gimaker:
"Shoddy" as in it doesn't test the actual functionality, but only a small part of the implementation. There's currently no test that verifies that agents are actually reaped by the service.

Since the reaping logic operates on a list of hosts (with various `continue` statements in there) I wanted to do the tests on an actual list to test that logic, as opposed to running a parameterized test multiple times on single-item lists.

mattnworb:
Oh right good point. Disregard my comments about ParameterizedTest then. But I still think you could get mileage out of calling runOneIteration (or whatever does the actual `get list` and `deregister`) and calling `verify` on the mock

helios-services/src/main/java/com/spotify/helios/master/MasterService.java
gimaker: Nope -- this, that the service doesn't need to be started, is now also covered by the test.

helios-services/src/main/java/com/spotify/helios/master/DeadAgentReaper.java
gimaker: My idea for "disabling" reaping was to set the reaping timeout to something like 999 days. That ought to be good enough?

gimaker: Do you really think we need to be able to change how often that code is run? Adding CLI args for it seems like it would just add clutter.

helios-services/src/main/java/com/spotify/helios/master/MasterService.java
mattnworb: I see. Is that normal for `InterruptingScheduledService` and friends?

helios-services/src/main/java/com/spotify/helios/master/DeadAgentReaper.java
gimaker: The only good reason I can think of for adding those CLI args is to set the interval low for running integration tests.

helios-services/src/main/java/com/spotify/helios/master/DeadAgentReaper.java
gimaker: Done.

helios-services/src/main/java/com/spotify/helios/master/DeadAgentReaper.java
gimaker: Done.

helios-services/src/main/java/com/spotify/helios/master/DeadAgentReaper.java
mattnworb: I think that might really surprise someone if we were to ever actually reach uptime of 3 years :smirk: 

I don't expect to have to change it often but if there ever is it is nicer IMHO to have the option to change configuration rather than re-releasing Helios (and have the default value be the 30 min one here). 

I think the clutter of adding new CLI args has more to do with how we've implemented it with the Config, Parser, Server etc. I am curious if we would get more mileage out of just reading some sort of yaml/json file at startup and not having to have any new switch/flag be added in 3-4 places.

helios-services/src/main/java/com/spotify/helios/master/MasterService.java
gimaker: Well, hmm, maybe you do need to start it actually. Let me investigate...

gimaker: Okay, yeah. It needs to be started :)

mattnworb:
Btw one other thought, and this is probably wider than just this one change, but for these periodic tasks like "reaping", does it really make sense for each master in a cluster to be running these same tasks? I assume using transactions with the ZK operations prevents data inconsistencies but it seems unnecessary for each node to be doing the same work. We could use Curator's [`LeaderLatch`](http://curator.apache.org/curator-recipes/leader-latch.html) to designate one node as the batch job runner.

gimaker:
`HeliosSoloDeploymentTest`... *grumble grumble*

gimaker:
I feel pretty much down with this unless people have more comments/concerns.

mattnworb:
I am down with this too :+1:

PR #828, opened 2016-01-20
use bridge interface for http healthchecks

mattnworb:
If the agent is using `unix://` to communicate with Docker, then the
`HttpHealthChecker` ends up using `localhost` as the host in the URL it
constructs to healthcheck the job.

When the agent is running in helios-solo (i.e. in a Docker container
itself), the agent uses the default `unix:///var/run/docker.sock` endpoint
to talk to Docker but a HTTP request to
`localhost:external-port` will not work as `localhost` refers to the
container itself.

In the scenario where docker.host is using the unix protocol, this change
will have the `HttpHealthChecker` instead use the bridge address of the
container to send it's healthcheck to.

### DIFF
--- a/helios-integration-tests/src/test/java/com/spotify/helios/HeliosSoloHealthchecksIT.java
+++ b/helios-integration-tests/src/test/java/com/spotify/helios/HeliosSoloHealthchecksIT.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) 2014 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


mattnworb:

This change works in helios-solo but to be honest I am not sure if it makes any sense for the more-usual case of a non-container helios-agent that communicates with docker using unix sockets. In that case, connecting to the bridge interface and the _external_ port wouldn't work, would it? This logic is probably too specialized for helios-solo.

I also considered having helios-solo add a JVM system property at startup when launching the agent that could be checked here in the Healthchecker to tweak the logic, but that feels too hacky - this issue seems like it should be easily generalized.

PR #829, opened 2016-01-21
upstart: Export the ZK password env vars

rohansingh:
If the `HELIOS_ZK_MASTER_PASSWORD` or `HELIOS_ZK_AGENT_PASSWORD` are set in
the defaults file, we want to export them to the Helios process.


rohansingh:
@gimaker

davidxia:
:+1:

PR #831, opened 2016-01-21
Allow users to set executor service for HeliosClient

davidxia:


davidxia:
Is this the right way to this use-case? https://ghe.spotify.net/dflemstr/barge/pull/6

davidxia:
@mattnworb @gimaker

helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
```diff
@@ -511,6 +511,7 @@ public static Builder newBuilder() {
     private ClientCertificatePath clientCertificatePath;
     private Supplier<List<Endpoint>> endpointSupplier;
     private boolean sslHostnameVerification = true;
+    private ListeningScheduledExecutorService executorService = defaultExecutorService();
```
mattnworb: I think this should be set in the `build()` method instead so that an executor is not created/started here and then replaced for people calling `setExecutorService`

helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
```diff
@@ -563,22 +564,28 @@ public Builder setClientCertificatePath(final ClientCertificatePath clientCertif
       return this;
     }
 
+    public Builder setExecutorService(
+        final ListeningScheduledExecutorService executorService) {
```
mattnworb: this signature could probably just be `ScheduledExecutorService` and we wrap it in `listeningDecorator()`, I'm not sure that callers would care or have a LSES beforehand

helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
```diff
@@ -563,22 +568,31 @@ public Builder setClientCertificatePath(final ClientCertificatePath clientCertif
...
       final ScheduledExecutorService executor = MoreExecutors.getExitingScheduledExecutorService(
           (ScheduledThreadPoolExecutor) newScheduledThreadPool(4), 0, SECONDS);
+      return MoreExecutors.listeningDecorator(executor);
+    }
 
-      final ListeningScheduledExecutorService listeningExecutor =
-          MoreExecutors.listeningDecorator(executor);
+    private RequestDispatcher createDispatcher() {
+      if (executorService == null) {
+        executorService =  defaultExecutorService();
```
gimaker: nitpick: There's an extra space after the `=`

mattnworb:
lgtm

PR #832, opened 2016-01-21
Fix ZooKeeperRegistrarService backoff

davidxia:
We were not actually backing off before for non KeeperExceptions.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/agent/AgentService.java
+++ b/helios-services/src/main/java/com/spotify/helios/agent/AgentService.java
@@ -373,11 +373,13 @@ public class AgentService extends AbstractIdleService implements Managed {
     client.start();
 
     // Register the agent
-    zkRegistrar = new ZooKeeperRegistrarService(
-        client,
-        new AgentZooKeeperRegistrar(this, config.getName(),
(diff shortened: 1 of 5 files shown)


davidxia:
@gimaker @mattnworb @rohansingh

mattnworb:
so it would retry immediately (and infinitely)?

davidxia:
yup :(

mattnworb:
I was going to suggest adding a test to ensure the desired behavior, by mocking out the `zooKeeperRegistrar` and having the first call return an exception and the second one succeed, but I am not really sure how you would test the `Thread.sleep` part.

:+1:

davidxia:
I can pass in an implementation of the `Clock` interface like we do elsewhere.

mattnworb:
but that wouldn't help test that the thread actually slept, would it?

gimaker:
:+1:

davidxia:
@mattnworb Sorry, I should've wrote `Sleeper` instead of `Clock`. My second commit will mock the `Sleeper` and test that the `.sleep()` is called with correct args.

mattnworb:
nice, I didn't know about `Sleeper` :+1:

PR #833, opened 2016-01-25
helios-solo: use bridge interface for http healthchecks

mattnworb:
If the agent is using `unix://` to communicate with Docker, then the
`HttpHealthChecker` ends up using `localhost` as the host in the URL it
constructs to healthcheck the job.

When the agent is running in helios-solo (i.e. in a Docker container
itself), the agent uses the default `unix:///var/run/docker.sock` endpoint
to talk to Docker but a HTTP request to
`localhost:external-port` will not work as `localhost` refers to the
container itself.

In the scenario where the agent is running in helios-solo *and* docker.host
is using the unix protocol, this change will have the
`HttpHealthChecker` instead use the bridge address of the container to send
it's healthcheck to.

In order for the agent to know if it is running within helios-solo, added a
system property in the helios-solo startup script. This property is also
enabled for the master as well in case it is needed in the future.

This is a tweak of #828 so that the logic to use the bridge interface is only applied when using helios-solo.

### DIFF
--- a/helios-integration-tests/src/test/java/com/spotify/helios/HeliosSoloHealthchecksIT.java
+++ b/helios-integration-tests/src/test/java/com/spotify/helios/HeliosSoloHealthchecksIT.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) 2014 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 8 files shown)


mattnworb:
:fireworks: :tada: all the tests pass on the first try :tada: :fireworks: 

@negz @gimaker @rohansingh @davidxia

mattnworb:
to be clear having the HttpHealthChecker branch on a system property to check if it is running inside helios-solo or not feels pretty hacky; but I wasn't sure how to programmatically have the agent know if it is running inside a container (which is on the same host as the container for the actual job).

negz:
LGTM

It does feel a little hacky, but I can't think of a better way.

davidxia:
:100:

gimaker:
Would prefer to change the logic so that it uses the bridge address when Helios ís running in a container, regardless if it's helios-solo or not.

mattnworb:
@gimaker fair point. See ce5a65c

gimaker:
:+1:

PR #834, opened 2016-01-26
fix logging of exception when loading AgentProxy

mattnworb:
The current log message ends up logged as:

```
15:45:51.632 [main] DEBUG c.spotify.helios.client.HeliosClient - {}
java.lang.RuntimeException: The environment variable SSH_AUTH_SOCK is not set. Please configure your ssh-agent.
    at com.spotify.sshagentproxy.AgentProxies$DefaultAgentProxy.fromEnvironmentVariable(AgentProxies.java:61) ~[ssh-agent-proxy-0.1.4.jar:na]
```

When the last argument is a Throwable, slf4j doesn't treat it as a
parameter in the message so `{}` is logged by itself.


mattnworb:
@davidxia

gimaker:
:+1:

PR #836, opened 2016-01-26
fix occasional NullPointerException in DefaultDeployer

mattnworb:
For currently unknown reasons, occasionally
`client.hostStatus(candidateHost)` will return a completed
ListenableFuture with a `null` value when called by DefaultDeployer
(which HeliosClient does when the host doesn't exist). It's not clear
why the host would have `null` status when a few lines earlier the host
was retrieved from `client.hosts()`.

But in any case, this fixes the NPE that occurs when this happens.

### DIFF
--- a/helios-testing/src/test/java/com/spotify/helios/testing/DefaultDeployerTest.java
+++ b/helios-testing/src/test/java/com/spotify/helios/testing/DefaultDeployerTest.java
@@ -29,10 +29,9 @@ import com.spotify.helios.common.descriptors.HostStatus.Builder;
 import com.spotify.helios.common.descriptors.JobId;
 import com.spotify.helios.common.descriptors.TaskStatus;
 
+import org.junit.Rule;
 import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.mockito.Mock;
-import org.mockito.runners.MockitoJUnitRunner;
+import org.junit.rules.ExpectedException;
 
 import java.util.List;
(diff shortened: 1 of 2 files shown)


mattnworb:
@spotify/helios-team

negz:
LGTM

helios-testing/src/test/java/com/spotify/helios/testing/DefaultDeployerTest.java
```diff
@@ -42,10 +41,9 @@
 import static java.util.Collections.emptyMap;
 import static java.util.concurrent.TimeUnit.MINUTES;
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.fail;
+import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
-@RunWith(MockitoJUnitRunner.class)
```
davidxia: I've seen this in a few other test classes as well. Just for my own knowledge, What's this supposed to do and why don't we need it?

mattnworb: it takes the fields annotated with `@Mock` and sets their value to `Mockito.mock(TheClass.class)` at setup-time. I replaced it here because it felt like less code to just create the one mock myself in line 70 (and moreso because then I couldn't initialize the DefaultDeployer in a single line and I didn't want to add a `@Before` method :grinning: )

PR #838, opened 2016-01-26
Fix deregistering of hosts where the are jobs without a status entry

gimaker:
Prior to this change, deregistering a host would fail if it had a job
"deployed" to it that lacked a corresponding node in the
/status/hosts/<host>/jobs tree.

One easy way to force this particular state is to:
 * Kill a host (i.e. make it report as DOWN)
 * Deploy a job to the now dead host
 * De-register the host

### DIFF
--- a/helios-system-tests/src/main/java/com/spotify/helios/system/DeregisterTest.java
+++ b/helios-system-tests/src/main/java/com/spotify/helios/system/DeregisterTest.java
@@ -31,11 +31,15 @@ import com.spotify.helios.common.protocol.CreateJobResponse;
 import com.spotify.helios.common.protocol.HostDeregisterResponse;
 import com.spotify.helios.common.protocol.JobDeleteResponse;
 import com.spotify.helios.common.protocol.JobDeployResponse;
+import com.spotify.helios.servicescommon.ZooKeeperRegistrarUtil;
+import com.spotify.helios.servicescommon.coordination.DefaultZooKeeperClient;
+import com.spotify.helios.servicescommon.coordination.Paths;
 
 import org.hamcrest.Matchers;
 import org.junit.Rule;
 import org.junit.Test;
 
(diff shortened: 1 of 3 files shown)


gimaker:
@mattnworb @davidxia @rohansingh @negz

davidxia:
Nice catch. :+1: But we were seeing new agents failing to deregister old agents even when the status node was most likely there right?

gimaker:
@davidxia I haven't seen what you describe myself, but please share the details if you have.

For reference, what it might looks like in a situation where it fails today is this:

```
/config/hosts/host1/jobs/job-1
/config/hosts/host1/jobs/job-2
/status/hosts/host1/jobs/job-1
...
```

Currently when reregistering, it will try to delete all jobs that are present in `/status/hosts/host1/jobs`. In the above example is just `job-1`, and since `/config/hosts/host1/jobs` isn't empty it will fail to deregister the host.

gimaker:
Kudos to @mattnworb for figuring this out, btw.

helios-system-tests/src/main/java/com/spotify/helios/system/DeregisterTest.java
```diff
@@ -105,6 +109,48 @@ public void testDeregister() throws Exception {
     assertEquals(JobDeleteResponse.Status.OK, deleteResponse.getStatus());
   }
 
+  // Verify that we can deregister a host there are jobs deployed to it, for for which there's no
```
mattnworb: nitpick: `for for`

mattnworb:
:+1: nice job figuring out the steps to reproduce

davidxia:
I fixed ZooKeeperRegistrarServiceUtilTest here https://github.com/spotify/helios/pull/839

davidxia:
:+1:

PR #840, opened 2016-02-01
Add health check gauge

davidxia:


mattnworb:
I think it would be nicer to have one gauge per healthcheck, rather than one overall gauge, so the values can be reported separately. 

Also I think the same gauges should be added to AgentService as well.

Perhaps something like this in MasterService/AgentService

```java
    // after all environment.healthchecks().register(..) calls...
    for (final String name : environment.healthChecks().getNames()) {
      environment.metrics().register("helios." + name + ".ok", new Gauge<Boolean>() {
        @Override
        public Boolean getValue() {
          final HealthCheck.Result result = environment.healthChecks().runHealthCheck(name);
          return result.isHealthy();
        }
      });
    }

```

gimaker:
For our particular use case, we need the overall metric as well, right? Since we do alerting on aggregate value.

mattnworb:
We should be able to set up an alert if *any* of the healthchecks go to 0; basically a group by `host, healthcheckname`

davidxia:
@spotify/helios-team Finally got the tests to pass.

helios-services/pom.xml
```diff
@@ -554,6 +554,14 @@
...
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <source>1.8</source>
+          <target>1.8</target>
+        </configuration>
+      </plugin>
```
mattnworb: How come this is needed? The same config is in `<pluginManagement>` of the parent pom

circle.yml
```diff
@@ -26,6 +26,8 @@ machine:
   # tells CircleCI that we want virtualenv so that we can pip install codecov without errors
   python:
     version: 2.7.10
+  java:
+    version: oraclejdk8
```
mattnworb: looks like this duplicates what is in line 15

mattnworb:
:+1: for the actual healthcheck change but I think the two things pointed out above are not necessary

davidxia:
Thanks. Yea I forgot about those.

davidxia:
@negz @gimaker

solo/base/version.txt
```diff
@@ -1 +1 @@
-0.3
+0.4
```
gimaker: why was this needed? otherwise lgtm

gimaker:
Tests would be nice, too.

solo/base/version.txt
davidxia: 0.4 has java8 installed. 0.3 doesn't

mattnworb:
I upgraded java in helios-solo-base back in that original PR, in 0.3

```
$> docker run -it spotify/helios-solo-base:0.3 bin/bash
root@3024bb6592c7:/# java -version
java version "1.8.0_60"
Java(TM) SE Runtime Environment (build 1.8.0_60-b27)
Java HotSpot(TM) 64-Bit Server VM (build 25.60-b23, mixed mode)
root@3024bb6592c7:/#
```

but it doesn't hurt to bump the version again

davidxia:
weird, could've sworn I logged into circleCI's heios-solo container after tests failed and ran `java -version` only to see 1.7. /shrugs

helios-system-tests/src/main/java/com/spotify/helios/system/AgentArgsBuilder.java
```diff
@@ -0,0 +1,170 @@
...
+    this.httpEndpoint = httpEndpoint;
+    return this;
+  }
+
+  public AgentArgsBuilder adminEndpoint(final String adminEndpoint) {
+    this.adminEndpoint = adminEndpoint;
+    return this;
+  }
+
+  public List<String> build() {
```
davidxia: Should I add some invariants here like `checkNotNull(dockerHost)`? And if so which ones?

helios-system-tests/src/main/java/com/spotify/helios/system/AgentArgsBuilder.java
```diff
@@ -0,0 +1,170 @@
...
+    final ImmutableList.Builder<String> b = ImmutableList.builder();
+    if (verbosity > 0) {
```
davidxia: These `if`s seem ugly. I'm wondering if there's a prettier way.

mattnworb:
I think a much simpler test would be to extract the anonymous class for the "Gauge of a Healthcheck" to it's own class and write a unit test for that, to see what value it returns when the `Healthcheck` returns different values. Then you wouldn't need these descriptors or pojos or tests that make http calls to a servlet.

helios-system-tests/src/main/java/com/spotify/helios/system/SystemTestBase.java
```diff
@@ -602,7 +615,34 @@ protected AgentMain startDefaultAgent(final String host, final String... args)
         dockerPortRange.upperEndpoint()
     );
     argsList.addAll(asList(args));
-    return startAgent(argsList.toArray(new String[argsList.size()]));
+    return startAgentWithArgs(argsList);
+  }
+
+  protected AgentMain startAgentWithArgs(final List<String> args) throws Exception {
+    final AgentMain agentMain = startAgent(args.toArray(new String[args.size()]));
+    waitForAgentToBeFullyUp();
```
davidxia: I think this will fix test failures like DeregisterTest.testJobsArePreservedWhenReregistering()` here https://circleci.com/gh/spotify/helios/3795.

davidxia:
@mattnworb I've added a unit test in the last commit. I left the system tests since they test the master and agent are actually registering the gauges and emitting those metrics. Let me know if these tests are not worth it though.

helios-services/src/main/java/com/spotify/helios/master/MasterService.java
```diff
@@ -186,6 +187,12 @@ public MasterService(final MasterConfig config,
 
     environment.lifecycle().manage(zooKeeperHealthChecker);
     environment.healthChecks().register("zookeeper", zooKeeperHealthChecker);
+
+    // Report health checks as a boolean gauge metric
```
mattnworb: strictly speaking the "boolean" part of these comments are outdated now

davidxia: Ah thanks. For context, I changed it to an Integer because every other gauge is numeric, and I didn't know how to make jackson parse something that could be more than one type.

helios-system-tests/src/main/java/com/spotify/helios/system/SystemTestBase.java
```diff
@@ -602,7 +615,34 @@ protected AgentMain startDefaultAgent(final String host, final String... args)
...
+    return agentMain;
+  }
+
+  protected void waitForAgentToBeFullyUp() throws Exception {
```
mattnworb: this looks identical to `waitForMasterToBeFullUp()`, can you extract the callable so that both functions can use the same code?

davidxia: Sure thing. I might abandon this as it seems to break a bunch of tests...

mattnworb:
If you think the system test adds value I won't quibble with adding it; although I don't think I would have chosen to test that piece of logic myself.

But:
- adding `Gauge` and `Metrics` as `Descriptor` subclasses feels weird to me, since they really aren't
- Instead of making them Descriptors to gain access to easy JSON to POJO serialization/deserialization, I personally find it easier to just use `ObjectMapper` and it's tree API (JsonNode, ObjectNode) in tests
- I think these classes, and AgentArgsBuilder, should be in src/test/java if they are only for testing purposes and never to be used in the actual main source code

davidxia:
Good points. I'll remove the system test. It added too much code for my
taste anyways.

On Fri, Feb 5, 2016 at 10:49 AM, Matt Brown <notifications@github.com>
wrote:

> If you think the system test adds value I won't quibble with adding it;
> although I don't think I would have chosen to test that piece of logic
> myself.
>
> But:
>
>    - adding Gauge and Metrics as Descriptor subclasses feels weird to me,
>    since they really aren't
>    - Instead of making them Descriptors to gain access to easy JSON to
>    POJO serialization/deserialization, I personally find it easier to just use
>    ObjectMapper and it's tree API (JsonNode, ObjectNode) in tests
>    - I think these classes, and AgentArgsBuilder, should be in
>    src/test/java if they are only for testing purposes and never to be used in
>    the actual main source code
>
> —
> <https://github.com/spotify/helios/pull/840#issuecomment-180411079>.
>



-- 
David Xia
Software Engineer
dxia@spotify.com

davidxia:
finally tests passed

mattnworb:
:+1: :+1:

helios-services/src/main/java/com/spotify/helios/agent/AgentService.java
```diff
@@ -303,9 +305,17 @@ public AgentService(final AgentConfig config, final Environment environment)
     if (!config.getNoHttp()) {
 
       environment.healthChecks().register("docker", dockerHealthChecker);
+      environment.healthChecks().register("zookeeper", zkHealthChecker);
+
+      // Report health checks as a gauge metric
+      environment.healthChecks().getNames().forEach(
+          name -> environment.metrics().register("helios." + name + ".ok", (Gauge<Boolean>) () -> {
```
gimaker: `HealthCheckGauge` should probably be used here too.

gimaker:
:+1: aside from my one comment

PR #841, opened 2016-02-01
Support resolution of namespaced agent names in Helios Solo.

negz:
@spotify/helios-team 

Hopefully this PR is reaching its final form. It:
- Fixes the helios solo container's `start.sh` such that namespaced Helios agent names (i.e. `mynamespace.solo.local`) are resolvable by SkyDNS.
- Tweaks the namespaced Helios agent names to end in `.local` for compatibility with our SkyDNS setup.
- Supports specifying arbitrary env vars to be set inside the `HeliosSoloDeployment` container.
- Supports reading said env vars from `helios.conf`
- Updates an existing service discovery integration test to use a `HeliosSoloDeployment`.

The culmination of the above should mean we can specify our whacky Spotify SRV record format as an env var in `spotify-helios-testing`'s `helios.conf` and leave our defaults sane for everyone else. That will look something like:

```
helios.solo.profile : "local"
helios.solo.profiles : {
  local : {
    env : {
        REGISTRAR_HOST_FORMAT: "_spotify-${service}._${protocol}.services.${domain}"
    }
  }
}
```


mattnworb:
Take a look at the tests I added in `HeliosSoloDeploymentTest`, where the `DockerClient` is mocked out and the test asserts the values passed to `dockerClient.createContainer` etc. You could have a test of the default logic that it matches the hard-coded format and another test that `builder.registrationFormat("foo")` gets passed in as expected, etc

mattnworb:
I think a way to override this format with the builder is missing though?

negz:
@mattnworb Whoops - I cut it to move it and never pasted. Fixed.

I'll take a look at the new stuff in `HeliosSoloDeploymentTest`. I actually just spotted `HeliosSoloIT` in the integration tests which AFAIK tests exactly what I want to, but currently doesn't use a `HeliosSoloDeployment` to turn up solo. I might update that too unless there's a reason for it not to.

mattnworb:
In a test of a real container / helios-solo instance you'd have to find a way to test what registrar format helios-solo uses when talking to skydns, either by querying skydns or making sure a second container that does dns lookups for the first one works. 

If you think of the environment variable as the API to helios-solo then I would just test that this class communicates with the API contract correctly myself :smile:

negz:
https://github.com/spotify/helios/blob/master/helios-integration-tests/src/test/java/com/spotify/helios/HeliosSoloIT.java#L182

@mattnworb - If I understand the above we have an existing IT that does something like what you described. It's currently using its own bespoke Helios Solo deployment code, so we should probably either update it to use the HeliosSoloDeployment or just scrap it.

mattnworb:
@negz good catch. I never noticed that test before. I agree on modernizing it

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -423,6 +426,7 @@ public String toString() {
     private DockerHost dockerHost;
     private DockerHost containerDockerHost;
     private String heliosSoloImage = "spotify/helios-solo:latest";
+    private String registrarHostFormat = "_spotify-${service}._${protocol}.services.${domain}";
```
gimaker: :/

gimaker: Well that's unfortunate.

gimaker:
Would it be possible have a saner default, and override this in e.g. spotify-helios-testing somehow? I guess you already considered that option and ruled it out for good reason though?

gimaker:
Otherwise LGTM.

negz:
@gimaker I definitely considered it, but only ruled it out for a good reason if you consider being lazy a good reason. :)

mattnworb:
Since TemporaryJobs [already looks in `helios.conf` and `helios-base.conf` for a bunch of configuration](https://github.com/spotify/helios/blob/master/docs/testing_framework.md#configuration-by-file) it shouldn't be too hard to add registrarHostFormat there also.

The only thing I'd urge if we go with that approach is that it's possible for the internal configuration from spotify-helios-testing (or any user, really) to _default_ to the value we prefer.

negz:
Note I've implemented the `REGISTRAR_HOST_FORMAT` env var by setting it in `helios-base.conf` in spotify-helios-testing per suggestion.

mattnworb:
lgtm

negz:
Argh. Turns out it's not so simple - setting the `REGISTRAR_HOST_FORMAT` in `helios-base.conf` affects TemporaryJobs containers - not the Helios Solo container where it needs to be picked up by the registration plugin. Back to the drawing board. :(

negz:
@spotify/helios-team PTAL. Pretty sure this will actually work now.

mattnworb:
the Conf stuff in 6b80697 looks nice, just a few small comments above

negz:
@spotify/helios-team PTAL. Pretty confident this is all good - just waiting to be blessed by the CircleCI gods.

mattnworb:
:+1:

gimaker:
:tada: :tada:

PR #842, opened 2016-02-02
Allow user to bind admin port on different host

davidxia:
like localhost to protect the endpoint if they so wish.
Default to 0.0.0.0.


davidxia:
@mattnworb @negz @gimaker

mattnworb:
lgtm

PR #844, opened 2016-02-02
Update authentication documentation

gimaker:
* Update the docs to match how the ssh-agent assisted cert
   generation works now.
 * Document the different ways of supplying credentials to the
   CLI and API respectively.

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -4,6 +4,8 @@ When using HTTPS, Helios client can present an X.509 client certificate to verif
 the user's identity. By presenting a digital signature to the Helios master, the
 client proves that it is who it claims to be.
 
+Note that only the client supports HTTPS and client certificates. The master does not. The cluster administrator must add the missing pieces that validate certificates and act accordingly. Suggestions for how such a setup might look can be found below.
+
 Generally this type of mutual TLS authentication requires the user to have their
 own SSL certificate issued by some certificate authority (CA). Often, this
 will be your company or organization's CA. However, for users who do not have a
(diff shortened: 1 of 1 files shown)


gimaker:
@mattnworb @rohansingh @davidxia @negz

gimaker:
I struggled with this, so please suggest changes or modify and add commits as you see fit.

docs/authentication.md
```diff
@@ -4,6 +4,8 @@ When using HTTPS, Helios client can present an X.509 client certificate to verif
 the user's identity. By presenting a digital signature to the Helios master, the
 client proves that it is who it claims to be.
 
+Note that Helios itself only contains functionality that lets the client present certificates. The Helios masters do not verify certificates themselves. The cluster administrator must add the missing pieces that validate certificates and act accordingly.
```
davidxia: We can add:

Only the Helios client supports HTTPS. The master doesn't. One needs a web server like nginx sitting in front to handle the TLS handshake and termination.

This web server can validate the client cert itself or by asking another service that checks a keystore like ldap. We could even open-source the x509-verifier and include sample nginx configs.

davidxia: Oh I see you have that below.

docs/authentication.md
```diff
@@ -32,35 +34,48 @@ The Helios client will now need to present a client certificate to authenticate.
...
+ * Set the `HELIOS_CERT_PATH` environment variable to a path containing the pre-issued X.509 certificate and corresponding private key, named `cert.pem` and `key.pem` respectively, or
+ * Let the Helios client generate a temporary X.509 certificate that is signed with your private SSH key via ssh-agent, as described above.
+
+The latter is typically not suitable for non-interactive use cases (e.g. a script run periodically by cron). If the `HELIOS_CERT_PATH` environment variable is set this method takes precedence.
+
+### Authenticating with the Helios API
+
+When you use the Helios API (via the `HeliosClient` class) you have the same options as when using the CLI, as explained above. You can also explicitly set the path where the certificate and key is via the `HeliosClient.Builder.setClientCertificatePath()` method. If a certificate path is set explicitly the `HELIOS_CERT_PATH` is ignored.
+
+Again, for non-interactive use cases relying on ssh-agent is typically not a suitable solution so we recommend that you specify a path either explicitily or via the `HELIOS_CERT_PATH` environment variable.
```
davidxia: typo in `explicitily`.

docs/authentication.md
gimaker: Maybe just remove that whole paragraph since it is indeed covered below?

gimaker: The reason that I put it there is that I wanted to state that Helios doesn't offer a complete authentication solution early on in the documentation.

davidxia: But it doesn't mention that the master doesn't support HTTPS, so maybe we can leave something about that here?

gimaker: Reworded it a bit but left it in there.

docs/authentication.md
```diff
@@ -32,35 +34,48 @@ The Helios client will now need to present a client certificate to authenticate.
...
+(you probably do if you use `ssh`), Helios can generate an X.509 certificate for
 you on the fly. Here's how it works:
 
 1. The Helios client checks for a running SSH agent, and if one is found, asks
    for a list of your public SSH keys. Private keys are never exposed by
    ssh-agent.
 
-2. The Helios client generates a self-signed X.509 certificate containing your
-   SSH public key and username, which it sends to the HTTPS server.
+2. The Helios client generates a temporary X.509 certificate containing your username.
```
mattnworb: We should note where this username comes from. Whether it is `System.getProperty("user.name")` or some environment variable or CLI switch.

docs/authentication.md
```diff
@@ -32,35 +34,48 @@ The Helios client will now need to present a client certificate to authenticate.
...
+   It then signs it with your private SSH key via ssh-agent and sends it to the HTTPS server.
 
-3. To prove that you own the corresponding private key, the Helios client asks
-   ssh-agent to cryptographically hash and sign part of the TLS handshake with
-   the server.
+3. The server looks up the public key for the username specified in the certificate, and
+   uses it to verify that the certificate was signed by the corresponding private key. If the
+   signature can be verified we know that the user is who they claim to be. Each user
+   essentially acts as a certificate authority.
```
mattnworb: I think we should mention that this description of what the server does with the certificate is completely specific to the helios-master installation, i.e. however the admins have set it up themselves.

mattnworb:
Other than above, looks good to me, nice job :+1: :ok_hand:

docs/authentication.md
gimaker: Yeah, but it's tricky given that it depends on whether you're using the CLI or the API. For the API you have to set the user explicitly, and for the CLI it comes from the `--username` arg which defaults to `System.getProperty("user.name")`.

gimaker: Tricky is the wrong word -- but explaining it is quite lengthy and wordy. Still worth it?

mattnworb: yea, just copy and paste the second sentence of your comment above :smile:

gimaker: I added a similar sentence below the actual list - if you approve?

mattnworb: :+1:

docs/authentication.md
```diff
@@ -32,35 +34,50 @@ The Helios client will now need to present a client certificate to authenticate.
...
+ * Set the `HELIOS_CERT_PATH` environment variable to a path containing the pre-issued X.509 certificate and corresponding private key, named `cert.pem` and `key.pem` respectively, or
+ * Let the Helios client generate a temporary X.509 certificate that is signed with your private SSH key via ssh-agent, as described above.
+
+The latter is typically not suitable for non-interactive use cases (e.g. a script run periodically by cron). If the `HELIOS_CERT_PATH` environment variable is set this method takes precedence.
+
+### Authenticating with the Helios API
+
+When you use the Helios API (via the `HeliosClient` class) you have the same options as when using the CLI, as explained above. You can also explicitly set the path where the certificate and key is via the `HeliosClient.Builder.setClientCertificatePath()` method. If a certificate path is set explicitly the `HELIOS_CERT_PATH` is ignored.
+
+Again, for non-interactive use cases relying on ssh-agent is typically not a suitable solution so we recommend that you specify a path either explicitly or via the `HELIOS_CERT_PATH` environment variable.
```
gimaker: @davidxia @rohansingh Do you have the commands we typically use to generate a cert available somewhere for me to copy-paste?

Might be nice to put a complete example in there. "Generate a cert/key pair by doing X. Then create your HeliosClient like this ..."

davidxia:
```
openssl req -new -newkey rsa:2048 -days 3650 -nodes -x509 -keyout key.pem \
     -out unsigned_cert.pem -subj "/CN=jenkins" -sha256
```

gimaker: That only creates an unsigned cert, no? What about the (self-)signing part?

gimaker: Thinking about it a little more, it probably doesn't make sense to give an exampled detail here -- how you should do it will depend on the specific setup. E.g. we use self-signed certs and maintain a list of trusted certs, but other's might have proper PKI setup.

PR #848, opened 2016-02-04
Fix the default Prober to support UDP probing as well as TCP

pehrs:
I made a suggested change to the DefaultProber to support UDP ports as well as TCP
The current implementation will hang when probing UDP ports.
This implementation mimics the nmap behavior by sending a "PING" message to the port and wait for timeout (OK) or port not reachable (NOT OK)

### DIFF
--- a/helios-testing/src/test/java/com/spotify/helios/testing/UdpProberTest.java
+++ b/helios-testing/src/test/java/com/spotify/helios/testing/UdpProberTest.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) 2014 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.helios.testing;
+
+import com.spotify.helios.common.descriptors.Job;
+import com.spotify.helios.common.descriptors.JobId;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import java.util.Map;
+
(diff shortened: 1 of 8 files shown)


davidxia:
@pehrs Thanks for the PR. Just curious what's your use case?

mattnworb:
I kicked off a rebuild of the CircleCI build as the timeout looks unrelated to this change.

mattnworb:
LGTM, but I think it would be nice if a test could be added that verifies that the udp prober works in an actual TemporaryJob

pehrs:
@davidxia I have a service that exposes an UDP port and the default Prober implementation assumes that ports are TCP no UDP

davidxia:
@mattnworb @pehrs I added a test for UDP probing.

helios-testing/src/test/java/com/spotify/helios/testing/UdpProberTest.java
```diff
@@ -0,0 +1,64 @@
...
+      job = temporaryJobs.job()
+          .image(ALPINE)
+          .command(asList("nc", "-p", "4711", "-lu"))
+          .port("default", 4711, "udp")
+          .deploy(testHost1);
+    }
+
+    @After
+    public void tearDown() {
+      // The TemporaryJobs Rule above doesn't undeploy the job for some reason...
```
davidxia: Any idea about this?

davidxia:
@mattnworb @pehrs I added a test for this. Let me know if it looks OK.

mattnworb:
Thanks for adding the test. I am confused though on how it asserts that the UDP port is "probed"? Simply by asserting that the job is running ok?

Also, why the indirection in the test where the outer `@Test` calls `testResult()` in a nested class that looks a lot like a test itself?

davidxia:
Yup. By specifying `.port("default", 4711, "udp")` here https://github.com/spotify/helios/pull/848/files#diff-6da031aab2d8f23a98de0f1b4163735cR62, the prober will send a UDP packet that will be ignored instead of a port not reachable.

A temp job will block on the `deploy()` until the probing for all specified ports is successful and will throw an error if any port fails to be probed.

As for the indirection, I wanted to use certain variables defined in `TemporaryJobsTestBase` here https://github.com/spotify/helios/blob/master/helios-testing/src/test/java/com/spotify/helios/testing/TemporaryJobsTestBase.java#L35, and I think I have to nest the test in order to use them. Let me know if I'm wrong or if there's a better way.

mattnworb:
Nah, I don't know a better way

:+1:

mattnworb:
@pehrs this should be available in `helios-testing:0.8.753` now.

PR #851, opened 2016-02-05
Time ZK calls via ReportingZooKeeperClient

davidxia:
Make ReportingZooKeeperClient call ZooKeeperModelReport.time()
which gets a Timer from ZooKeeperMetrics and uses that to time
each ZK call.


davidxia:
@spotify/helios-team

davidxia:
Let me know if tests like this are worth having

```
@Test
  public void testEnsurePath() throws Exception {
    final ReportingZooKeeperClient reportingClient =
        new ReportingZooKeeperClient(client, reporter, TAG);
    reportingClient.ensurePath("path");
    verify(reporter).time(eq(TAG), eq("ensurePath"), Matchers.any(ZooKeeperCallable.class));
  }
```

helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/ReportingZooKeeperClient.java
```diff
@@ -51,122 +51,86 @@ public ReportingZooKeeperClient(final ZooKeeperClient client,
...
-      throw e;
-    }
+      return null;
+    });
   }
 
   @Override
   public void ensurePath(String path, boolean excludingLast) throws KeeperException {
-    try {
+    reporter.time(tag, "ensurePath", (ZooKeeperCallable<Void>) () -> {
```
mattnworb: I don't think these casts to `(ZooKeeperCallable<Void>)` are necessary; or at least they seem to be in some spots in the diff and not others.

helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/ReportingZooKeeperClient.java
```diff
@@ -235,12 +177,7 @@ public Stat exists(String path) throws KeeperException {
 
   @Override
   public ZooKeeper.States getState() throws KeeperException {
-    try {
-      return client.getState();
-    } catch (KeeperException e) {
-      reporter.checkException(e, tag, "getState");
-      throw e;
-    }
+    return reporter.time(tag, "ensurePath", client::getState);
```
mattnworb: should be "getState" here

helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/ReportingZooKeeperClient.java
davidxia: You're right. IDE didn't catch these.

helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/ReportingZooKeeperClient.java
```diff
@@ -267,22 +204,14 @@ public CuratorFramework getCuratorFramework() {
...
 
   @Override
   public List<ACL> getAcl(final String path) throws KeeperException {
-    try {
-      return client.getAcl(path);
-    } catch (KeeperException e) {
-      reporter.checkException(e, tag, "getAcl");
-      throw e;
-    }
+    return reporter.time(tag, "ensurePath", () -> client.getAcl(path));
```
mattnworb: "getAcl"

mattnworb:
My own two cents is that it is not a very useful test of `ReportingZooKeeperClient` to verify that it called a certain method of the underlying client. It's more useful to test the values that those methods return.

What I would test though instead would be the new `Timer time(name, callable)` method in `ZooKeeperMetricsImpl` - you could have a test that asserts that a call like `.time("foo", ..)` adds a timer to the registry with the expected full name using the prefix.

mattnworb:
also yay for java 8 language features :tada:

helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/ZooKeeperCallable.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.apache.zookeeper.KeeperException;
+
+/**
+ * A task that returns a result and may throw a {@link KeeperException}.
+ * We use this class when we don't want a {@link java.util.concurrent.Callable}
+ * that throws an Exception.
+ */
+@FunctionalInterface
+public interface ZooKeeperCallable<T> {
```
gimaker: This is irrelevant for the function of the code but I would probably just have made this an inner class^Winterface of `ZooKeeperModelReporter` to reduce clutter - given how "big" this interface is :)

helios-services/src/main/java/com/spotify/helios/servicescommon/statistics/NoopZooKeeperMetrics.java
```diff
@@ -17,7 +17,14 @@
...
 
+import com.codahale.metrics.Timer;
+
 public class NoopZooKeeperMetrics implements ZooKeeperMetrics {
   @Override
   public void zookeeperTransientError() {}
+
+  @Override
+  public Timer timer(String name) {
+    return new Timer();
```
gimaker: We shouldn't use a real timer, that actually gathers and keeps data, for a "no op" implementation. Unfortunately `Timer` isn't an interface so we can't really return a no-op timer...

gimaker: If you changed this interface slightly you could do an actual no-op implementation. E.g. something like:

```
void updateTimer(String name, final long durationMillis);
```

davidxia: It'd be nice to actually have a no op, but the nice thing about the current interface is that a Timer takes care of timing the duration. It's a bit more unwieldy if the caller has to handle timing a Callable when the Timer class already does this.

helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/ZooKeeperCallable.java
davidxia: thanks

helios-services/src/main/java/com/spotify/helios/servicescommon/statistics/NoopZooKeeperMetrics.java
davidxia: Hm, it might not be too bad. Let me try

PR #852, opened 2016-02-08
Fix javadoc errors for java 8

davidxia:


mattnworb:
Ah I didn't realize the errors you were having were these kinds - I think you can configure the javadoc plugin to not fail on these.

The more annoying ones I've seen are where [the new java 8 javadoc tool validates the html in any javadoc comment](http://blog.joda.org/2014/02/turning-off-doclint-in-jdk-8-javadoc.html)

PR #853, opened 2016-02-09
allow alternate java installation in startup scripts

mattnworb:
Changes `bin/helios-agent` and `bin/helios-master` to respect a `$JAVA`
environment variable which contains the path to the java command to run
the helios process with.

If the environment variable is not set, then the scripts fall back to (the
prior behavior of) a simple `java` command.

This can be useful for environments where the helios agent and/or master
should be run with a version of Java that is not the default version on
the host.

### DIFF
--- a/bin/helios-master
+++ b/bin/helios-master
@@ -34,7 +34,14 @@ if [ $JMXPORT ]; then
           -Dcom.sun.management.jmxremote.authenticate=false )
 fi
 
-exec java \
+# Allow the path to the java installation to be specified in the JAVA environment variable.
+# This can be useful if you want to run helios-agent using a different version of Java than the
+# default version on the host (possibly set up through update-alternatives)
+if [ -z $JAVA ]; then
+    JAVA=java
+fi
+
+exec $JAVA \
     $HELIOS_MASTER_JVM_OPTS \
     "${args[@]}" \
     -Djava.net.preferIPv4Stack=true \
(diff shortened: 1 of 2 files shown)


mattnworb:
for instance if the `$DEFAULTFILE` that the upstart scripts source contains a declaration of `JAVA=`, then this should respect that path for launching java when running helios-agent or helios-master.

mattnworb:
@spotify/helios-team

gimaker:
:+1: :100:

davidxia:
:+1:

bin/helios-master
```diff
@@ -34,7 +34,14 @@ if [ $JMXPORT ]; then
           -Dcom.sun.management.jmxremote.authenticate=false )
 fi
 
-exec java \
+# Allow the path to the java installation to be specified in the JAVA environment variable.
+# This can be useful if you want to run helios-agent using a different version of Java than the
+# default version on the host (possibly set up through update-alternatives)
+if [ -z $JAVA ]; then
+    JAVA=java
+fi
```
negz: I know I'm a bit late to the game here but for future reference there's BASH fanciness for this:
```bash
messremb(negz@~)$ echo $JAVA
messremb(negz@~)$ echo ${JAVA:-/usr/bin/java}
/usr/bin/java
messremb(negz@~)$ export JAVA=/usr/bin/megajava
messremb(negz@~)$ echo ${JAVA:-/usr/bin/java}
/usr/bin/megajava
```

mattnworb: Thanks. I remembered about this after making the commit; although I like the more verbose style here though to be extra clear.

PR #854, opened 2016-02-10
Remove commit hashes from Docker image tags

rohansingh:
Remove the Git commit hashes from the helios-agent and helios-master images
that we build. Instead match the behavior for helios-solo, wherein the tag
is just the same as the Helios version.

This is in preparation for pushing the images for public consumption.

### DIFF
--- a/helios-services/pom.xml
+++ b/helios-services/pom.xml
@@ -12,8 +12,8 @@
   <packaging>jar</packaging>
 
   <properties>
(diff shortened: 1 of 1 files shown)


rohansingh:
@spotify/helios-team

negz:
LGTM

rohansingh:
Test failure seems unrelated. I'm merging.

mattnworb:
:+1:

PR #856, opened 2016-02-10
Fix race in ZooKeeperAclTest

davidxia:
testAgentCreatedNodesHaveAcls(). Check for znode at
/config/hosts/<hostname> instead of /status/hosts/<hostname>/up.
We call awaitHostRegistered() before this, but the up znode
may not have been written yet as that happens after the
agent writes the other znodes. See AgentZooKeeperRegistrar.tryToRegister().

### DIFF
--- a/helios-system-tests/src/main/java/com/spotify/helios/system/ZooKeeperAclTest.java
+++ b/helios-system-tests/src/main/java/com/spotify/helios/system/ZooKeeperAclTest.java
@@ -65,7 +65,7 @@ public class ZooKeeperAclTest extends SystemTestBase {
 
     final CuratorFramework curator = zk().curatorWithSuperAuth();
 
-    final String path = Paths.statusHostUp(TEST_HOST);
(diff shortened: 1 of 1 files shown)


mattnworb:
Doesn't the call to `awaitHostRegistered` earlier in the test take care of this? It seems like you would just be reducing the window of a race here by checking an earlier-written path, instead of removing it altogether. Maybe `awaitHostRegistered` should not just check that the host is in `hosts -q` but also that the status is UP.

davidxia:
awaitHostRegistered() checks that the hostname is returned from ZooKeeperMasterModel.listHost(). listHost() returns the znode children of `/config/hosts`. Since AgentZooKeeperRegistrar.tryToRegister() writes `/config/hosts/<hostname>`, this should guarantee that when awaitHostRegistered() returns, `/config/hosts/<hostname>` will exist.

davidxia:
@spotify/helios-team

mattnworb:
ah I understand now, makes sense. :+1:

PR #858, opened 2016-02-11
Minor simplifications to HeliosSoloDeploymentIT

davidxia:


mattnworb:
this change would make it harder to debug if there was > 1 job running and you were trying to figure out which they were, although maybe there are more logs that would tell you that instead

davidxia:
ah you're right.

PR #859, opened 2016-02-11
Log more DefaultZooKeeperClient.deleteRecursive()

davidxia:


davidxia:
@spotify/helios-team

helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/DefaultZooKeeperClient.java
```diff
@@ -241,9 +241,14 @@ public void deleteRecursive(final String path) throws KeeperException {
       final CuratorTransactionFinal t = client.inTransaction().check().forPath(path).and();
       for (final String node : reverse(nodes))  {
         t.delete().forPath(node).and();
+        log.info("Added delete {} to transaction", node);
```
gimaker: Eh... isn't this a bit *too* verbose?
```
```

gimaker: The general pattern seems to be that DefaultZooKeeperClient does not log anything at all, sans failed cluster-id checks. Instead the caller is responsible for logging, which I think makes sense since whether an exception is a failure or merely expected behaviour depends.

davidxia: It's a bit verbose. I'll revert this change and just run it on a personal branch.

PR #864, opened 2016-02-17
Label Docker images with Helios version

rohansingh:
Can be useful for figuring out what version of Helios the latest images
contain. For example:

    $ docker inspect -f \
        '{{index .Config.Labels "com.spotify.helios.version"}}' \
        spotify/helios-agent:latest
    0.8.0-SNAPSHOT


rohansingh:
@spotify/helios-team

mattnworb:
you could move the duplicated `<labels>` up out of the `<executions>` section and into the parent `<configuration>` of the plugin, so you don't have to duplicate them, but otherwise :+1:

PR #865, opened 2016-02-19
Don't pull the probe image if we already have it.

negz:
Reviewers: @mattnworb 
Peanuts: @spotify/helios-team 

Its freshness is not a concern.

### DIFF
--- a/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
@@ -31,6 +31,7 @@ import com.spotify.docker.client.DockerCertificateException;
 import com.spotify.docker.client.DockerClient;
 import com.spotify.docker.client.DockerException;
 import com.spotify.docker.client.DockerHost;
+import com.spotify.docker.client.ImageNotFoundException;
 import com.spotify.docker.client.messages.ContainerConfig;
 import com.spotify.docker.client.messages.ContainerCreation;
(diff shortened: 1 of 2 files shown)


davidxia:
Just a general question from me: doesn't `docker pull` first compare the sha256 hashes of the local and remote layers an image is made of and only pulls layer content from a remote registry if any of the hashes are different?

gimaker:
LGTM but any chance we can add tests for this?

mattnworb:
@davidxia I think so. A problem arises though if the network connection to Docker Hub has an issue; for some test users it ended up failing their build and tests, even though they had the alpine image already.

@negz nice change. Agree with Staffan on the test but otherwise :+1:

davidxia:
:+1:

mattnworb:
:+1: for the unit tests

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -245,6 +247,17 @@ private String checkDockerAndGetGateway() throws HeliosDeploymentException {
     return gateway;
   }
 
+  private void pullIfAbsent(final String image) throws DockerException, InterruptedException {
+    try {
+      dockerClient.inspectImage(image);
+      log.info("helios-solo image {} is present. Not pulling it.", image);
```
davidxia: derp, how did we miss this? The log messages and `pull(PROBE_IMAGE)` are wrong. PR here to fix: https://github.com/spotify/helios/pull/918

PR #868, opened 2016-02-23
Randomly allocate dynamic ports

negz:
@spotify/helios-team 

Please let me know if you think more unit tests would be valuable.

### DIFF
--- a/helios-system-tests/src/main/java/com/spotify/helios/system/MultiplePortJobTest.java
+++ b/helios-system-tests/src/main/java/com/spotify/helios/system/MultiplePortJobTest.java
@@ -45,7 +45,9 @@ import static com.spotify.helios.common.descriptors.TaskStatus.State.RUNNING;
 import static java.util.Arrays.asList;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 
 public class MultiplePortJobTest extends SystemTestBase {
 
@@ -56,7 +58,10 @@ public class MultiplePortJobTest extends SystemTestBase {
   public void test() throws Exception {
     startDefaultMaster();
 
+    // 'Reserve' a 2 port wide range of ports.
     final Range<Integer> portRange = temporaryPorts.localPortRange("agent1", 2);
+
+    // Start an agent using the aforementioned 2 port wide range.
     final AgentMain agent1 = startDefaultAgent(testHost(), "--port-range=" +
                                                           portRange.lowerEndpoint() + ":" +
                                                           portRange.upperEndpoint());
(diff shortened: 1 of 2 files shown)


negz:
@mattnworb You mentioned you'd also taken a pass at refactoring `PortAllocator`. Any suggestions as to how I could clean it up further?

mattnworb:
@negz on [this branch](https://github.com/mattnworb/helios/commits/new-port-allocation) I started refactoring the PortAllocator by extracting an interface for it. My idea was to keep the original implementation as "SimplePortAllocator" and then create a variant that could do randomization like this. I was thinking then that the choice of which allocation could be done via a CLI argument to the agent. I didn't actually get to the point of creating another implementation yet though.

Afterwards I realized that a better way would be to just extract the "dynamic port allocation" part to an interface, and leave the static port assignment in PortAllocator (as it seems pretty uncontroversial and unlikely to change strategies - that is why they are static).

But perhaps having multiple implementations and making the dynamic port allocation "strategy" a CLI argument is way overkill.

mattnworb:
Instead of shuffling the list on each call to `allocate`, I think it would be simpler in some ways to shuffle it once at construction-time and maintain an index into the list (like the old implementation did) so that each new call to allocate just moves the index up (and checks if it can use the value in `potentialPorts[index]`

gimaker:
> Instead of shuffling the list on each call to allocate, I think it would be simpler in some ways to shuffle it once at construction-time and maintain an index into the list (like the old implementation did) so that each new call to allocate just moves the index up (and checks if it can use the value in potentialPorts[index]

+1 on this. Doing that would also avoid what the old comment in the code says:

> Index for port allocation. Reused between allocations so we do not immediately reuse ports.

With a big port range it's unlikely that ports are immediately unused but possible.

negz:
@mattnworb @gimaker SGTM - I've already updated the code to take @mattnworb's suggestion into account. Just need to figure out that failing integration test before I update the PR.

negz:
@spotify/helios-team PTAL.

helios-system-tests/src/main/java/com/spotify/helios/system/MultiplePortJobTest.java
```diff
@@ -92,14 +92,22 @@ public void test() throws Exception {
...
       assertNotNull(jobId2);
       deployJob(jobId2, testHost());
       final TaskStatus firstTaskStatus2 = awaitJobState(client, testHost(), jobId2, RUNNING,
                                                         LONG_WAIT_SECONDS, SECONDS);
 
-      assertEquals(expectedMapping1, firstTaskStatus1.getPorts());
-      assertEquals(expectedMapping2, firstTaskStatus2.getPorts());
+      // Verify we allocated dynamic ports from within the specified range.
+      assertTrue(portRange.contains(firstTaskStatus1.getPorts().get("foo").getExternalPort()));
+      assertTrue(portRange.contains(firstTaskStatus2.getPorts().get("foo").getExternalPort()));
```
mattnworb: not really important but a small tip: I've taken to writing checks like this using assertThat instead, like

```java
assertThat(portRange, hasEntry(firstTaskStatus2.getPorts().get("foo").getExternalPort()));
```

because when it fails, junit will give you an error message like "Expected {4711, 4722} to have entry 1234" (or whatever), whereas `assertTrue(something that is false)` just reports an empty and unhelpful AssertionError

mattnworb: or i guess if the thing you are really testing is that `firstTaskStatus2.getPorts().get("foo").getExternalPort()` is in some expected range it would be nicer to reverse it like

```java
assertThat(blah.getExternalPort(), isIn(portRange))
```
which reads nicer when it fails (something like "expected N to be one of {1, 2, 3, 4}")

mattnworb:
:+1: looks good to me!

helios-system-tests/src/main/java/com/spotify/helios/system/MultiplePortJobTest.java
negz: These patterns are handy to know, and I'd prefer to use them, but I don't think I can due to the fact that `portRange` is a Guava `Range` and not a `Collection`. I could probably do a song and dance to convert it into some kind of `Collection`, but I think that would be even messier.

PR #873, opened 2016-02-25
Fix OldJobReaper logging

davidxia:


gimaker:
:+1:

helios-services/src/main/java/com/spotify/helios/master/OldJobReaper.java
```diff
@@ -46,12 +49,14 @@
   private static final Clock SYSTEM_CLOCK = new SystemClock();
   private static final long INTERVAL = 1;
   private static final TimeUnit INTERVAL_TIME_UNIT = TimeUnit.DAYS;
+  private static final DateFormat DATE_FORMATTER =
+      new SimpleDateFormat("EEE, d MMM yyyy HH:mm:ss Z");
```
mattnworb: SimpleDateFormat is not thread-safe. In this case I don't think it matters since only one thread calls `runOneIteration` at a time, but I think it's best to avoid using/declaring it in a way that could later become problematic if more than one thread was to be used here.

davidxia: Good point. I've created a new PR using the thread-safe jodatime here https://github.com/spotify/helios/pull/874

PR #874, opened 2016-02-26
Use thread-safe joda-time for OldJobReaper logging

davidxia:


mattnworb:
nice, didn't realize the joda-time formatter was thread-safe itself.

in retrospect this was pretty silly of me to nag about

:+1:

PR #878, opened 2016-03-01
Include username in X509 cert cache key

rohansingh:
Previously, the cached certificate would be used even if the username was
different. For example, if a user ran `helios -u A` and then `helios -u B`,
the cached certificate with `/UID=A` would still be used.


gimaker:
checkstyle error, otherwise :+1:

davidxia:
:+1:

mattnworb:
:+1:

mattnworb:
@rohansingh merge?

PR #879, opened 2016-03-01
fix NPE in master zookeeper registrar

mattnworb:
The masters do not use a CountDownLatch to coordinate the startup of
services that should only write to Zookeeper after successful
registration.

There is a small bug in the ZooKeeperRegistrarService constructor that
takes the `Optional<CountDownLatch>` reference from the builder, but the
builder uses `null` as a default value on accident.


davidxia:
:+1:

mattnworb:
@spotify/helios-team

PR #881, opened 2016-03-01
add test of deployment group host matching logic

mattnworb:
... and also simplified the loop that matches host labels to DG selectors.

As a part of looking through this area of the code base to learn better how it works, thought to add a test to understand how the labeled loop here works, and then thought it could be a little bit simpler with no loss in behavior.


mattnworb:
@spotify/helios-team

mattnworb:
I find it comical that this causes the code coverage to go down

gimaker:
There's already a `HostMatcherTest` class has tests for `RollingUpdateService.HostMatcher`

gimaker:
https://github.com/spotify/helios/blob/master/helios-services/src/test/java/com/spotify/helios/rollingupdate/HostMatcherTest.java

mattnworb:
@gimaker d'oh, thanks. I'll refactor that test to add my additional cases.

mattnworb:
@gimaker consolidated my new test with the existing one

gimaker:
:+1:

PR #882, opened 2016-03-02
ensure ports are waited on when using TemporaryJobs.jobWithConfig()

mattnworb:
If you construct a TemporaryJob like:

```java
job = temporaryJobs.jobWithConfig("helios_job_config.json")
  .imageFromBuild()
  .deploy();
```

and the job config looks like:

```json
{
  "ports": {
    "http": { "internalPort" : 8080}
  }
}
```

then TemporaryJobs does not check when deploying to see if the job's
external port can be reached in the same way that it would if the
TemporaryJob was constructed like:

```
job = temporaryJobs.job()
  .imageFromBuild()
  .port("http", 8080)
  .deploy();
```

This difference in behavior is because the `waitPorts` set in
TemporaryJobBuilder is not updated in the same way when passed a
`Job.Builder` that has ports defined as it is when
`TemporaryJobBuilder.port(..)` is called.

### DIFF
--- a/helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobBuilder.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobBuilder.java
@@ -66,7 +66,7 @@ public class TemporaryJobBuilder {
   private final String jobNamePrefix;
   private final Map<String, String> env;
   private final TemporaryJobReports.ReportWriter reportWriter;
-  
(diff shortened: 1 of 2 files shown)


mattnworb:
@spotify/helios-team

gimaker:
:+1: good catch. checkstyle errors though

davidxia:
:+1:

PR #884, opened 2016-03-09
Make HeliosSoloDeployment clean up leftover jobs

davidxia:
Sometimes TemporaryJobs doesn't undeploy its jobs.
This might be due to a race condition between tearing down
of TemporaryJobs and HeliosSoloDeployment.

Make HeliosSoloDeployment undeploy and remove leftover
temp jobs.

### DIFF
--- a/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
@@ -87,6 +87,7 @@ public class HeliosSoloDeployment implements HeliosDeployment {
   public static final String HELIOS_SOLO_PROFILE = "helios.solo.profile";
   public static final String HELIOS_SOLO_PROFILES = "helios.solo.profiles.";
   public static final int HELIOS_MASTER_PORT = 5801;
+  private static final int DEFAULT_WAIT_SECONDS = 30;
 
   private final DockerClient dockerClient;
   /** The DockerHost we use to communicate with docker */
@@ -102,11 +103,13 @@ public class HeliosSoloDeployment implements HeliosDeployment {
   private final HostAndPort deploymentAddress;
   private final HeliosClient heliosClient;
   private boolean removeHeliosSoloContainerOnExit;
+  private final int jobUndeployWaitSeconds;
 
   HeliosSoloDeployment(final Builder builder) {
     this.heliosSoloImage = builder.heliosSoloImage;
     this.pullBeforeCreate = builder.pullBeforeCreate;
     this.removeHeliosSoloContainerOnExit = builder.removeHeliosSoloContainerOnExit;
+    this.jobUndeployWaitSeconds = builder.jobUndeployWaitSeconds;
 
     final String username = Optional.fromNullable(builder.heliosUsername).or(randomString());
 
@@ -458,7 +461,7 @@ public class HeliosSoloDeployment implements HeliosDeployment {
             log.info("Undeploy response for job {} is {}.", jobId, undeployResponse.getStatus());
 
             log.info("Waiting for job {} to actually be undeployed...", jobId);
-            awaitJobUndeployed(heliosClient, host, jobId, 400, TimeUnit.SECONDS);
+            awaitJobUndeployed(heliosClient, host, jobId, jobUndeployWaitSeconds, TimeUnit.SECONDS);
             log.info("Job {} successfully undeployed.", jobId);
           }
         }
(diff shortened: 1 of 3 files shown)


davidxia:
TODO: add test

davidxia:
@gimaker @mattnworb

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -410,10 +433,78 @@ public void close() {
...
+
+        for (Map.Entry<String, TaskStatus> status : statuses.entrySet()) {
+          final String host = status.getKey();
+          if (status.getValue().getGoal().equals(Goal.START)) {
+            log.info("Job {} is still set to START on host {}. Undeploying it now.", jobId, host);
+            final JobUndeployResponse undeployResponse = heliosClient.undeploy(jobId, host).get();
+            log.info("Undeploy response for job {} is {}.", jobId, undeployResponse.getStatus());
+
+            log.info("Waiting for job {} to actually be undeployed...", jobId);
+            awaitJobUndeployed(heliosClient, host, jobId, 400, TimeUnit.SECONDS);
```
mattnworb: 400 seconds??

mattnworb: I think it would be nice if this was customizable in any case, for instance when constructing the HeliosSoloDeployment instance (there are other builder args already)

mattnworb:
I think the test would be more reliable as a unit test, to assert that each job returned by `listJobs` has `stop` called on it

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
davidxia: Yea, that's too long. I'll make it configurable.

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -410,10 +436,78 @@ public void close() {
   }
 
   /**
+   * Undeploy jobs left over by {@link TemporaryJobs}. TemporaryJobs should clean these up,
+   * but sometimes a few are left behind for whatever reason.
+   */
+  private void undeployLeftoverJobs() {
```
gimaker: Hmm. This sounds like admitting defeat. "It doesn't work, so let's do it again!" Not sure I like it -- undeploying a second time because the first undeploy failed for reasons unknown.

gimaker: (But I don't have a big problem with this, it's fine -- waiting for a job be undeployed when without calling undeploy (defensively) seems like a worse option :))

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -410,10 +436,78 @@ public void close() {
...
+
+        for (Map.Entry<String, TaskStatus> status : statuses.entrySet()) {
+          final String host = status.getKey();
+          if (status.getValue().getGoal().equals(Goal.START)) {
+            log.info("Job {} is still set to START on host {}. Undeploying it now.", jobId, host);
+            final JobUndeployResponse undeployResponse = heliosClient.undeploy(jobId, host).get();
+            log.info("Undeploy response for job {} is {}.", jobId, undeployResponse.getStatus());
+
+            log.info("Waiting for job {} to actually be undeployed...", jobId);
+            awaitJobUndeployed(heliosClient, host, jobId, jobUndeployWaitSeconds, TimeUnit.SECONDS);
```
gimaker: The await should be outside the if statement, right?

I.e. the job might previously have been told to undeploy but still be running.

davidxia: yup

davidxia:
@mattnworb I added a couple of unit tests. I left the original test. Let me know if I should delete that one and whether to add more unit tests for the various conditionals in `undeployLeftoverJobs()`.

helios-testing/src/test/java/com/spotify/helios/testing/HeliosSoloDeploymentTest.java
```diff
@@ -183,4 +231,144 @@ public void testDoesPullAbsentProbeImage() throws Exception {
...
+
+  @Test
+  public void testUndeployLeftoverJobs() throws Exception {
+    final HeliosSoloDeployment solo = (HeliosSoloDeployment) HeliosSoloDeployment.builder()
+        .dockerClient(dockerClient)
+        .heliosClient(heliosClient)
+        .build();
+
+    final SettableFuture<Map<JobId, Job>> jobsFuture = SettableFuture.create();
+    jobsFuture.set(ImmutableMap.of(JOB1.getId(), JOB1, JOB2.getId(), JOB2));
```
mattnworb: could also be `Futures.immediateFuture(...)`

mattnworb:
test looks good to me

PR #887, opened 2016-03-10
Fix HeliosSoloDeployment.undeployLeftoverJobs()

davidxia:
See if there are jobs running on any helios agent. If we are using
TemporaryJobs, that class should've undeployed them at this point.
Any jobs still running at this point have only been partially
cleaned up. We look for jobs via hostStatus() because the job
may have been deleted from the master, but the agent may still not
have had enough time to undeploy the job from itself.

### DIFF
--- a/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
@@ -46,9 +46,7 @@ import com.spotify.docker.client.messages.PortBinding;
 import com.spotify.helios.client.HeliosClient;
 import com.spotify.helios.common.descriptors.Goal;
 import com.spotify.helios.common.descriptors.HostStatus;
-import com.spotify.helios.common.descriptors.Job;
 import com.spotify.helios.common.descriptors.JobId;
-import com.spotify.helios.common.descriptors.JobStatus;
 import com.spotify.helios.common.descriptors.TaskStatus;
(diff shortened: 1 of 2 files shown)


davidxia:
@negz @mattnworb

mattnworb:
In the case that the job has been deleted from the master, but the agent is in the process of undeploying it, does it really make sense to undeploy it again? Maybe it is simpler to just have HSD wait a little bit longer?

I think maybe we might need to customize the cleanup logic that HSD goes through, where somehow shutting down the container causes everything that helios-solo-container has ever deployed to be removed/undeployed - which means that helios-solo would have to keep track of what it has deployed somehow. 

I think trying to do cleanup on the java side, in HeliosSoloDeployment.close(), is bound to always miss stuff.

davidxia:
When you say "trying to do cleanup on the java side, in HeliosSoloDeployment.close(), is bound to always miss stuff," do you mean have logic in [start.sh](https://github.com/spotify/helios/blob/master/solo/docker/start.sh) instead?

davidxia:
@gimaker

davidxia:
@mattnworb I explored how to optionally set HSD in TemporaryJobs so that when a job is deployed, TemporaryJobs sends a reference of the docker container ID to HSD. But there are many more levels of indirection than I first thought: TemporaryJobs -> TemporaryJobsBuilder -> Deployer -> DefaultDeployer -> TemporaryJob.

I'm leaning towards just merging this. Let me know what you think.

davidxia:
Merging for now. I'm open to better solutions as we think of them.

PR #889, opened 2016-03-16
Update helios-solo IT docs to encourage RuleChain

davidxia:


davidxia:
@gimaker

docs/integration_tests_with_helios_solo.md
```diff
@@ -9,13 +9,14 @@ running the integration test have Docker installed.
 # How to use it
 
 If you have a test that is already using [TemporaryJobs][], the test can run
-within a self-contained helios-solo instance with a few small modifications.
+through a self-contained helios-solo Docker container with a few small modifications.
+Your test will create and run a Docker container that has a running Helios cluster
+inside tear it down when the tests finish.
```
gimaker: "... inside tear it down" missing an and?

docs/integration_tests_with_helios_solo.md
```diff
@@ -28,14 +29,26 @@ Then modify the `TemporaryJobs` declaration to use the client that points to
...
+are created in that order and torn down in the reverse order. Otherwise, you might
+have leftover containers from your tests.
+
+
+```java
+@ClassRule
+public static final RuleChain chain = RuleChain
+    .outerRule(soloResource)
+    .around(temporaryJobs);
+```
```
gimaker: Seems reasonable but can't really comment much on this ^

mattnworb:
Not directly related to this change but it worries me that the API for using this new flavor of Helios-solo keeps changing. It makes it harder for users to keep up with updates or makes a larger barrier to change. 

If we have time it might make sense to revisit this so that all of this setup can be done inside TemporaryJobs or some other Helios-supplied component instead of user code.

davidxia:
Technically this PR was just pointing out an existing issue. But yes, we should minimize user effort.

PR #891, opened 2016-03-17
fix small errors in healthcheck section

mattnworb:
When using `--http-check` on command line, the argument for path needs to start with a `/`. Also update the path in the example to match the JSON example, and clarify that only one of the two (CLI args or json) needs to be specified.


mattnworb:
found by @javoire

PR #892, opened 2016-03-28
Support adding/dropping Linux capabilities

davidxia:
Also allow agent to be started with a capability whitelist.
If this is specified, only these capabilities will be allowed
to be added.

### DIFF
--- a/docs/user_manual.md
+++ b/docs/user_manual.md
@@ -133,8 +133,8 @@ example that uses all the available configuration keys with an explanation of ea
 
 ```json
 {
- "addedCapabilities" : [ "IPC_LOCK", "SYSLOG" ],
- "droppedCapabilities" : [ "SYS_BOOT", "KILL" ],
+ "addCapabilities" : [ "IPC_LOCK", "SYSLOG" ],
+ "dropCapabilities" : [ "SYS_BOOT", "KILL" ],
   "command" : [ "server", "serverconfig.yaml" ],
   "env" : {
     "JVM_ARGS" : "-Ddw.feature.randomFeatureFlagEnabled=true"
@@ -187,10 +187,10 @@ Note that the recommended best practice is to save all your job creation
 parameters in version-controlled files in your project directory. This allows
 you to tie your Helios job params to changes in your application code.
 
-#### addedCapabilities
+#### addCapabilities
 The Linux capabilities to add to the container. Optional. See [Docker docs][1].
 
-#### droppedCapabilities
+#### dropCapabilities
 The Linux capabilities to remove from the container. Optional. See [Docker docs][1].
 
 #### command
--- a/docs/user_manual.md
+++ b/docs/user_manual.md
@@ -133,6 +133,8 @@ example that uses all the available configuration keys with an explanation of ea
 
 ```json
 {
+ "addedCapabilities" : [ "IPC_LOCK", "SYSLOG" ],
+ "droppedCapabilities" : [ "SYS_BOOT", "KILL" ],
   "command" : [ "server", "serverconfig.yaml" ],
   "env" : {
     "JVM_ARGS" : "-Ddw.feature.randomFeatureFlagEnabled=true"
@@ -184,6 +186,12 @@ All fields are optional except for `id` and `image`.
 Note that the recommended best practice is to save all your job creation
 parameters in version-controlled files in your project directory. This allows
 you to tie your Helios job params to changes in your application code.
+
+#### addedCapabilities
+The Linux capabilities to add to the container. Optional. See [Docker docs][1].
+
+#### droppedCapabilities
+The Linux capabilities to remove from the container. Optional. See [Docker docs][1].
 
 #### command
 The command(s) to pass to the container. Optional.
@@ -465,3 +473,5 @@ in your job creation commmand, once deployed on a host named
 `foo.example.com`, from inside the container you would see
 `HELIOS_PORT_foo` set to something like `foo.example.com:23238`
 depending on what port was allocated when it was deployed.
+
+  [1]: https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities
--- a/helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
@@ -41,6 +41,7 @@ import java.util.regex.Pattern;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
 import static java.lang.String.format;
+import static java.util.Collections.emptySet;
 import static java.util.regex.Pattern.compile;
 
 public class JobValidator {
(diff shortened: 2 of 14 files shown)


helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
```diff
@@ -212,6 +223,8 @@ public Job(@JsonProperty("id") final JobId id,
     this.securityOpt = Optional.fromNullable(securityOpt).or(EMPTY_SECURITY_OPT);
     this.networkMode = Optional.fromNullable(networkMode).orNull();
     this.metadata = Optional.fromNullable(metadata).or(EMPTY_METADATA);
+    this.capAdds = Optional.fromNullable(capAdds).or(EMPTY_CAPS);
+    this.capDrops = Optional.fromNullable(capDrops).or(EMPTY_CAPS);
```
mattnworb: pretty minor, and this pattern exists already, but even the guava Optional docs recommend using something like `Objects.firstNonNull(a, b)` for this use case: https://github.com/google/guava/wiki/UsingAndAvoidingNullExplained#convenience-methods

mattnworb:
Looks good, but a few thoughts:

- perhaps the "cap" whitelist should be stored on the helios master, so that it can report an error if someone tries to create a job with capabilities that are not allowed? Looks like in this patch, any unallowed capabilities are simply dropped.
- representing the "caps" as a Set instead of List makes checks like "is every requested cap in the whitelist" a simpler operation
- is "cap" a well enough known name for this sort of thing? I'd go with calling it "capability" inside the json unless it's really standard

mattnworb:
also I feel like it is safer for the default to be an empty whitelist, i.e. nothing is allowed

helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
davidxia: Thanks. I haven't read much of the Guava docs. But are they saying `firstNonNull()` is better than `or()`? It reads like both a fine.

> Whenever you want a null value to be replaced with some default value instead, use Objects.firstNonNull(T, T). As the method name suggests, if both of the inputs are null, it fails fast with a NullPointerException. If you are using an Optional, there are better alternatives -- e.g. first.or(second).

davidxia:
1 and 2. good ideas
3. `capAdds` was nice and terse compared to `addedCapabilities`, but I'll go with a bit longer but clearer.

helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
mattnworb: I interpret it more as don't bother creating an Optional just to use `or(..)` when `firstNonNull(a, b)` is clearer

davidxia:
yay, build is green

docs/user_manual.md
```diff
@@ -133,6 +133,8 @@ example that uses all the available configuration keys with an explanation of ea
 
 ```json
 {
+ "addedCapabilities" : [ "IPC_LOCK", "SYSLOG" ],
+ "droppedCapabilities" : [ "SYS_BOOT", "KILL" ],
```
mattnworb: forgive the nitpick, but how about `"addCapabilities" / "dropCapabilities"` in the job json so it's not using past tense?

helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
```diff
@@ -491,6 +500,37 @@ private boolean validateRepositoryName(final String repositoryName,
...
+    }
+
+    final Set<String> errors = Sets.newHashSet();
+
+    final Set<String> disallowedCaps = Sets.newHashSet();
+    for (final String cap : caps) {
+      if (!whitelistedCapabilities.contains(cap)) {
+        disallowedCaps.add(cap);
+      }
+    }
```
mattnworb: I think this can also be `final Set<String> disallowedCaps = Sets.difference(caps, whitelistedCapabilities)`

helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
```diff
@@ -491,6 +500,37 @@ private boolean validateRepositoryName(final String repositoryName,
...
+
+    if (!disallowedCaps.isEmpty()) {
+      errors.add(String.format(
+          "The following Linux capabilities haven't been whitelisted by the Helios master: '%s'. "
```
mattnworb: how about `The following Linux capabilities are not allowed: `

helios-services/src/main/java/com/spotify/helios/master/MasterParser.java
```diff
@@ -117,6 +125,12 @@ protected void addArgs(final ArgumentParser parser) {
         .help("In days. Jobs not deployed anywhere and with a job history showing they were last " +
               "used before the specified retention time will be removed. " +
               "This is disabled by default by setting it to a sentinel value of -1.");
+
+    whitelistedCapabilities = parser.addArgument("--whitelisted-capability")
+        .action(append())
+        .setDefault(new ArrayList<>())
+        .help("The Linux capabilities containers on this agent are allowed to add. "
```
mattnworb: use of "this agent" this might be copy and pasted from previous commit?

docs/user_manual.md
davidxia: sure :)

On Wed, Mar 30, 2016 at 4:30 PM, Matt Brown <notifications@github.com>
wrote:

> In docs/user_manual.md
> <https://github.com/spotify/helios/pull/892#discussion_r57958285>:
>
> > @@ -133,6 +133,8 @@ example that uses all the available configuration keys with an explanation of ea
> >
> >  ```json
> >  {
> > + "addedCapabilities" : [ "IPC_LOCK", "SYSLOG" ],
> > + "droppedCapabilities" : [ "SYS_BOOT", "KILL" ],
>
> forgive the nitpick, but how about "addCapabilities" / "dropCapabilities"
> in the job json so it's not using past tense?
>
> —
> <https://github.com/spotify/helios/pull/892/files/e019aa14389f6e4669b79ba0a921969a4aaa993f#r57958285>
>



-- 
David Xia
Software Engineer
dxia@spotify.com

helios-client/src/test/java/com/spotify/helios/common/JobValidatorTest.java
```diff
@@ -416,4 +417,31 @@ public void testExpiry() {
...
+  @Test
+  public void testWhitelistedCapabilities_noWhitelist() throws Exception {
+    final Job job = Job.newBuilder()
+        .setName("foo")
+        .setVersion("1")
+        .setImage("foobar")
+        .setAddedCapabilities(ImmutableSet.of("cap1", "cap2"))
+        .build();
+
+    assertEquals(1, validator.validate(job).size());
```
mattnworb: to be extra careful with the set math, might want to test which capability was deemed to be not allowed

mattnworb:
a few minor things above but :+1: if they are all addressed

PR #894, opened 2016-03-29
avoid use of MoreObjects.toStringHelper

mattnworb:
Users of `helios-testing` that are stuck on previous versions of Guava
(for example, if building something that needs to use Hadoop, and using
TemporaryJobs within the test of that thing) fail to load `MoreObjects`
from the classpath as this was only introduced in Guava v18.

Since the functionality used here to implement `toString()` is so tiny,
it feels unnecessary to rely on an external library for it (which
creates the potential for conflicting dependency library problems).
Therefore just implement `toString()` with simple String concatenation
(which the compiler will replaced with StringBuilder for us).

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -299,6 +299,11 @@ page at http://checkstyle.sourceforge.net/config.html -->
       <property name="severity" value="warning"/>
     </module>
 
+    <module name="RegexpSinglelineJava">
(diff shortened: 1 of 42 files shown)


mattnworb:
This fixes NoClassDefFoundErrors that @marcusb gets when building a project that uses Hadoop *and* helios-testing.

davidxia:
:+1:

PR #896, opened 2016-03-31
Make `helios create` support add/drop capabiltiies

davidxia:
Disable capablity validation for JobValidator when it's
being run by the CLI.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
@@ -32,6 +32,7 @@ import com.spotify.helios.common.descriptors.ServicePorts;
 import com.spotify.helios.common.descriptors.TcpHealthCheck;
 
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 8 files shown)


davidxia:
@gimaker

davidxia:
@mattnworb

davidxia:
I forgot to support for CLI flags. The previous commit supported reading from a job file.

helios-tools/src/main/java/com/spotify/helios/cli/command/JobCreateCommand.java
```diff
@@ -532,6 +550,17 @@ int run(final Namespace options, final HeliosClient client, final PrintStream ou
       builder.setToken(token);
     }
 
+    final Set<String> addCapabilities = ImmutableSet.copyOf(
+        options.<String>getList(addCapabilityArg.getDest()));
+    if (!addCapabilities.isEmpty()) {
```
mattnworb: checking that these aren't empty before calling the setter seems like overkill

helios-tools/src/main/java/com/spotify/helios/cli/command/JobCreateCommand.java
```diff
@@ -532,6 +550,17 @@ int run(final Namespace options, final HeliosClient client, final PrintStream ou
...
+      builder.setAddCapabilities(addCapabilities);
```
mattnworb: if the setAddCapabilities method took a Collection instead of a Set, you could just pass the list, and rely on `ImmutableSet.copyOf(collection)` in the method to do the type conversion

mattnworb:
I don't follow why you have to disable the JobValidator behavior when testing the CLI, if the validation is on the server, isn't that covered by using a mock HeliosClient?

mattnworb:
and a Builder for a class with three params seems very verbose too

helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
```diff
@@ -529,4 +523,34 @@ private boolean validateRepositoryName(final String repositoryName,
...
   }
+
+  public static Builder newBuilder() {
+    return new Builder();
+  }
+
+  public static class Builder {
+
+    private boolean shouldValidateJobHash = true;
+    private boolean shouldValidateAddCapabilities = false;
```
mattnworb: if you really want to make this optional (and I don't think it needs to be), it should at least default to true, so that it isn't accidentally left in the off position

davidxia: I'll get rid of the builder :)

On Thu, Mar 31, 2016 at 4:06 PM, Matt Brown <notifications@github.com>
wrote:

> In helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
> <https://github.com/spotify/helios/pull/896#discussion_r58118888>:
>
> > @@ -529,4 +523,34 @@ private boolean validateRepositoryName(final String repositoryName,
> >    private boolean legalPort(final int port) {
> >      return port >= 0 && port <= 65535;
> >    }
> > +
> > +  public static Builder newBuilder() {
> > +    return new Builder();
> > +  }
> > +
> > +  public static class Builder {
> > +
> > +    private boolean shouldValidateJobHash = true;
> > +    private boolean shouldValidateAddCapabilities = false;
>
> if you really want to make this optional (and I don't think it needs to
> be), it should at least default to true, so that it isn't accidentally left
> in the off position
>
> —
> <https://github.com/spotify/helios/pull/896/files/6193a4dc713e6138693cb27dd2151fb0a105acaf#r58118888>
>



-- 
David Xia
Software Engineer
dxia@spotify.com

davidxia:
This should be good to go now.

mattnworb:
:+1:

PR #897, opened 2016-04-01
use docker-client 3.6.7

mattnworb:
in order to pull in https://github.com/spotify/docker-client/pull/408

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -98,6 +98,14 @@
       <id>bintray-readytalk</id>
       <url>http://dl.bintray.com/readytalk/maven</url>
     </repository>
+    <repository>
+      <id>ossrh</id>
+      <name>Sonatype OSS</name>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -98,6 +98,14 @@
       <id>bintray-readytalk</id>
       <url>http://dl.bintray.com/readytalk/maven</url>
     </repository>
+    <repository>
+      <id>ossrh</id>
+      <name>Sonatype OSS</name>
+      <url>https://oss.sonatype.org/content/repositories/releases/</url>
+      <snapshots>
+        <enabled>false</enabled>
```
davidxia: Does this prevent us from uploading snapshots?

davidxia:
:+1:

pom.xml
mattnworb: no, this section of the pom is only for telling maven where to find artifacts. `<dependencyManagement>` covers where to upload them.

PR #898, opened 2016-04-04
Make HeliosDeploymentResource before/after public

davidxia:
Allow users to directly call these methods if they cannot
use @ClassRule like TestNG


mattnworb:
:+1:

PR #899, opened 2016-04-06
[solo] Fix skydns - two wrongs make a right

davidxia:


davidxia:
@gimaker @rohansingh @mattnworb @negz

gimaker:
:+1:

davidxia:
We should fix the state of skydns. Right now skydns is rebuilt from master here https://github.com/skynetservices/skydns whenever someone manually builds helios solo base image. We have no way of knowing what kind of skydns we've built later.

mattnworb:
Nice explanation in the commit message 👍🏻

Is it meant to commit the -verbose flag to skydns in here? I wonder if the output would become large if solo was running for a while and sending lots of DNS requests (I'm not sure how docker logs stores things). 

For the unpredictable nature of building SkyDNS, is it available as a package to install with apt-get?

gimaker:
> We admit this is super funky, but there's no way to force skydns to
> speak only TCP right now.

And it shouldn't. What it should do is properly forward the response with the truncation flag set to the requester (who would then retry using TCP). Instead it just blows up and never responds to the requester at all :(

davidxia:
This doesn't work on my mac now. See #900 for an alternative approach

PR #900, opened 2016-04-06
[solo] Use unbound to proxy to skydns; fixes shit 

davidxia:
[solo] Use unbound to proxy to skydns; fixes shit

TL;DR When two DNS servers don't work, add one more!

When running some integration tests with HeliosSoloDeployment on Docker
hosts that use a local unbound instance as its DNS resolver (i.e.
specified in `/etc/resolv.conf` on the Docker host),
we saw tests failures due to failed SRV queries to skydns. Skydns is
running in the solo container and forwards DNS queries it doesn't know
about to nameservers specified in `/etc/resolv.conf` via logic in `start.sh`.

The skydns error output from the helios solo container spawned by
HeliosSoloDeployment looked like:

```
skydns: failure to forward request "dns: failed to unpack truncated
message"
```

Our guess is that large UDP responses from the upstream unbound
have the "Message Truncated" DNS flag set. When this type of response
reaches skydns, skydns blows up and doesn't tell the client about the
error. The client times out without retrying in TCP mode. The client
would've retried if it had received an error message from skydns.

Running `dig` against skydns works. We think this is because `dig` adds
an OPT record to its query that sets "udp payload size: 4096".

Here are outstanding issues in skydns that seem related:

* skynetservices/skydns#242
* skynetservices/skydns#45

Solution:

We start an unbound instance in the solo container and have it forward
DNS queries via UDP to the upstream skydns in the same container.
Unbound will add the OPT section that makes everything work.
Things are fixed. :)

We admit this is super funky...And this only might work for UDP packets
up to 4096 bytes, the default set by unbound in OPT.

Much thanks to @gimaker for helping and suggesting unbound inside the
container.


davidxia:
@negz @mattnworb @rohansingh @gimaker

solo/docker/start.sh
```diff
@@ -23,7 +23,8 @@ done
 curl -XPUT http://127.0.0.1:4001/v2/keys/skydns/${SKYDNS_PATH} \
     -d value="{\"host\":\"$HOST_ADDRESS\"}"
 
-skydns $SKYDNS_OPTS &
+skydns $SKYDNS_OPTS -verbose &
```
mattnworb: I'm not sure if `-verbose` should be put in the image. Seems like you could run the docker image with `-e SKYDNS_OPTS=-verbose` at runtime if you really wanted verbose logging.

davidxia: Yea ill remove.

On Thursday, April 7, 2016, Matt Brown <notifications@github.com> wrote:

> In solo/docker/start.sh
> <https://github.com/spotify/helios/pull/900#discussion_r58858049>:
>
> > @@ -23,7 +23,8 @@ done
> >  curl -XPUT http://127.0.0.1:4001/v2/keys/skydns/${SKYDNS_PATH} \
> >      -d value="{\"host\":\"$HOST_ADDRESS\"}"
> >
> > -skydns $SKYDNS_OPTS &
> > +skydns $SKYDNS_OPTS -verbose &
>
> I'm not sure if -verbose should be put in the image. Seems like you could
> run the docker image with -e SKYDNS_OPTS=-verbose at runtime if you
> really wanted verbose logging.
>
> —
> <https://github.com/spotify/helios/pull/900/files/0d1483b079b076c9cfbfb8e39955dafb9cde6ec8#r58858049>
>


-- 
David Xia
Software Engineer
dxia@spotify.com

mattnworb:
:+1: overall

davidxia:
@gimaker If it looks good to you, I'll merge and try to get a release out.

gimaker:
@davidxia what happened to the nice commit message? :(

gimaker:
Oh nevermind. I see that it's still there.

gimaker:
:+1:

PR #902, opened 2016-04-15
support dockerfile-maven-plugin in TemporaryJobsBuilder.imageFromBuild()

mattnworb:
Currently, the `imageFromBuild()` method in `TemporaryJobsBuilder()`
supports reading the `image_info.json` produced by docker-maven-plugin
to find the name of the image to use in the TemporaryJob. This file is
read from either the classpath or the filesystem if the former fails.

This change adds support for reading the `target/docker/image-name` file
produced by dockerfile-maven-plugin as well.

### DIFF
--- a/helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobBuilder.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobBuilder.java
@@ -17,12 +17,6 @@
 
 package com.spotify.helios.testing;
 
-import com.google.common.collect.Lists;
-import com.google.common.collect.Sets;
-import com.google.common.io.Files;
-import com.google.common.io.Resources;
-
-import com.fasterxml.jackson.databind.JsonNode;
 import com.spotify.helios.common.Json;
 import com.spotify.helios.common.descriptors.HealthCheck;
 import com.spotify.helios.common.descriptors.HttpHealthCheck;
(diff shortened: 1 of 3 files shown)


davidxia:
cc @gilles

mattnworb:
The tests for this are pretty hacky. A much cleaner approach would be to add some sort of "FileReader" interface to TemporaryJobBuilder that hides the logic of reading from the classpath or from the `target/` folder so I could swap that out in the test with a temporary file, but that felt like a more intrusive change.

mattnworb:
@davidxia @negz @gimaker @rohansingh @dflemstr

helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobBuilder.java
```diff
@@ -298,31 +300,73 @@ public TemporaryJob deploy(final List<String> hosts) {
...
+   * Set the {@link #image(String)} field for the Docker image to use in this job from the output of
+   * docker-maven-plugin (image_info.json on classpath) or dockerfile-maven-plugin
+   * (docker/image-name on classpath)
+   */
   public TemporaryJobBuilder imageFromBuild() {
     final String envPath = env.get("IMAGE_INFO_PATH");
     if (envPath != null) {
       return imageFromInfoFile(envPath);
-    } else {
-      final String name = fromNullable(env.get("IMAGE_INFO_NAME")).or("image_info.json");
```
mattnworb: removed this `or()` as the first part of the condition is always false in this branch

gilles:
nice.

For the test since you're reading via a `Resource`, you should be able to just put test files in `test/resources/target/...` and `test/resources/docker/`

mattnworb:
Found a test which was relying on `imageFromBuild()` produced by some other module in the project; the test doesn't actually care what image is used - seems like it worked on accident before. Fixed in 7518d82

davidxia:
👍

PR #903, opened 2016-04-19
Introduce a 'rolling operation' concept.

negz:
The target job ID and rollout options of a rolling update are associated with the rolling operation - not the deployment group itself. Each rolling operation is associated with a deployment group. The 'reason' for a rolling update (i.e. manual or 'hosts changed') is a property of the rolling operation.

Deployment groups may now have state 'stable' or 'unstable'. Stable deployment groups are known to be running the same job across all agents. Unstable groups are known to have some containers down, or to be running different jobs across agents. Rolling operations model the states previously associated with deployment groups (ROLLING_OUT, DONE, FAILED).

Some hacks (i.e. the DeploymentGroupResponse) have been put in place in order to avoid breaking the existing REST and CLI APIs.

This change means:
* A failed rolling update will no longer require human intervention to unblock future automatic (i.e. 'hosts changed') rolling updates.
* Rolling restarts could be implemented as a variant of a rolling operation - perhaps as a rolling operation with a null jobId?
* We could expose an audit trail of a deployment group's rolling updates.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/DeploymentGroup.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/DeploymentGroup.java
@@ -23,8 +23,6 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
-import org.jetbrains.annotations.Nullable;
-
 import java.util.List;
 
 import static java.util.Collections.emptyList;
@@ -36,7 +34,6 @@ import static java.util.Collections.emptyList;
  * <pre>
  * {
  *   "name":"foo-group",
- *   "job":"foo:0.1.0",
  *   "hostSelectors":[
  *     {
  *       "label":"foo",
@@ -48,14 +45,7 @@ import static java.util.Collections.emptyList;
  *       "operator":"EQUALS"
  *       "operand":"qux",
  *     }
- *   ],
- *   "rolloutOptions":{
- *     "migrate":false,
- *     "parallelism":2,
- *     "timeout":1000,
- *     "overlap":true,
- *     "token": "insecure-access-token"
- *   }
+ *   ]
  * }
  * </pre>
  */
@@ -64,46 +54,30 @@ import static java.util.Collections.emptyList;
 public class DeploymentGroup extends Descriptor {
 
   public static final String EMPTY_NAME = "";
-  public static final JobId EMPTY_JOB_ID = null;
 
   private final String name;
   private final List<HostSelector> hostSelectors;
-  private final JobId jobId;
-  private final RolloutOptions rolloutOptions;
 
   /**
    * Create a Job.
    *
    * @param name The docker name to use.
-   * @param jobId The job ID for the deployment group.
    * @param hostSelectors The selectors that determine which agents are part of the deployment
    *                       group.
    */
   public DeploymentGroup(
       @JsonProperty("name") final String name,
-      @JsonProperty("hostSelectors") final List<HostSelector> hostSelectors,
-      @JsonProperty("job") @Nullable final JobId jobId,
-      @JsonProperty("rolloutOptions") @Nullable final RolloutOptions rolloutOptions) {
+      @JsonProperty("hostSelectors") final List<HostSelector> hostSelectors) {
     this.name = name;
     this.hostSelectors = hostSelectors;
-    this.jobId = jobId;
-    this.rolloutOptions = rolloutOptions;
   }
 
   public String getName() {
     return name;
   }
 
-  public JobId getJobId() {
-    return jobId;
-  }
-
   public List<HostSelector> getHostSelectors() {
     return hostSelectors;
-  }
-
-  public RolloutOptions getRolloutOptions() {
-    return rolloutOptions;
   }
 
   public static Builder newBuilder() {
@@ -121,18 +95,11 @@ public class DeploymentGroup extends Descriptor {
 
     final DeploymentGroup that = (DeploymentGroup) o;
 
-    if (jobId != null ? !jobId.equals(that.jobId) : that.jobId != null) {
-      return false;
-    }
     if (hostSelectors != null ? !hostSelectors.equals(that.hostSelectors)
                                : that.hostSelectors != null) {
       return false;
     }
     if (name != null ? !name.equals(that.name) : that.name != null) {
-      return false;
-    }
-    if (rolloutOptions != null ? !rolloutOptions.equals(that.rolloutOptions)
-                               : that.rolloutOptions != null) {
       return false;
     }
 
@@ -143,8 +110,6 @@ public class DeploymentGroup extends Descriptor {
   public int hashCode() {
     int result = name != null ? name.hashCode() : 0;
     result = 31 * result + (hostSelectors != null ? hostSelectors.hashCode() : 0);
-    result = 31 * result + (jobId != null ? jobId.hashCode() : 0);
-    result = 31 * result + (rolloutOptions != null ? rolloutOptions.hashCode() : 0);
     return result;
   }
 
@@ -153,8 +118,6 @@ public class DeploymentGroup extends Descriptor {
     return "DeploymentGroup{" +
            "name='" + name + '\'' +
            ", hostSelectors=" + hostSelectors +
-           ", job=" + jobId +
-           ", rolloutOptions=" + rolloutOptions +
            '}';
   }
 
@@ -162,9 +125,7 @@ public class DeploymentGroup extends Descriptor {
     final Builder builder = newBuilder();
 
     return builder.setName(name)
-        .setJobId(jobId)
-        .setHostSelectors(hostSelectors)
-        .setRolloutOptions(rolloutOptions);
+        .setHostSelectors(hostSelectors);
   }
 
   public static class Builder implements Cloneable {
@@ -178,15 +139,11 @@ public class DeploymentGroup extends Descriptor {
     private static class Parameters implements Cloneable {
 
       public String name;
-      public JobId jobId;
       public List<HostSelector> hostSelectors;
-      public RolloutOptions rolloutOptions;
 
       private Parameters() {
         this.name = EMPTY_NAME;
-        this.jobId = EMPTY_JOB_ID;
         this.hostSelectors = emptyList();
-        this.rolloutOptions = null;
       }
     }
 
@@ -199,15 +156,6 @@ public class DeploymentGroup extends Descriptor {
       return this;
     }
 
-    public JobId getJobId() {
-      return p.jobId;
-    }
-
-    public Builder setJobId(final JobId jobId) {
-      p.jobId = jobId;
-      return this;
-    }
-
     public List<HostSelector> getHostSelectors() {
       return p.hostSelectors;
     }
@@ -217,17 +165,8 @@ public class DeploymentGroup extends Descriptor {
       return this;
     }
 
-    public RolloutOptions getRolloutOptions() {
-      return p.rolloutOptions;
-    }
-
-    public Builder setRolloutOptions(final RolloutOptions rolloutOptions) {
-      p.rolloutOptions = rolloutOptions;
-      return this;
-    }
-
     public DeploymentGroup build() {
-      return new DeploymentGroup(p.name, p.hostSelectors, p.jobId, p.rolloutOptions);
+      return new DeploymentGroup(p.name, p.hostSelectors);
     }
   }
 
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/DeploymentGroup.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/DeploymentGroup.java
@@ -53,7 +53,7 @@ import static java.util.Collections.emptyList;
 @JsonInclude(JsonInclude.Include.NON_NULL)
 public class DeploymentGroup extends Descriptor {
 
-  public static final String EMPTY_NAME = "";
+  private static final String EMPTY_NAME = "";
 
   private final String name;
   private final List<HostSelector> hostSelectors;
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperation.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperation.java
@@ -1,0 +1,247 @@
+/*
+ * Copyright (c) 2015 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.helios.common.descriptors;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import org.jetbrains.annotations.Nullable;
+
+import java.util.UUID;
+
+/**
+ * Represents a rolling operation on a Helios deployment group.
+ *
+ * An sample expression of it in JSON might be:
+ * <pre>
+ * {
+ *   "id":"some-cool-uuid",
+ *   "deploymentGroupName":"my-awesome-group",
+ *   "job":"foo:0.1.0",
+ *   "reason":"HOSTS_CHANGED",
+ *   "rolloutOptions":{
+ *     "migrate":false,
+ *     "parallelism":2,
+ *     "timeout":1000,
+ *     "overlap":true,
+ *     "token": "insecure-access-token"
+ *   }
+ * }
+ * </pre>
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class RollingOperation extends Descriptor {
+
+  public enum Reason {
+    MANUAL,
+    HOSTS_CHANGED,
+  }
+
+  private static final String EMPTY_DEPLOYMENT_GROUP_NAME = "";
+  private static final JobId EMPTY_JOB_ID = null;
+
+  private final String id;
+  private final String deploymentGroupName;
+  private final JobId jobId;
+  private final Reason reason;
+  private final RolloutOptions rolloutOptions;
+
+  /**
+   * Create a rolling operation.
+   *
+   * @param id The docker id to use.
+   * @param jobId The job ID for the deployment group.
+   */
+  public RollingOperation(
+      @JsonProperty("id") final String id,
+      @JsonProperty("deploymentGroupName") final String deploymentGroupName,
+      @JsonProperty("reason") final Reason reason,
+      @JsonProperty("job") @Nullable final JobId jobId,
+      @JsonProperty("rolloutOptions") @Nullable final RolloutOptions rolloutOptions) {
+    this.id = id;
+    this.deploymentGroupName = deploymentGroupName;
+    this.jobId = jobId;
+    this.reason = reason;
+    this.rolloutOptions = rolloutOptions;
+  }
+
+  public String getId() {
+    return id;
+  }
+
+  public String getDeploymentGroupName() {
+    return deploymentGroupName;
+  }
+
+  public JobId getJobId() {
+    return jobId;
+  }
+
+  public Reason getReason() {
+    return reason;
+  }
+
+  public RolloutOptions getRolloutOptions() {
+    return rolloutOptions;
+  }
+
+  public static Builder newBuilder() {
+    return new Builder();
+  }
+
+  @Override
+  public boolean equals(final Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    final RollingOperation that = (RollingOperation) o;
+
+    if (jobId != null ? !jobId.equals(that.jobId) : that.jobId != null) {
+      return false;
+    }
+    if (deploymentGroupName != null ? !deploymentGroupName.equals(that.deploymentGroupName)
+                                    : that.deploymentGroupName != null) {
+      return false;
+    }
+    if (id != null ? !id.equals(that.id) : that.id != null) {
+      return false;
+    }
+    if (reason != null ? !reason.equals(that.reason) : that.reason != null) {
+      return false;
+    }
+    if (rolloutOptions != null ? !rolloutOptions.equals(that.rolloutOptions)
+                               : that.rolloutOptions != null) {
+      return false;
+    }
+
+    return true;
+  }
+
+  @Override
+  public int hashCode() {
+    int result = id != null ? id.hashCode() : 0;
+    result = 31 * result + (deploymentGroupName != null ? deploymentGroupName.hashCode() : 0);
+    result = 31 * result + (jobId != null ? jobId.hashCode() : 0);
+    result = 31 * result + (reason != null ? reason.hashCode() : 0);
+    result = 31 * result + (rolloutOptions != null ? rolloutOptions.hashCode() : 0);
+    return result;
+  }
+
+  @Override
+  public String toString() {
+    return "RollingOperation{" +
+           "id='" + id + '\'' +
+           ", deploymentGroupName=" + deploymentGroupName +
+           ", job=" + jobId +
+           ", reason=" + reason +
+           ", rolloutOptions=" + rolloutOptions +
+           '}';
+  }
+
+  public Builder toBuilder() {
+    final Builder builder = newBuilder();
+
+    return builder
+        .setJobId(jobId)
+        .setDeploymentGroupName(deploymentGroupName)
+        .setReason(reason)
+        .setRolloutOptions(rolloutOptions);
+  }
+
+  public static class Builder implements Cloneable {
+
+    private final Parameters p;
+
+    private Builder() {
+      this.p = new Parameters();
+    }
+
+    private static class Parameters implements Cloneable {
+
+      public String id;
+      public String deploymentGroupName;
+      public JobId jobId;
+      public Reason reason;
+      public RolloutOptions rolloutOptions;
+
+      private Parameters() {
+        this.id = UUID.randomUUID().toString();
+        this.deploymentGroupName = EMPTY_DEPLOYMENT_GROUP_NAME;
+        this.jobId = EMPTY_JOB_ID;
+        this.reason = Reason.MANUAL;
+        this.rolloutOptions = null;
+      }
+    }
+
+    public String getId() {
+      return p.id;
+    }
+
+    public Builder setId(final String id) {
+      p.id = id;
+      return this;
+    }
+
+    public String getDeploymentGroupName() {
+      return p.deploymentGroupName;
+    }
+
+    public Builder setDeploymentGroupName(final String deploymentGroupName) {
+      p.deploymentGroupName = deploymentGroupName;
+      return this;
+    }
+
+    public JobId getJobId() {
+      return p.jobId;
+    }
+
+    public Builder setJobId(final JobId jobId) {
+      p.jobId = jobId;
+      return this;
+    }
+
+    public Reason getReason() {
+      return p.reason;
+    }
+
+    public Builder setReason(final Reason reason) {
+      p.reason = reason;
+      return this;
+    }
+
+    public RolloutOptions getRolloutOptions() {
+      return p.rolloutOptions;
+    }
+
+    public Builder setRolloutOptions(final RolloutOptions rolloutOptions) {
+      p.rolloutOptions = rolloutOptions;
+      return this;
+    }
+
+    public RollingOperation build() {
+      return new RollingOperation(p.id, p.deploymentGroupName, p.reason, p.jobId, p.rolloutOptions);
+    }
+  }
+
+}
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperation.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperation.java
@@ -169,78 +169,62 @@ public class RollingOperation extends Descriptor {
         .setRolloutOptions(rolloutOptions);
   }
 
-  public static class Builder implements Cloneable {
+  public static class Builder {
 
-    private final Parameters p;
-
-    private Builder() {
-      this.p = new Parameters();
-    }
-
-    private static class Parameters implements Cloneable {
-
-      public String id;
-      public String deploymentGroupName;
-      public JobId jobId;
-      public Reason reason;
-      public RolloutOptions rolloutOptions;
-
-      private Parameters() {
-        this.id = UUID.randomUUID().toString();
-        this.deploymentGroupName = EMPTY_DEPLOYMENT_GROUP_NAME;
-        this.jobId = EMPTY_JOB_ID;
-        this.reason = Reason.MANUAL;
-        this.rolloutOptions = null;
-      }
-    }
+    private String id = UUID.randomUUID().toString();
+    private String deploymentGroupName = EMPTY_DEPLOYMENT_GROUP_NAME;
+    private JobId jobId = EMPTY_JOB_ID;
+    private Reason reason = Reason.MANUAL;
+    private RolloutOptions rolloutOptions;
 
     public String getId() {
-      return p.id;
+      return this.id;
     }
 
     public Builder setId(final String id) {
-      p.id = id;
+      this.id = id;
       return this;
     }
 
     public String getDeploymentGroupName() {
-      return p.deploymentGroupName;
+      return this.deploymentGroupName;
     }
 
     public Builder setDeploymentGroupName(final String deploymentGroupName) {
-      p.deploymentGroupName = deploymentGroupName;
+      this.deploymentGroupName = deploymentGroupName;
       return this;
     }
 
     public JobId getJobId() {
-      return p.jobId;
+      return this.jobId;
     }
 
     public Builder setJobId(final JobId jobId) {
-      p.jobId = jobId;
+      this.jobId = jobId;
       return this;
     }
 
     public Reason getReason() {
-      return p.reason;
+      return this.reason;
     }
 
     public Builder setReason(final Reason reason) {
-      p.reason = reason;
+      this.reason = reason;
       return this;
     }
 
     public RolloutOptions getRolloutOptions() {
-      return p.rolloutOptions;
+      return this.rolloutOptions;
     }
 
     public Builder setRolloutOptions(final RolloutOptions rolloutOptions) {
-      p.rolloutOptions = rolloutOptions;
+      this.rolloutOptions = rolloutOptions;
       return this;
     }
 
     public RollingOperation build() {
-      return new RollingOperation(p.id, p.deploymentGroupName, p.reason, p.jobId, p.rolloutOptions);
+      return new RollingOperation(
+          this.id, this.deploymentGroupName, this.reason, this.jobId, this.rolloutOptions);
     }
   }
 
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperationStatus.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperationStatus.java
@@ -23,20 +23,30 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
- * The state of a deployment group.
+ * The state of a rolling operation.
+ *
+ * An sample expression of it in JSON might be:
+ * <pre>
+ * {
+ *   "state":"FAILED",
+ *   "error":"Stopped by user"
+ * }
+ * </pre>
  */
 @JsonIgnoreProperties(ignoreUnknown = true)
-public class DeploymentGroupStatus extends Descriptor {
+public class RollingOperationStatus extends Descriptor {
 
   public enum State {
-    STABLE,
-    UNSTABLE,
+    NEW,
+    ROLLING_OUT,
+    FAILED,
+    DONE,
   }
 
   private final State state;
   private final String error;
 
-  private DeploymentGroupStatus(
+  private RollingOperationStatus(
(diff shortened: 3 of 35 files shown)


helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperation.java
```diff
@@ -0,0 +1,247 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import org.jetbrains.annotations.Nullable;
+
+import java.util.UUID;
+
+/**
+ * Represents a Helios deployment group.
```
negz: :s/deployment group/rolling operation/

mattnworb:
At a high level this makes a lot of sense, need to dig further into the diff though.

Would there be a need to migrate data already in zookeeper if/when deploying this change to a new format? 

I think the part of this where you ensured that the REST API is unchanged is nice, I wouldn't call that a hack at all.

The idea of having a history of rolling operations sounds really nice, but storing the "history" of anything in zookeeper generally becomes problematic for a number of reasons: the data set size is limited to what will fit in RAM, the znode data model (something resembling a filesystem) is not a great fit for most queries, etc. I think we would be better off storing long-term history in another, non-zookeeper, system (like a traditional RDBMS)

negz:
@mattnworb All good points.

I hadn't considered the deployment issue. I think it's likely that a migration would be needed and both versions couldn't coexist at the moment. I'd have to double check to be sure, but I think chances of everything just working are pretty slim.

Space wise, could we perhaps trim the history to some reasonable number? Currently the additional overhead is:
* Two new JSON blobs created per rolling update (one for jobId and rollout options, another for status)
* A list of rolling updates associated with each deployment group

We could possibly trim this down to the 10-20 latest rolling updates per DG?

davidxia:
A migration of ZK data will be tricky. Will older `helios` CLIs be able to handle the new types of data? We've had issues with data/schema changes causing old CLIs to break.

davidxia:
I saw the tests failed because `com.spotify.helios.system.DeploymentGroupTest` hung for more than 10 minutes. I retried and the same thing happened. Is there something that makes that test hang?

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperation.java
```diff
@@ -0,0 +1,247 @@
...
+ *
+ * An sample expression of it in JSON might be:
```
davidxia: This JSON doesn't match the class attributes.

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperation.java
```diff
@@ -0,0 +1,247 @@
...
+
+  /**
+   * Create a rolling operation.
+   *
+   * @param id The docker id to use.
+   * @param jobId The job ID for the deployment group.
+   */
+  public RollingOperation(
+      @JsonProperty("id") final String id,
+      @JsonProperty("deploymentGroupName") @Nullable final String deploymentGroupName,
```
davidxia: Do we need this to be annotated with `@Nullable`? Isn't this required? Same for job and reason below.

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperation.java
```diff
@@ -0,0 +1,247 @@
...
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class RollingOperation extends Descriptor {
+
+  public enum Reason {
+    MANUAL,
+    HOSTS_CHANGED,
+  }
+
+  public static final String EMPTY_ID = "";
```
davidxia: unused

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperation.java
davidxia: Or is this for backwards compatibility?

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperation.java
```diff
@@ -0,0 +1,247 @@
...
+
+  public static class Builder implements Cloneable {
+
+    private final Parameters p;
+
+    private Builder() {
+      this.p = new Parameters();
+    }
+
+    private static class Parameters implements Cloneable {
```
davidxia: Is this class necessary or can we just use only a Builder?

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperation.java
```diff
@@ -0,0 +1,247 @@
...
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class RollingOperation extends Descriptor {
+
+  public enum Reason {
+    MANUAL,
+    HOSTS_CHANGED,
+  }
+
+  public static final String EMPTY_ID = "";
+  public static final String EMPTY_DEPLOYMENT_GROUP_NAME = "";
```
davidxia: These two can be private.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -448,36 +551,58 @@ public void updateDeploymentGroupHosts(final String groupName, final List<String
...
-          }
+        final RollingOperation lastGoodOp = getLastGoodRollingOperation(client, groupName);
+        if (lastGoodOp != null)  {
+
+          // New rolling operation!
+          final RollingOperation rolling = RollingOperation.newBuilder()
+              .setDeploymentGroupName(groupName)
+              .setJobId(lastGoodOp.getJobId())
+              .setRolloutOptions(lastGoodOp.getRolloutOptions())
+              .build();
```
davidxia: set the reason here too?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -496,22 +650,42 @@ public void rollingUpdate(final DeploymentGroup deploymentGroup,
...
-    final DeploymentGroup updated = deploymentGroup.toBuilder()
+    if (getJob(jobId) == null) {
+      throw new JobDoesNotExistException(jobId);
+    }
+
+    final RollingOperation rolling = RollingOperation.newBuilder()
+        .setDeploymentGroupName(deploymentGroup.getName())
         .setJobId(jobId)
         .setRolloutOptions(options)
         .build();
```
davidxia: set the reason?

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperationStatus.java
```diff
@@ -0,0 +1,125 @@
...
+
+package com.spotify.helios.common.descriptors;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * The state of a rolling operation.
```
davidxia: We can include a JSON example just to be consistent.

negz:
@davidxia Thanks for the review! Yeah the integration tests are not passing yet so there's probably still bugs in here. The idea was to get it out for review to get feedback while I was working on finding and fixing them.

davidxia:
For the new classes, it's helpful to have javadocs explaining how they're used by other classes.

mattnworb:
> A migration of ZK data will be tricky. Will older helios CLIs be able to handle the new types of data? We've had issues with data/schema changes causing old CLIs to break.

@davidxia The CLI doesn't talk to Zookeeper, does it?

(it drives me nuts that Github doesn't have inline responses for top-level comments)

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -448,36 +551,58 @@ public void updateDeploymentGroupHosts(final String groupName, final List<String
...
 
-        if (deploymentGroup.getJobId() != null) {
-          final DeploymentGroupStatus deploymentGroupStatus = getDeploymentGroupStatus(groupName);
-          if (deploymentGroupStatus == null || deploymentGroupStatus.getState() != FAILED) {
-            final RollingUpdateOp op =
-                getInitRollingUpdateOps(deploymentGroup, hosts, HOSTS_CHANGED, client);
-            ops.addAll(op.operations());
-            events = op.events();
-          }
+        final RollingOperation lastGoodOp = getLastGoodRollingOperation(client, groupName);
```
gimaker: [Under the assumption that the last good rolling operation == the last op that succeeded]

IMO, hosts changing should cause the current target version to be deployed, not the last succeesfully deployed version.

For example, if I deploy version B but the rollout fails on some random host. Then when I add hosts I want version B to be deployed on them, not some older version A.

davidxia:
I looked at about half of it, I'll go through the other half soon.

mattnworb:
@negz do you think it would be feasible to outline somewhere (here in a comment, or in a doc added to the repo) what the new ZK schema looks like for what you are adding? I think that could help in reviewing.

davidxia:
It doesn't, but even if data and code in the masters are compatible, that
data could make old clients blow up right? An example is how the
`HeliosClient`'s `deploymentGroup()` now returns a `DeploymentGroupResponse`
instead of a `DeploymentGroup`. I think old clients will choke because they
won't know how to deserialize this. Let me know if I'm missing something.

davidxia:
It doesn't, but even if data and code in the masters are compatible, that
data could make old clients blow up right? An example is how the
`HeliosClient.deploymentGroup()` now returns a `DeploymentGroupResponse`
instead of a `DeploymentGroup`. I think old clients will choke because they
won't know how to deserialize this. Let me know if I'm missing something.

mattnworb:
as long as the JSON returned by the endpoint in the new version matches the JSON that was returned by the old version (i.e. what an old CLI client expects), then it should be fine.

davidxia:
Ah, that's right.

On Mon, Apr 25, 2016 at 2:22 PM, Matt Brown <notifications@github.com>
wrote:

> as long as the JSON returned by the endpoint in the new version matches
> the JSON that was returned by the old version (i.e. what an old CLI client
> expects), then it should be fine.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/spotify/helios/pull/903#issuecomment-214469734>
>



-- 
David Xia
Software Engineer
dxia@spotify.com

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: Done, thanks! I swore I was doing that before - must have been a regression.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperation.java
negz: I don't actually know. I was talking about this with @mattnworb the other day - we seem to have Builders with both patterns. I used this pattern because I copy pasta'd from `DeploymentGroup`.

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperation.java
negz: You're right - group name and reason should not be nullable.

I wonder about jobId though. For some reason that was nullable on `DeploymentGroup`. Maybe it shouldn't be on `RollingOperation`? Though I was thinking maybe a `RollingOperation` with a `null` jobId could perform a rolling restart....

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperation.java
negz: Ack

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperation.java
negz: Ack

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperationStatus.java
negz: Done.

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperation.java
davidxia: In that case, I'd ditch the `Parameters` and just make it a simple Builder.

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperationStatus.java
```diff
@@ -0,0 +1,133 @@
...
+ * An sample expression of it in JSON might be:
+ * <pre>
+ * {
+ *   "state":"FAILED",
+ *   "error":"Stopped by user"
+ * }
+ * </pre>
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class RollingOperationStatus extends Descriptor {
```
davidxia: Any reason to not have `@JsonInclude(JsonInclude.Include.NON_NULL)` as we do elsewhere?

helios-client/src/main/java/com/spotify/helios/common/descriptors/DeploymentGroup.java
```diff
@@ -64,48 +54,32 @@
 public class DeploymentGroup extends Descriptor {
 
   public static final String EMPTY_NAME = "";
```
davidxia: Can be private.

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperationTasks.java
```diff
@@ -26,32 +26,32 @@
 import static com.google.common.base.Preconditions.checkNotNull;
 
 @JsonIgnoreProperties(ignoreUnknown = true)
```
davidxia: Any reason to not have @JsonInclude(JsonInclude.Include.NON_NULL) as we do elsewhere?

helios-client/src/main/java/com/spotify/helios/common/protocol/DeploymentGroupResponse.java
```diff
@@ -0,0 +1,159 @@
...
+    private JobId jobId;
+    private RolloutOptions rolloutOptions;
+
+    public Builder name(final String name) {
+      this.name = name;
+      return this;
+    }
+
+    public Builder hostSelectors(final List<HostSelector> hostSelectors) {
+      this.hostSelectors = hostSelectors;
```
davidxia: We should make a defensive copy of this list, ie `ImmutableList.copyOf()`.

helios-client/src/main/java/com/spotify/helios/common/protocol/DeploymentGroupStatusResponse.java
```diff
@@ -187,11 +200,65 @@ public boolean equals(final Object o) {
...
+      return this;
+    }
+
+    public Builder error(final String error) {
+      this.error = error;
+      return this;
+    }
+
+    public Builder hostStatuses(final List<HostStatus> hostStatuses) {
+      this.hostStatuses = hostStatuses;
```
davidxia: We should make a defensive copy of this list, ie ImmutableList.copyOf().

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperationStatus.java
negz: No, I must have missed that elsewhere. What does it do?

helios-client/src/main/java/com/spotify/helios/common/descriptors/DeploymentGroup.java
negz: Ack.

helios-client/src/main/java/com/spotify/helios/common/protocol/DeploymentGroupResponse.java
negz: Good point. Will do.

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperationStatus.java
davidxia: It will only include non null values. See if it makes sense to use.

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperation.java
negz: Sound good.

(It would be really nice to switch the whole codebase over to one consistent builder pattern.)

davidxia: I'm not sure why `DeploymentGroup` needs `Parameters` either. It makes sense for `Job` to have it since it allows us to customize `Job.Builder` attributes easily.

helios-client/src/test/java/com/spotify/helios/common/descriptors/RollingOperationTest.java
```diff
@@ -0,0 +1,65 @@
...
+public class RollingOperationTest {
+
+  @Test
+  public void verifyBuilder() {
+    final RollingOperation.Builder builder = RollingOperation.newBuilder();
+
+    // Input to setXXX
+    final String setDeploymentGroupName = "foo-group";
+    //noinspection ConstantConditions
+    final List<HostSelector> setHostSelectors =
```
davidxia: Unused var

helios-client/src/test/java/com/spotify/helios/common/descriptors/RollingOperationTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+public class RollingOperationTest {
+
+  @Test
+  public void verifyBuilder() {
+    final RollingOperation.Builder builder = RollingOperation.newBuilder();
+
+    // Input to setXXX
+    final String setDeploymentGroupName = "foo-group";
+    //noinspection ConstantConditions
```
davidxia: can probably remove this line

helios-client/src/test/java/com/spotify/helios/common/descriptors/RollingOperationTest.java
```diff
@@ -0,0 +1,65 @@
...
+    final List<HostSelector> setHostSelectors =
+        ImmutableList.of(HostSelector.parse("foo=bar"), HostSelector.parse("baz=qux"));
+    final JobId setJobId = JobId.fromString("foo:0.1.0");
+    final RolloutOptions setRolloutOptions = RolloutOptions.newBuilder()
+        .setTimeout(1000)
+        .setParallelism(2)
+        .setMigrate(false)
+        .build();
```
davidxia: test overlap and token as well?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -141,9 +147,10 @@ public int compare(TaskStatusEvent arg0, TaskStatusEvent arg1) {
       STRING_LIST_TYPE =
       new TypeReference<List<String>>() {};
 
-  private static final String DEPLOYMENT_GROUP_EVENTS_KAFKA_TOPIC = "HeliosDeploymentGroupEvents";
-  private static final DeploymentGroupEventFactory DEPLOYMENT_GROUP_EVENT_FACTORY =
-      new DeploymentGroupEventFactory();
+  public static final int MAX_ROLLING_OPERATION_HISTORY = 10;
```
davidxia: Can be private. While you're at it, a lot of the attrs above can also be private.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -340,14 +348,20 @@ public void addDeploymentGroup(final DeploymentGroup deploymentGroup)
...
     try {
       try {
         client.ensurePath(Paths.configDeploymentGroups());
         client.ensurePath(Paths.statusDeploymentGroups());
         client.transaction(
             create(Paths.configDeploymentGroup(deploymentGroup.getName()), deploymentGroup),
-            create(Paths.statusDeploymentGroup(deploymentGroup.getName())),
+            create(Paths.statusDeploymentGroup(deploymentGroup.getName()), status),
             create(Paths.statusDeploymentGroupHosts(deploymentGroup.getName()),
+                   Json.asBytesUnchecked(emptyList())),
```
davidxia: Why didn't we need this before?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -362,7 +376,7 @@ public void addDeploymentGroup(final DeploymentGroup deploymentGroup)
   public DeploymentGroup getDeploymentGroup(final String name)
       throws DeploymentGroupDoesNotExistException {
     log.debug("getting deployment-group: {}", name);
-    final ZooKeeperClient client = provider.get("getDeploymentGroup");
+    final ZooKeeperClient client = provider.get("getDeploymentGroupResponse");
```
davidxia: Perhaps keep this as before for easier debugging, ie have it match the method name?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -486,187 +577,273 @@ public void updateDeploymentGroupHosts(final String groupName, final List<String
...
+        final RollingOperationStatus status = getRollingOperationStatus(client, rollingOpId);
+        if (status.getState() == ROLLING_OUT) {
+          throw new HeliosRuntimeException(
+              "Please wait for rolling-operation " + rollingOpId + " to finish.");
+        }
+      } catch (RollingOperationDoesNotExistException e) {
+        return;
+      }
+      client.ensurePath(Paths.configRollingOps());
+      client.ensurePath(Paths.statusRollingOps());
```
davidxia: Shouldn't we also remove `Paths.statusRollingOpsTasks`? And shouldn't we use the [same code @gimaker had here](https://github.com/spotify/helios/pull/903/files#diff-8d500c531ffb1f7836e9d5b60585a002L408) to ensure consistency?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -401,28 +416,12 @@ public void removeDeploymentGroup(final String name) throws DeploymentGroupDoesN
...
-      if (client.exists(Paths.statusDeploymentGroupTasks(name)) != null) {
-        operations.add(delete(Paths.statusDeploymentGroupTasks(name)));
-      } else {
-        operations.add(create(Paths.statusDeploymentGroupTasks(name)));
-        operations.add(delete(Paths.statusDeploymentGroupTasks(name)));
-      }
-
       operations.add(delete(Paths.configDeploymentGroup(name)));
       operations.add(delete(Paths.statusDeploymentGroupHosts(name)));
+      operations.add(delete(Paths.statusDeploymentGroupRollingOps(name)));
```
davidxia: Is it intentional that you don't remove `/status/rolling-operations/<uuid>` and `/status/rolling-operations-tasks/<uuid>`?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -433,6 +432,92 @@ public void removeDeploymentGroup(final String name) throws DeploymentGroupDoesN
     }
   }
 
+  public List<RollingOperation> getRollingOperations(final String groupName)
+    throws DeploymentGroupDoesNotExistException {
+    final ZooKeeperClient client = provider.get("getRollingOperations");
+    final List<RollingOperation> ops = new ArrayList<RollingOperation>();
```
davidxia: `ArrayList<RollingOperation>` can be `ArrayList<>` since java7+ has type inference for generic instance creation.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -433,6 +432,92 @@ public void removeDeploymentGroup(final String name) throws DeploymentGroupDoesN
     }
   }
 
+  public List<RollingOperation> getRollingOperations(final String groupName)
```
davidxia: can be package local

davidxia: Is this just `@VisibleForTesting`?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -433,6 +432,92 @@ public void removeDeploymentGroup(final String name) throws DeploymentGroupDoesN
...
+    try {
+      final Node node = client.getNode(Paths.statusDeploymentGroupRollingOps(groupName));
+      return Json.read(node.getBytes(), new TypeReference<List<String>>() {});
+    } catch (JsonMappingException e) {
+      // Can't parse JSON. May as well act as if there's been no rolling operations.
+      return emptyList();
+    } catch (NoNodeException e) {
+      throw new DeploymentGroupDoesNotExistException(groupName, e);
+    } catch (KeeperException | IOException e) {
+      throw new HeliosRuntimeException("getting deployment group rolling-operations failed", e);
```
davidxia: Let's include the group name.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -433,6 +432,92 @@ public void removeDeploymentGroup(final String name) throws DeploymentGroupDoesN
...
+  }
+
+  private RollingOperation getLastRollingOperation(final ZooKeeperClient client,
+                                                       final String groupName)
+      throws DeploymentGroupDoesNotExistException {
+    // The most recent rolling operations should always be at the head of the list returned by
+    // getRollingOperations().
+    for (final String rollingOpId : getRollingOperationIds(client, groupName)) {
+      try {
+        final RollingOperation rolling = getRollingOperation(client, rollingOpId);
```
davidxia: nitpick: redundant var

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -433,6 +432,92 @@ public void removeDeploymentGroup(final String name) throws DeploymentGroupDoesN
...
+    final ZooKeeperClient client = provider.get("getLastRollingOperation");
+    return getLastRollingOperation(client, groupName);
+  }
+
+  private RollingOperation getLastRollingOperation(final ZooKeeperClient client,
+                                                       final String groupName)
+      throws DeploymentGroupDoesNotExistException {
+    // The most recent rolling operations should always be at the head of the list returned by
+    // getRollingOperations().
+    for (final String rollingOpId : getRollingOperationIds(client, groupName)) {
```
davidxia: Check if the list is not empty and `.get(0)` instead?

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperation.java
```diff
@@ -0,0 +1,247 @@
...
+        this.reason = Reason.MANUAL;
+        this.rolloutOptions = null;
+      }
+    }
+
+    public String getId() {
+      return p.id;
+    }
+
+    public Builder setId(final String id) {
```
davidxia: Is there a reason to have this setter? We use it in some tests, but I don't see us testing the Id itself (and we shouldn't, I think).

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -448,36 +533,42 @@ public void updateDeploymentGroupHosts(final String groupName, final List<String
...
+        final List<ZooKeeperOperation> operations = Lists.newArrayList();
+        operations.add(set(Paths.statusDeploymentGroupHosts(groupName), Json.asBytes(hosts)));
+
+        final RollingOperation lastOp = getLastRollingOperation(client, groupName);
+        if (lastOp == null) {
+          // This deployment group has never been deployed to, so don't deploy to the new hosts.
+          log.info("starting zookeeper transaction for updateDeploymentGroupHosts on "
+                   + "deployment-group name {}. List of operations: {}",
+                   groupName, operations);
+          client.transaction(operations);
```
davidxia: Previously, if the DG didn't exist, no ZK data would be changed. Do we want to update this znode if the deployment group doesn't exist?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -486,187 +577,273 @@ public void updateDeploymentGroupHosts(final String groupName, final List<String
...
+        .setState(NEW)
+        .build();
 
-    operations.add(set(Paths.configDeploymentGroup(updated.getName()), updated));
 
     try {
-      final RollingUpdateOp op = getInitRollingUpdateOps(updated, MANUAL, client);
-      operations.addAll(op.operations());
+      // Fail early if the deployment group does not exist.
+      if (client.exists(Paths.configDeploymentGroup(rolling.getDeploymentGroupName())) == null) {
```
davidxia: Why not use `ZooKeeperMasterModel.getDeploymentGroup()`?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -486,187 +577,273 @@ public void updateDeploymentGroupHosts(final String groupName, final List<String
...
+        throw new DeploymentGroupDoesNotExistException(rolling.getDeploymentGroupName());
+      }
 
-      log.info("starting zookeeper transaction for rolling-update on "
-               + "deployment-group name={} jobId={}. List of operations: {}",
-          deploymentGroup.getName(), jobId, operations);
+      // Create the base paths for rolling operations, if necessary.
+      client.ensurePath(Paths.configRollingOps());
```
davidxia: placeholder for me. I've only reviewed up to here.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: Most likely it just didn't occur to me at the time. Though possibly it was to reuse the existing `ZookeeperClient`. I suppose I could overload `getDeploymentGroup` to allow passing in a client.

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperation.java
negz: No, this was mostly just for testing/internal purposes.

How might I ensure predictable IDs for rolling operations in tests if I removed this?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: Is this a style preference, or are there other advantages? I've always felt this way was cleaner. (I think I remember you talking about this with @gimaker recently?)

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: For now yes (I didn't know about `@VisibleForTesting`), but I expect we'll want to expose it in future so we can add functionality to list recent rolling ops for a DG via the API/CLI.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: No, I think that's an oversight. Good catch.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: Not sure. I'm afraid you'd have to ask whoever wrote it originally.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: Yeah, oversight on my part I think.

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperationStatus.java
negz: We're not very consistent with this at the moment. My implementation follows what pattern I can see - i.e. this isn't used in `DeploymentGroupStatus` either (but is in `DeploymentGroup`).

FWIW I'm not a fan of the concept in general. I prefer a JSON API with `nil` values than a JSON API that returns different keys under different circumstances.

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperationTasks.java
negz: We're not very consistent with this at the moment. My implementation follows what pattern I can see - i.e. this wasn't used in `DeploymentGroupTasks` either (but is in `DeploymentGroup`).

FWIW I'm not a fan of the concept in general. I prefer a JSON API with `nil` values than a JSON API that returns different keys under different circumstances.

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperation.java
davidxia: Just test they exist?

helios-client/src/test/java/com/spotify/helios/common/descriptors/RollingOperationTest.java
negz: Thanks. Could we make our linter catch this?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
davidxia: Yea, I think it's clearer/more readable to `.get(0)` when you actually mean to get the first item. [In this case with staffan](https://github.com/spotify/helios/pull/913/files#diff-3af83d8153f42fbd08f07e93b7432ccdR118), it actually made sense to use a for loop because I was actually iterating through the whole way.

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperationStatus.java
davidxia: Hm, we actually don't use NON_NULL anywhere expect `DeploymentGroup`. Just leave this as is then.

helios-client/src/test/java/com/spotify/helios/common/descriptors/RollingOperationTest.java
davidxia: My IntelliJ is catching this. But sure, add something to checkstyle.xml.

helios-client/src/test/java/com/spotify/helios/common/descriptors/RollingOperationTest.java
negz: I don't think testing RolloutOptions fits here. I've added a separate test for RolloutOptions.

davidxia: Sorry, I must've missed that one. Can you link me to it?

negz: I mean I added one in response to your comment. I'll push it soon.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: I actually have no idea why this changed. Might have been a find and replace snafu.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: Left it public but added it to the interface for future use.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: Okaaaay but I prefer my way. :(

helios-client/src/main/java/com/spotify/helios/common/descriptors/RollingOperation.java
negz: https://github.com/spotify/helios/pull/903/files#diff-64b87e65f5290ca066c979f34868c419R135

The above test currently requires me to create RollingOperations with a predictable ID, because I can't mock out calls to ZK paths composed of UUID that I don't know in advance of its creation.

I could possibly avoid this situation by breaking said test up into a few smaller ones, but it doesn't seem that bad to me to expose a knob to override the randomly generated UUID even if we don't use it much.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: Isn't that still the case?

My understanding is that previously it would hit `client.getNode(Paths.statusDeploymentGroupHosts(groupName))`, raise a `NoNodeException`, and thus `return` if the DG didn't exist. That should still be the case.

At the line you've commented on it's handling the case where the deployment group exists, but has never been deployed to before. So `client.transaction(operations)` should only be committing the operation that adds the updated hots.

What am I missing?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: Done (except the tasks bit per your other comment).

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java

mattnworb:
88 comments 😲

negz:
We decided not to proceed with this PR in our retro today. We like the change conceptually, but we're concerned that the time and effort required to validate and deploy it outweighs its benefits. I'll leave the code in my `rollup` branch.

PR #904, opened 2016-04-22
Add "in" and "notin" host selector operators

gimaker:
Examples:
 * role in (foo, bar)
 * some_label notin (qux,foo)

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/HostSelector.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/HostSelector.java
@@ -49,6 +49,30 @@ public class HostSelector extends Descriptor {
       public boolean test(final String a, final Object b) {
         return !Objects.equals(a, b);
       }
+    }),
+    IN("in", new BiPredicate<String, Object>() {
+      @Override
+      public boolean test(final String a, final Object b) {
+        final String[] parts = ((String) b).replaceAll("(\\(|\\)| )", "").split(",");
+        for (final String part : parts) {
+          if (part.equals(a)) {
+            return true;
+          }
+        }
+        return false;
+      }
+    }),
+    NOT_IN("notin", new BiPredicate<String, Object>() {
+      @Override
+      public boolean test(final String a, final Object b) {
+        final String[] parts = ((String) b).replaceAll("(\\(|\\)| )", "").split(",");
+        for (final String part : parts) {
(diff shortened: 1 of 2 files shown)


gimaker:
@mattnworb @davidxia @negz @rohansingh

davidxia:
Would be nice to have docs :)

gimaker:
> Would be nice to have docs :)

PRs welcome :)

helios-client/src/main/java/com/spotify/helios/common/descriptors/HostSelector.java
```diff
@@ -49,6 +49,30 @@ public boolean test(final String a, final Object b) {
...
+      @Override
+      public boolean test(final String a, final Object b) {
+        final String[] parts = ((String) b).replaceAll("(\\(|\\)| )", "").split(",");
+        for (final String part : parts) {
+          if (part.equals(a)) {
+            return false;
+          }
+        }
+        return true;
+      }
```
mattnworb: This looks to be the same logic except the boolean values are reversed. How about making re-using the IN definition to avoid duplication?

```java
NOT_IN("notin", new BiPredicate<String, Object>() {
  @Override
  public boolean test(final String a, final Object b) {
    return !IN.test(a, b);
  }
}
```

mattnworb:
tangential question but why is the second type of BiPredicate in Operator declared as `<String, Object>`, for example

```java
Operator(final String operatorName, final BiPredicate<String, Object> predicate)
```

when none of the implementations match a non-String value? i.e. why not `BiPredicate<String, String>`?

davidxia:
I had the same question. Everything seems to still work if I change `Object operand` to `String operand` and `BiPredicate<String, Object> predicate` to `BiPredicate<String, String> predicate`.

gimaker:
> tangential question but why is the type of BiPredicate in Operator declared as <String, Object>, for example

I believe it was to allow cases where the "operand" is not a string -- like in this case, where technically it's a set.

Still on the fence if I should have the `parse()` method parse the set from the string and let the in and notin operators deal with sets, or if I should just do it the way it is now. Also not sure it matters much, and neither seems ideal. If you go with the latter changing it to `BiPredicate<String, String>` would make sense but it's tehnically an API break so would rather not (even though it should be safe).

mattnworb:
👍

davidxia:
👍

PR #905, opened 2016-04-25
fix JobCreateCommand: don't overwrite capabilities from json file

mattnworb:
Added or dropped capabilities from a job json file were overwritten if the CLI switch had an empty value (i.e. was not specified).


mattnworb:
@davidxia @gimaker @negz

helios-integration-tests/src/test/resources/logback-test.xml
```diff
@@ -9,7 +9,6 @@
   </appender>
 
   <logger name="com.spotify.helios.client" level="info"/>
-  <logger name="com.spotify.hermes" level="info"/>
```
gimaker: spurious change, or intentional?

mattnworb: intentional; why was this line ever here? No code in helios integration tests has this package name.

My guess is it was copy & pasted from somewhere else.

helios-tools/src/main/java/com/spotify/helios/cli/command/JobCreateCommand.java
```diff
@@ -546,8 +546,15 @@ int run(final Namespace options, final HeliosClient client, final PrintStream ou
       builder.setToken(token);
     }
 
-    builder.setAddCapabilities(options.<String>getList(addCapabilityArg.getDest()));
-    builder.setDropCapabilities(options.<String>getList(dropCapabilityArg.getDest()));
+    final List<String> addCaps = options.<String>getList(addCapabilityArg.getDest());
```
gimaker: For port mappings and registrations, it seems like we merge the values if they come from more than one source. Perhaps do the same here for consistency?

helios-integration-tests/src/test/resources/logback-test.xml
gimaker: If you're cleaning up, maybe remove `<logger name="com.spotify.helios.client" level="info"/>` as well then?

gimaker:
:+1:, comments aside

helios-tools/src/main/java/com/spotify/helios/cli/command/JobCreateCommand.java
mattnworb: I don't have a strong preference, just trying to fix a broken part of the current implementation

mattnworb:
cc @mzeo

PR #907, opened 2016-04-25
Fix equality comparisons of in and notin host selectors

gimaker:
This fixes an issue where, for example, the following host selectors
were considered different:
 * a in (a, b)
 * a in (a,b)
 * a in (b, a)
 * a in (a, a, b)
 * ...

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/HostSelector.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/HostSelector.java
@@ -17,11 +17,17 @@
 
 package com.spotify.helios.common.descriptors;
 
+import com.google.common.collect.Iterables;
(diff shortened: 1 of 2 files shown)


gimaker:
@negz @davidxia @mattnworb

davidxia:
LGTM

mattnworb:
👍  but where was it broken before? I can't tell what the fix is here

gimaker:
>  ￼but where was it broken before? I can't tell what the fix is here

@mattnworb equality was broken before. `a in (a,b)` and `a in (a, b)` were considered different even they aren't really (the extra space does not matter). This in turn can result in conflicts when creating deployments. E.g. instead of getting `NOT_MODIFIED` you'd get `CONFLICT` by just changing the formatting of the host selectors.

PR #908, opened 2016-04-25
fix: target/docker/image-name ends in a newline

mattnworb:
Make sure that the contents of the file are stripped of newlines when
reading `target/docker/image-name`.


mattnworb:
I swear I've tested this logic this time. 

@spotify/helios-team

PR #910, opened 2016-05-03
[solo] Randomize solo agent IDs

davidxia:
Generate random helios solo agent IDs so concurrent solo instances
won't affect each other, e.g. reap each others jobs.


mattnworb:
👍

mattnworb:
codecov complains because the diff doesn't touch testable files

```
Nothing to diff

This commit did not change files that are tracked by Codecov.
```

davidxia:
Thanks to @negz for pointing out that we have code in `HeliosSoloDeployment` that sets env vars for the solo startup. I think it's better to put this logic there as we're doing something similar for `--name`.

I think name is used for agent registration/hostname and the id is used to make sure the same hostname won't take over unless it also has the same id. This id is also used for the namespace in `Reaper`.

gimaker:
:+1:

negz:
LGTM

PR #911, opened 2016-05-04
Add support for reaper grace period

spkrka:


spkrka:
Ping @mattnworb @davidxia @gimaker @rohansingh 

I might need help adding this setting for helios-solo. My idea is to set the grace period to something like a day for helios-solo, which would keep containers alive for the period of the integration test even if there are multiple helios-solo instances running.

mattnworb:
Thanks! I think this makes a lot of sense to have in general. I also think that #910 will fix the multiple-helios-solo-collision issue.

If you are interested in adding tests for this new logic, I think `com.spotify.helios.system.ReapingTest` would be the right place 😉 

To set this for helios-solo, I think you would just want to modify `solo/docker/start.sh` to pass the argument to `AgentMain`.

spkrka:
Yes, if #920 solves it, I guess my primary motivation disappears. Not sure if this is still useful in general, though I do think the current reaping might be overly aggressive.

If you want to adopt this PR that's fine, otherwise I can close it.

davidxia:
@spkrka Thanks for your PR. I just released 0.8.780 which has this patch https://github.com/spotify/helios/pull/910. Do you mind if we close this? We can always reopen and adopt if we find we need it later.

spkrka:
Yes, that seems like a good plan!

PR #912, opened 2016-05-04
Try to improve connection related error messages

spkrka:


spkrka:
Ping @mattnworb @davidxia @rohansingh @gimaker

mattnworb:
Thanks, great idea. But I'm afraid that [this logic in RetryingRequestDispatcher](https://github.com/spotify/helios/blob/3e01679e81a9baed9744a1fceb8ae28541c96bc1/helios-client/src/main/java/com/spotify/helios/client/RetryingRequestDispatcher.java#L108-L111) will make this irrelevant - IIRC the underlying exception is only logged when the retry loop has finished.

Perhaps that class should log the exception message on each try, combined with the addition of `ipUri` in this commit

spkrka:
Thanks, perhaps the errors I saw when running helios-solo was from an older version, so this change is pointless. Should I close the PR?

mattnworb:
personally I think it would make more sense if the error message from the RetryingRequestDispatcher was more informative than

```
Failed to connect, retrying in 5 seconds.
Failed to connect, retrying in 5 seconds.
```

i.e. if it included the error message from your change as well. So I'd be in favor of keeping this and modifying the other class too, but I am curious what the rest of the team thinks as well.

spkrka:
I agree!

davidxia:
@mattnworb @spkrka @gimaker @negz How about this? https://github.com/spotify/helios/pull/914

PR #914, opened 2016-05-05
Improve client connection error messages

davidxia:
Show the URI of the master the client is trying
to connect to when errors occur for debugging purposes.


davidxia:
@gimaker @negz @mattnworb

gimaker:
:+1:

PR #917, opened 2016-05-12
WIP: helios-testing: TempJobs manages HeliosSoloDeployment

davidxia:
instead of user needing to manage them.

I.e.
```
@Rule public TemporaryJobs temporaryJobs = TemporaryJobs.create();
```

instead of

```
private static final HeliosDeploymentResource soloResource =
  new HeliosDeploymentResource(HeliosSoloDeployment.fromEnv().build());

private static final TemporaryJobs temporaryJobs =
  TemporaryJobs.builder().client(soloResource.client()).build();

@ClassRule
public static final RuleChain chain = RuleChain
    .outerRule(soloResource)
    .around(temporaryJobs);
```


davidxia:
@negz @gimaker @mattnworb

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -103,12 +115,20 @@
   private final HeliosClient heliosClient;
   private boolean removeHeliosSoloContainerOnExit;
   private final int jobUndeployWaitSeconds;
+  private final SoloMasterProber soloMasterProber;
+  private final SoloHostProber soloHostProber;
+
+  private static volatile HeliosDeployment soloDeployment;
```
mattnworb: this field is only used within the`getOrCreateHeliosSoloDeployment()` method; I think it would be cleaner if the reference to the static HeliosSoloDeployment was moved to TemporaryJobs rather than the HeliosSoloDeployment class itself.

In other words - make the class that wants to use HSD and doesn't care if it gets an old instance or creates a new instance be the thing that holds the reference.

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -74,18 +85,19 @@
  * master and one Helios agent deployed in Docker. Helios Solo uses the Docker instance it is
  * deployed on to run its jobs.
  */
-public class HeliosSoloDeployment implements HeliosDeployment {
+class HeliosSoloDeployment implements HeliosDeployment {
```
mattnworb: changing the access modifier will force any users of the previous style of HeliosSoloDeployment/TemporaryJobs fail to compile - is that intentional?

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -144,6 +168,94 @@
             .setUser(username)
             .setEndpoints("http://" + deploymentAddress)
             .build());
+
+    connectToWatchdog(dockerHost.address(), watchdogPort);
+
+    try {
+      start();
+    } catch (TimeoutException e) {
+      Throwables.propagate(e);
```
mattnworb: usual calling style here is `throw Throwables.propogate ...`

helios-testing/src/main/java/com/spotify/helios/testing/SoloHostProber.java
```diff
@@ -0,0 +1,37 @@
...
+ * A interface that extends {@link Callable}.
+ * This interface is used by {@link HeliosSoloDeployment} to check whether there are agents
+ * available in helios-solo. The main purpose of this interface is to make HeliosSoloDeployment
+ * easy to test by allowing us to easily mock the helios client calls to the master in helios-solo.
+ */
+interface SoloHostProber extends Callable<Boolean> {
+
+  void setHostAndPort(HostAndPort hostAndPort);
+
+  void setHeliosClient(HeliosClient heliosClient);
```
mattnworb: Assuming that the purpose of extending `Callable` is to use it with `Polling.awaitUnchecked(..)`, I think it would be clearer instead to have something like

```java
interface SoloHostProber {
    boolean check(HeliosClient client, HostAndPort hap);
}

// in the place where it is used
Polling.awaitUnchecked(time, new Callable() { return soloHostProber.check(client, hap); });
```

then you don't have to remember to properly call some methods in-order before others. It would also reduce boilerplate in the implementations of these interfaces.

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -144,6 +168,94 @@
...
+
+  private static void connectToWatchdog(final String host, final int port) {
+    final ExecutorService executorService = Executors.newSingleThreadExecutor(
+        new ThreadFactoryBuilder()
+            .setNameFormat("helios-solo-watchdog-%d")
+            .setDaemon(true).build());
+
+    log.info("Connecting to helios-solo watchdog at {}:{}", host, port);
+
+    executorService.submit(new Runnable() {
```
mattnworb: might make sense to break this anonymous class into it's own class to make it easier to test on it's own

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -144,6 +168,94 @@
...
+            final DataOutputStream writer = new DataOutputStream(s.getOutputStream());
+            final BufferedReader reader = new BufferedReader(
+                new InputStreamReader(s.getInputStream()));
+            writer.writeBytes("HELO\n");
+            writer.flush();
+            final String line = reader.readLine();
+            if (line.startsWith("HELO")) {
+              break;
+            } else {
+              s.close();
```
mattnworb: why would the line not be HELO? would this be a case where you want to fail totally? Maybe at least log something.

mattnworb:
Regarding the watchdog - it feels weird to launch a thread that connects a socket and then blocks forever (or attempts to do so via sleeping in a loop that never finishes). Also, the class that launches this thread doesn't wait for the connection to be established - should it? You could avoid the need to have a thread block forever (which I assume is to keep the socket that it holds from being GCed) by moving the socket/connection up to the `HeliosSoloDeployment` class level, and have the class establish the connection before moving on with it's other work. Plus then you can cleanly close the connection / watchdog when HSD is closed.

mattnworb:
Would this change completely remove support for the `-Dhelios.testing.profile` switch? So any user who was using profiles to be able to specify a helios cluster to deploy their TemporaryJobs would transparently lose that functionality with this change?

solo/docker/start.sh
```diff
@@ -82,6 +82,37 @@ com.spotify.helios.master.MasterMain \
...
+        states=$($HELIOS_CLI_CMD status --json -j "$job" | jq -r ".[].taskStatuses | .[] | .state")
+        tot=$(echo "$states" | wc -l)
+        stopped=$(echo "$states" | grep "STOPPED" | wc -l)
+        if [ "$stopped" -eq "$tot" ]; then
+            echo "All instances of $job are stopped"
+            break
+        fi
+    done
+
+    echo "Some containers may still be running at this point, but they'll eventually be killed.\n
```
mattnworb: I could be wrong but I don't think echo will expand `\n`

davidxia: Oops, I meant printf

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
davidxia: sounds good

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
davidxia: Yes, the idea is for users to not to use `HeliosSoloDeployment` at all. I'm also planning on making breaking changes to remove the innumerable configuration options.

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
davidxia: missed that one, thanks

helios-testing/src/main/java/com/spotify/helios/testing/SoloHostProber.java
davidxia: Thanks! This is much better.

davidxia:
Maybe we can avoid the whole watchdog mechanism and use a JVM shutdown hook instead. I'll try that out.

solo/docker/start.sh
gimaker: `echo -e` expands `\n` and so does `printf` (plain `echo` does not)

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
davidxia: I'll log something.

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -563,15 +645,17 @@ public String toString() {
     private DockerHost dockerHost;
     private DockerHost containerDockerHost;
     private HeliosClient heliosClient;
-    private String heliosSoloImage = "spotify/helios-solo:latest";
+    private String heliosSoloImage = "spotify/helios-solo:TEST";
```
negz: Why change this?

solo/docker/start.sh
```diff
@@ -82,6 +82,37 @@ com.spotify.helios.master.MasterMain \
...
+    HELIOS_CLI_CMD="java -cp /*.jar -Xmx128m -Djava.net.preferIPv4Stack=true com.spotify.helios.cli.CliMain -z http://localhost:5801"
+
+    python "/watchdog.py"
+
+    echo "Cleaning up jobs..."
+
+    # Clean up created jobs (make sure containers are stopped)
+    jobs=$($HELIOS_CLI_CMD --json jobs | jq -a -r "keys[]")
+    for job in $jobs; do
+	      $HELIOS_CLI_CMD undeploy --yes -a "$job"
```
gimaker: nitpick: mixed tabs and spaces here (and I'm probably to blame for it!)

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -737,4 +843,100 @@ public HeliosDeployment build() {
...
+            "HeliosMasterSocketProber.hostAndPort must be set first.");
+      }
+
+      final SocketAddress address = new InetSocketAddress(hostAndPort.getHostText(),
+                                                          hostAndPort.getPort());
+      log.debug("attempting to connect to {}", address);
+
+      try {
+        final Socket s = new Socket();
+        s.connect(address, 100);
```
negz: Would this be susceptible to the same issue described in `connectToWatchdog()`? i.e. Do we need to read/write some data to ensure it's really connected?

helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
```diff
@@ -195,6 +192,10 @@ void after(final Optional<TemporaryJobReports.ReportWriter> writer) {
 
     for (final TemporaryJob job : jobs) {
       jobIds.add(job.job().getId());
+      // TODO (dxia) Undeploying a job doesn't gauruntee the container will be stopped immediately.
```
negz: Typo in guarantee. :)

solo/docker/watchdog.py
```diff
@@ -0,0 +1,62 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2015 Spotify AB
+
+import socket
+import asyncore
+import time
+import sys
+import os
+
+PORT=33333
```
negz: Is it worth setting this as an env var, rather than having it static in the Python script?

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -645,7 +646,7 @@ public String toString() {
     private DockerHost dockerHost;
     private DockerHost containerDockerHost;
     private HeliosClient heliosClient;
-    private String heliosSoloImage = "spotify/helios-solo:TEST";
+    private String heliosSoloImage = "spotify/helios-solo:latest";
```
negz: Oh, I see now this goes back to latest here.

helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
davidxia: This is one of the words I always misspell. Along with diarrhea

davidxia:
@mattnworb Regarding https://github.com/spotify/helios/pull/917#issuecomment-219407338, yes. I was planning on this TempJob change only supporting helios-solo.

mattnworb:
@davidxia can you update `docs/testing_framework.md ` too then? (ignoring the question of if it's a good idea to break it like this since we have a lot of internal code using the profile args)

circle.sh
```diff
@@ -41,6 +41,16 @@ case "$1" in
 
     case $CIRCLE_NODE_INDEX in
       0)
+        sudo apt-get install -y jq
```
gimaker: nice!

helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
```diff
@@ -68,32 +59,25 @@
...
-import static com.google.common.base.Strings.isNullOrEmpty;
 import static com.google.common.collect.Lists.newArrayList;
-import static com.spotify.helios.testing.Jobs.undeploy;
 import static java.lang.String.format;
-import static java.lang.System.getProperty;
-import static java.util.Arrays.asList;
 import static java.util.concurrent.TimeUnit.MINUTES;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
 public class TemporaryJobs implements TestRule {
```
gimaker: If we're making breaking changes already, I think it'd make sense to split this class up into two parts:

 * One class with all the helios logic, but without any JUnit stuff. Make it AutoCloseable and usable when you don't use JUnit.
 * One simple class that adds the JUnit fluff.

davidxia: 👍

mattnworb: This would make it very simple to add TestNG support too, and simplify the case where people want to use TemporaryJobs in something that isn't a test.

davidxia:
Replaced by #922 as neither staffan nor I could figure out how to make tests pass. #922 has smaller, easier to understand commits that all pass tests.

solo/docker/watchdog.py
davidxia: good idea. It'll prevent mistakes if we ever change the port.

circle.sh
davidxia: Why nice?

PR #920, opened 2016-05-18
TemporaryJobs: handle DOCKER_HOST with unix sockets

mattnworb:
`TemporaryJobs.builder()` attempts to set up default values for the
`endpoints` to connect to (representing the address of the Helios
cluster to talk to), prior to the user setting any endpoints or
HeliosClient of their own.

After checking a few environment variables for which Helios host to talk
to, `TemporaryJobs.Builder` then checks if `DOCKER_HOST` is set and uses
it as the host in a `http://` URI.

This fails if `DOCKER_HOST` is a URI to a Unix socket (e.g.
`unix:///var/run/docker.sock`), such as when using Docker for Mac.

To fix this, change the Builder to ignore `DOCKER_HOST` values that
start with `unix://`, and rely on the user to set the endpoints or
HeliosClient to use themselves.

### DIFF
--- a/helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
@@ -596,7 +596,7 @@ public class TemporaryJobs implements TestRule {
       final String dockerHost = env.get("DOCKER_HOST");
       if (dockerHost == null) {
         endpoints("http://localhost:5801");
-      } else {
+      } else if (!dockerHost.startsWith("unix://")) {
         try {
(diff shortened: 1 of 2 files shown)


mattnworb:
@spotify/helios-team

negz:
LGTM

Though ideally `TemporaryJobs` would forego *all* the magic Helios-client-discovery logic if someone explicitly passes a client to the `Builder`.

mattnworb:
yeah, it is kind of silly to have all this default logic and construction of things in the constructor of `Builder` rather than in it's `build()` method.

gimaker:
:+1:

Issue #916, opened 2016-05-11
Helios Testing broken with Docker for Mac (beta)


daenney:
The Docker for Mac beta ships a Docker setup packaged and well running on top of xhyve. Though the `docker` CLI is smart enough to find it (it knows where the socket is) other tools that look for the docker daemon need the `DOCKER_HOST` environment variable to be set. In this case, it should point at `unix:///var/tmp/docker.sock`.

Unfortunately setting `DOCKER_HOST` seems to break the `TemporaryJobs.builder()` as that contains a fair amount of glue to try and figure out where the heck Docker went. This results in the following stack trace:

```
[main] WARN com.spotify.helios.client.Endpoints - Unable to resolve hostname null into IP address
java.net.UnknownHostException: null: unknown error
    at java.net.Inet6AddressImpl.lookupAllHostAddr(Native Method)
    at java.net.InetAddress$2.lookupAllHostAddr(InetAddress.java:928)
    at java.net.InetAddress.getAddressesFromNameService(InetAddress.java:1323)
    at java.net.InetAddress.getAllByName0(InetAddress.java:1276)
    at java.net.InetAddress.getAllByName(InetAddress.java:1192)
    at java.net.InetAddress.getAllByName(InetAddress.java:1126)
    at org.apache.http.impl.conn.SystemDefaultDnsResolver.resolve(SystemDefaultDnsResolver.java:45)
    at com.spotify.helios.client.Endpoints.of(Endpoints.java:92)
    at com.spotify.helios.client.Endpoints.of(Endpoints.java:79)
    at com.spotify.helios.client.HeliosClient$Builder.setEndpoints(HeliosClient.java:532)
    at com.spotify.helios.client.HeliosClient$Builder.setEndpointStrings(HeliosClient.java:548)
    at com.spotify.helios.testing.TemporaryJobs$Builder.endpointStrings(TemporaryJobs.java:663)
    at com.spotify.helios.testing.TemporaryJobs$Builder.endpoints(TemporaryJobs.java:658)
    at com.spotify.helios.testing.TemporaryJobs$Builder.configureWithEnv(TemporaryJobs.java:602)
    at com.spotify.helios.testing.TemporaryJobs$Builder.<init>(TemporaryJobs.java:562)
    at com.spotify.helios.testing.TemporaryJobs.builder(TemporaryJobs.java:507)
    at com.spotify.helios.testing.TemporaryJobs.builder(TemporaryJobs.java:502)
[...]

PR #921, opened 2016-05-18
exec healthchecks: remove check of executionDriver

mattnworb:
There are some environments, such as docker-machine on Mac or Docker for
Mac, where `docker exec` works just fine but they seem to return an
empty string for `ExecutionDriver` from the `/info` endpoint in the
Docker Remote API.

This change removes the check that only executes the exec healthcheck
when `executionDriver` starts with `native`, so that exec healthchecks
can still be used (and jobs with them can be deployed) in these
environments.


mattnworb:
@spotify/helios-team @mbruggmann

mattnworb:
This is related to #916 but also I was surprised to see that an exec healthcheck didn't work when deploying things to a local helios-solo with docker-machine / vmware.

PR #922, opened 2016-05-19
WIP: helios-testing: refactor with breaking changes

davidxia:
This PR makes the following changes in order:
- Make `TemporaryJobs` instead of the user manage `HeliosSoloDeployment`
- Make `HeliosSoloDeployment` package local
- Make helios-solo container have logic to clean up and kill itself to prevent leaked containers even when `TemporaryJobs` is SIGKILLed
- one `HeliosSoloDeployment` per JVM process per test run
- split TemporaryJobs into two classes - one that implements AutoCloseable and another that implements JUnit's Rule

TODOs:
- update docs
- delete other stuff?


davidxia:
CircleCI now takes twice as long to build :) I suspect many of the tests no longer make sense, eg HeliosIT or the tests that inherit `TemporaryJobsTestBase`.

davidxia:
Tests became flaky starting with at least commit e014893. Ignoring the mind-blown inception of HeliosIT deploying a helios master, ZK, and agent as separate containers using helios-solo, how is the agent job already deployed? https://circleci.com/gh/spotify/helios/4547. I don't see any other instances where we deploy the agent as a job.

mattnworb:
is this ready for review? I am confused by the "WIP" titles

davidxia:
@gimaker Would you say this one is ready for review?

davidxia:
@mattnworb This is ready for review. I'm trying to keep each commit green and reviewable by itself when taken in order.

mattnworb:
@davidxia great, will take a look today. Can we merge and release #927 first though, so that users can upgrade to it and not have to deal with any incompatibilities before having to deal with the incompatibilities in this change?

davidxia:
Yes, we can.

On Thu, May 26, 2016 at 9:35 AM, Matt Brown <notifications@github.com>
wrote:

> @davidxia <https://github.com/davidxia> great, will take a look today.
> Can we merge and release #927 <https://github.com/spotify/helios/pull/927>
> first though, so that users can upgrade to it and not have to deal with any
> incompatibilities before having to deal with the incompatibilities in this
> change?
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/spotify/helios/pull/922#issuecomment-221872106>
>



-- 
David Xia
Software Engineer
dxia@spotify.com

gimaker:
Reviewable, yes. Releasable... maybe.

helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
```diff
@@ -136,46 +113,6 @@ private static synchronized HeliosDeployment getOrCreateHeliosSoloDeployment() {
...
   }
 
-  /**
-   * Perform setup. This is normally called by JUnit when TemporaryJobs is used with @Rule.
-   * If @Rule cannot be used, call this method before calling {@link #job()}.
-   *
-   * Note: When not being used as a @Rule, jobs will not be monitored during test runs.
-   */
-  public void before() {
-    deployer.readyToDeploy();
```
gimaker: The coupling between `DefaultDeployer` and `TemporaryJobsResource` is pretty weird.

Note that `TemporaryJobs` is currently not usable without `TemporaryJobsResource` (== I think this commit needs some more work). If you try to use `TemporaryJobs` with `TemporaryJobsResource` then `Deployer.readyToDeploy()` will not be called, which means that `TemporaryJobs.deploy()` will throw an exception.

circle.sh
```diff
@@ -41,6 +41,16 @@ case "$1" in
...
       0)
+        # We need to build the helios-solo image because we use TemporaryJobs in helios-testing
+        # tests, and TemporaryJobs uses the helios-solo container.
+        sudo apt-get install -y jq
+        mvn -P build-images -P build-solo package -DskipTests=true -Dmaven.javadoc.skip=true \
+          -B -V -pl helios-services
+
+        # tag the helios-solo image we just built
+        solo_image=$(cat helios-services/target/test-classes/solo-image.json | jq -r '.image')
+        docker tag -f $solo_image spotify/helios-solo:latest
```
mattnworb: this looks like duplication from what is under the `5)` branch

1. if it is really necessary to duplicate, can these commands be moved to a function?
2. won't there be a race between building and tagging the image as `spotify/helios-solo:latest`
3. would it be simpler to move the tests that need this image to be executed by the `5)` node?

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -31,62 +31,61 @@
...
+  private static final String BOOT2DOCKER_SIGNATURE = "Boot2Docker";
+  static final String PROBE_IMAGE = "onescience/alpine:latest";
+  private static final String HELIOS_NAME_SUFFIX = ".solo.local"; //  Required for SkyDNS discovery.
+  private static final String HELIOS_ID_SUFFIX = "-solo-host";
+  static final String HELIOS_CONTAINER_PREFIX = "helios-solo-container-";
+  static final String HELIOS_PROBE_PREFIX = "helios-probe-";
+  private static final String HELIOS_SOLO_PROFILE = "helios.solo.profile";
+  private static final String HELIOS_SOLO_PROFILES = "helios.solo.profiles.";
+  static final int HELIOS_MASTER_PORT = 5801;
+  static final int HELIOS_SOLO_WATCHDOG_PORT = 33333;
```
mattnworb: random question: why the mix of private and package-private access levels? It seems unnecessary to me to default everything to private and then carve out exceptions for the ones that happen to be used by a test - because if you can relax the access level as soon as it's needed, then what is the point of ever having it be private 😄  I would just make them all the same minimum level (package private in this case)

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -122,29 +123,100 @@
...
+          } catch (IOException e) {
+            return null;
+          }
+        }
+      });
+    } catch (TimeoutException e) {
+      throw Throwables.propagate(e);
+    }
+
+    log.info("Connecting to helios-solo watchdog at {}:{}", host, port);
```
mattnworb: if the connection is already made at this point, the log line should really be "*Connected* to helios-solo watchdog at ..."

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -706,54 +732,77 @@ public Builder heliosClient(final HeliosClient heliosClient) {
...
 
     /**
      * Configures, deploys, and returns a {@link HeliosSoloDeployment} using the as specified by
      * this Builder.
      *
-     * @return A Helios Solo deployment configured by this Builder.
+     * @return A HeliosSoloDeployment configured by this Builder.
      */
-    public HeliosSoloDeployment build() {
+    public HeliosDeployment build() {
```
mattnworb: declaring that the `build()` method returns a superclass isn't really useful at all since the class is `HeliosSoloDeployment.Builder`, and it's not like the `Builder` itself has any parent classes where someone might refer to it as `HeliosDeployment.Builder` etc. 

Worst case this type of change forces code using this to have to cast the `build()` result to `(HeliosSoloDeployment)` which is just unnecessary. 

(also I am a little sour since this undoes something I just did in https://github.com/spotify/helios/commit/144600b120bd095dd2882f0dd63a5a065949056e )

helios-testing/src/main/java/com/spotify/helios/testing/SoloAgentProber.java
```diff
@@ -0,0 +1,85 @@
...
+ * This class is used by {@link HeliosSoloDeployment} to check whether there are agents
+ * available in helios-solo.
+ */
+class SoloAgentProber {
+
+  private static final Logger log = LoggerFactory.getLogger(SoloAgentProber.class);
+
+  /**
+   * @param client {@link HeliosClient}
+   * @return true if we can connect to the specified host and port. Null, otherwise.
```
mattnworb: I see that this signature is somewhat forced on this class by `Polling.awaitUnchecked`, but a method that can only ever return `true` or `null` feels pretty weird

helios-testing/src/main/java/com/spotify/helios/testing/SoloWatchdogConnector.java
```diff
@@ -0,0 +1,71 @@
...
+      writer.writeBytes("HELO\n");
+      writer.flush();
+      final String line = reader.readLine();
+      if (line.startsWith("HELO")) {
+        log.info("Connected to helios-solo watchdog");
+      } else {
+        throw new IOException("We didn't get back the HELO we sent to the watchdog process.");
+      }
+    } catch (IOException e) {
+      log.debug("Failed to connect to helios-solo watchdog");
```
mattnworb: why catch and log the exception here when it is just rethrown, rather than handling it once at the call-site?

helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
```diff
@@ -17,207 +17,100 @@
...
-  private final ThreadLocal<TemporaryJobReports.ReportWriter> reportWriter;
-
-  private final ExecutorService executor = MoreExecutors.getExitingExecutorService(
-      (ThreadPoolExecutor) Executors.newFixedThreadPool(
-          1, new ThreadFactoryBuilder()
-              .setNameFormat("helios-test-runner-%d")
-              .setDaemon(true)
-              .build()),
-      0, SECONDS);
+  private static volatile HeliosDeployment soloDeployment;
```
mattnworb: why does this class have both a static and member reference to a `HeliosDeployment` instance? that's confusing

mattnworb: ah nevermind, figured it out - but maybe some comments would help

mattnworb:
Is it no longer possible to control `REGISTRAR_HOST_FORMAT` via the typesafe config file? If not, that will break a large number of internal user's tests. 

If that functionality is unchanged then the TODOs in HeliosSoloIT seems incorrect

solo/docker/start.sh
```diff
@@ -82,6 +82,37 @@ com.spotify.helios.master.MasterMain \
 $HELIOS_MASTER_OPTS \
 &
 
-set +x
-# Sleep or execute command line
-while :; do sleep 1; done
+if [ "$HELIOS_SOLO_SUICIDE" -eq 1 ]; then
```
mattnworb: I think this may cause an error if `$HELIOS_SOLO_SUICIDE` is not defined

From a quick test:

```
+ '[' '' -eq 1 ']'
/tmp/su.sh: line 3: [: : integer expression expected
```

`1` should probably be wrapped in quotes

mattnworb: actually it should be `[[` and `]]` not `[ ]`

helios-testing/src/test/java/com/spotify/helios/testing/Utils.java
```diff
@@ -0,0 +1,44 @@
...
+import static org.mockito.Matchers.argThat;
+
+
+class Utils {
+
+  static InetSocketAddress socketWithHostAndPort(final String host, final int port) {
+    return argThat(new ArgumentMatcher<InetSocketAddress>() {
+      @Override
+      public boolean matches(Object argument) {
+        if (argument instanceof InetSocketAddress) {
```
mattnworb: this could be `CustomTypeSafeMatcher`

helios-testing/src/test/java/com/spotify/helios/testing/Utils.java
```diff
@@ -0,0 +1,44 @@
...
+
+
+import org.mockito.ArgumentMatcher;
+
+import java.net.InetSocketAddress;
+
+import static org.mockito.Matchers.argThat;
+
+
+class Utils {
```
mattnworb: maybe name this `TestMatchers` or something similar to make it clear it is about matchers and not a catch-all static method class?

helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
gimaker: IMO it'd make more sense to move the singleton to either a) the `HeliosSoloDeployment` class, or b) it's own class or such.

helios-testing/src/test/java/com/spotify/helios/testing/UdpProberTest.java
```diff
@@ -66,7 +66,7 @@ public void setup() {
     @After
     public void tearDown() {
       // The TemporaryJobs Rule above doesn't undeploy the job for some reason...
-      job.undeploy();
+      // temporaryJobs.deployer().undeploy(job.job(), job.hosts());
```
mattnworb: is this meant to be commented out? just remove the method if it has no actual code in it

mattnworb:
`docs/integration_tests_with_helios_solo.md` and probably `docs/testing_framework.md` should be updated as they are outdated now

helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
mattnworb: +1

helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
```diff
@@ -17,207 +17,100 @@
...
 
   TemporaryJobs(final Builder builder, final Config config) {
-    this.client = checkNotNull(builder.client, "client");
+    if (builder.heliosDeployment != null) {
+      this.heliosDeployment = builder.heliosDeployment;
+    } else {
+      this.heliosDeployment = getOrCreateHeliosSoloDeployment();
```
mattnworb: `this.heliosDeployment` is never used outside of this constructor, it doesn't need to be a field.

helios-testing/src/main/java/com/spotify/helios/testing/HeliosDeploymentResource.java
```diff
@@ -27,31 +26,26 @@
...
 import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
 /**
- * A HeliosDeploymentResource makes the supplied {@link HeliosDeployment} available to a JUnit
+ * A HeliosDeploymentResource makes the supplied {@link HeliosSoloDeployment} available to a JUnit
  * test, and guarantees to tear it down afterward.
  */
 public class HeliosDeploymentResource extends ExternalResource {
```
mattnworb: Is this class still necessary? It's not used anywhere anymore.

mattnworb:
If TemporaryJobs now managed HeliosSoloDeployment, is it intended or accidental that it doesn't close the HSD instance in `TemporaryJobs.close()`?

helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
```diff
@@ -499,35 +248,44 @@ static Builder builder(final Map<String, String> env) {
...
+      // Undeploying a job doesn't guarantee the container will be stopped immediately.
+      // Luckily TaskRunner tells docker to wait 120 seconds after stopping to kill the container.
+      // So containers that don't immediately stop **should** only stay around for at most 120
+      // seconds.
+      errors.addAll(undeploy(job.job(), job.hosts()));
+    }
+
+    for (final AssertionError error : errors) {
+      log.error(error.getMessage());
+    }
```
mattnworb: If TemporaryJobs now manages HeliosSoloDeployment, is it intended or accidental that it doesn't close the HSD instance here in `TemporaryJobs.close()`?

gimaker:
> If TemporaryJobs now managed HeliosSoloDeployment, is it intended or accidental that it doesn't close the HSD instance in TemporaryJobs.close()?

Intended, if you ask me. I think it makes sense to decouple the lifetime of the `HeliosDeployment` from the lifetime of the `TemporaryJobs` instance.

For example, you might want to reuse `HeliosDeployment` across multiple tests, but clean up jobs after each of them. I think it's fair to make the user responsible for closing the `HeliosDeployment`.

helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
gimaker: > If TemporaryJobs now managed HeliosSoloDeployment, is it intended or accidental that it doesn't close the HSD instance in TemporaryJobs.close()?

Intended, if you ask me. I think it makes sense to decouple the lifetime of the `HeliosDeployment` from the lifetime of the `TemporaryJobs` instance.

For example, you might want to reuse `HeliosDeployment` across multiple tests, but clean up jobs after each of them. I think it's fair to make the user responsible for closing the `HeliosDeployment`.

gimaker: I'd like to deconstruct `TemporaryJobs` even more, right now it serves two purposes:

 - It glues everything together and is the place where you provide configuration, and
 - It keeps tracks of created jobs and undeploys them when `close()` is called

Why not separate those responsibilities out into separate classes (that are hopefully easier to reason about on their own)? E.g. a separate collection-of-temporary-jobs-that-takes-care-undeploying and a temporary-job-factory.

circle.sh
davidxia: Let's do 1. We've split the tests in a way to minimize test times.

helios-testing/src/main/java/com/spotify/helios/testing/SoloAgentProber.java
davidxia: Is there a better way? A lot of the anonymous functions written for `Polling` do the same.

helios-testing/src/main/java/com/spotify/helios/testing/SoloWatchdogConnector.java
davidxia: Not sure why I did this. You're right.

helios-testing/src/test/java/com/spotify/helios/testing/Utils.java
davidxia: The single method here is only used in one class. I'll just move the method there and delete the class.

davidxia: There's a similar method that uses `CustomTypeSafeMatcher` in `SoloMasterProberTest`. I'll just move that here.

helios-testing/src/test/java/com/spotify/helios/testing/UdpProberTest.java
davidxia: Forgot about this one.

davidxia:
@mattnworb regarding https://github.com/spotify/helios/pull/922#issuecomment-221967892. I'm pretty sure users can still specify `REGISTRAR_HOST_FORMAT` with a `helios.conf` file in their resources directory.

I had to change the SRV record in `HeliosSoloIT` because `HeliosSoloDeployment` is created automatically now by `TemporaryJobs` and I didn't know how to programmatically alter the `helios.conf` file in the resources file. So I'll remove the TODO. But let me know if there's a better way to test this or if I'm mistaken about the preservation of the functionality.

davidxia:
I'll hold off on updating the docs for now. Things are probably going to change. I noted it as a TODO in the PR description above.

mattnworb:
I like the idea of including a documentation update in the PR that makes the change itself, so that it is much harder to forget to update the docs.

davidxia:
Yea, I was thinking the same.

On Tuesday, May 31, 2016, Matt Brown <notifications@github.com> wrote:

> I like the idea of including a documentation update in the PR that makes
> the change itself, so that it is much harder to forget to update the docs.
>
> —
> You are receiving this because you modified the open/close state.
> <https://github.com/spotify/helios/pull/922#issuecomment-222753523>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAdVbQqszX9AEdm7Rd-C4F5XKvMejSr1ks5qHGotgaJpZM4IiRpz>
> .
>


-- 
David Xia
Software Engineer
dxia@spotify.com

helios-testing/src/main/java/com/spotify/helios/testing/SoloAgentProber.java
davidxia: Use awaitility here?

PR #925, opened 2016-05-20
WIP: Use a singleton HSD for all tests during a test run

gimaker:
This PR makes the following changes:

* one `HeliosSoloDeployment` per JVM process per test run
* split `TemporaryJobs` into two classes - one that implements `Closeable` and another that implements JUnit's `Rule`


davidxia:
As of 6f39ad7, not specifying a `HeliosDeployment` for `TemporaryJobs` will start helios-solo:

```java
public class FooTest {

  @Rule
  public TemporaryJobs temporaryJobs = TemporaryJobs.create();

  @Test
  public void test() throws Exception {
    final TemporaryJob memcacheServerJob = temporaryJobs.job()
        .image("spotify/lasertag-memcached:latest")
        .port("memcached", 11211)
        .deploy();
    System.out.println("hi");
  }

  @Test
  public void test2() throws Exception {
    final TemporaryJob memcacheServerJob = temporaryJobs.job()
        .image("spotify/lasertag-memcached:latest")
        .port("memcached", 11211)
        .deploy();
    System.out.println("there");
  }
}
```

Specifying one will use that pre-existing one:

```java
public class FooTest2 {

  @Rule
  public TemporaryJobs temporaryJobs = TemporaryJobs.builder()
      .heliosDeployment(ExistingHeliosDeployment.newBuilder()
                            .domain("shared.cloud.spotify.net")
                            .build()
      ).build();

  @Test
  public void test3() throws Exception {
    final TemporaryJob memcached = temporaryJobs.job()
        .image("spotify/lasertag-memcached:latest")
        .port("memcached", 11211)
        .deploy("heliosagent1.example.com");
    memcached.hosts();

    System.out.println("HOO");

    final Undeployer undeployer = temporaryJobs.undeployer();
    final List<AssertionError> errors = undeployer.undeploy(memcached.job(), memcached.hosts());
    System.out.println(errors);
  }
}
```

davidxia:
What class should do the undeploying of a `TemporaryJob`? It's the `TemporaryJob` itself right now in master, but we said this class was doing too much. In my example above, there's an `Undeployer` you get from `TemporaryJobs`, but that seems ugly as well.

helios-testing/src/main/java/com/spotify/helios/testing/ExistingHeliosDeployment.java
```diff
@@ -0,0 +1,237 @@
...
+  public void cleanup() {
+    jobPrefixFile.delete();
+  }
+
+  @Override
+  public void close() {
+    try {
+      heliosClient.close();
+    } catch (IOException e) {
+      log.warn("HeliosClient did not close cleanly: {}", e);
```
mattnworb: if you pass a throwable as the last argument to a log method, logback (or slf4j, not sure which) interprets it as you wanting to log the stacktrace, so this will look like

```
WARN HeliosClient did not close cleanly: {}
java.util.IOException: blah
  at foo.bar.blah():123
  at ...
```

if you just want to log the message and not the stacktrace do `log.warn("HeliosClient did not close cleanly: {}", e.toString())`

helios-testing/src/main/java/com/spotify/helios/testing/ExistingHeliosDeployment.java
```diff
@@ -0,0 +1,237 @@
...
+          // Set flag indicating if any errors occur
+          if (!errors.isEmpty()) {
+            jobRemovalFailed = true;
+          }
+        }
+
+        // If all jobs were removed successfully, then delete the prefix file. Otherwise,
+        // leave it there so we can try again next time.
+        if (!jobRemovalFailed) {
+          prefixFile.delete();
```
mattnworb: `file.delete()` will return `false` if file cannot be deleted, maybe handle that?

helios-testing/src/main/java/com/spotify/helios/testing/HeliosDeployment.java
```diff
@@ -0,0 +1,49 @@
...
+
+import com.spotify.helios.client.HeliosClient;
+
+import java.net.URI;
+import java.util.Set;
+
+/**
+ * This class represents a Helios cluster, i.e. master, ZooKeeper, and agent nodes.
+ */
+public interface HeliosDeployment extends AutoCloseable {
```
mattnworb: why extend `AutoCloseable` instead of `Closeable`?

helios-testing/src/main/java/com/spotify/helios/testing/HeliosDeploymentResource.java
```diff
@@ -64,8 +65,8 @@ public void before() throws Throwable {
     Polling.awaitUnchecked(30, TimeUnit.SECONDS, new Callable<Boolean>() {
       @Override
       public Boolean call() throws Exception {
-        final HostAndPort hap = deployment.address();
-        final SocketAddress address = new InetSocketAddress(hap.getHostText(), hap.getPort());
+        final URI uri = Iterables.getOnlyElement(deployment.uris());
```
mattnworb: if the Set in `uris()` had more than one elemnent, `Iterables.getOnlyElement()` will throw an exception. Is it really necessary to fail if there is > 1 URI?

mattnworb: and if all implementations of `uri()` today only return a collection with one element, then why model it as a collection at all?

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -144,11 +144,12 @@ private HeliosSoloDeployment(final Builder builder) {
     }
 
     // Running the String host:port through HostAndPort does some validation for us.
-    this.deploymentAddress = HostAndPort.fromString(dockerHost.address() + ":" + heliosPort);
+    final URI uri = URI.create("http://" + dockerHost.address() + ":" + heliosPort);
```
mattnworb: comment above this line is outdated now

helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
```diff
@@ -160,32 +164,35 @@ public void after() {
     } catch (InterruptedException ignore) {
     }
 
-    final List<AssertionError> errors = newArrayList();
+    final List<AssertionError> errors = new ArrayList<>();
+
+    if (existingHeliosDeployment) {
+      for (final TemporaryJob job : jobs) {
+        // Undeploying a job doesn't guaruntee the container will be stopped immediately.
```
mattnworb: typo: guarantee

helios-testing/src/main/java/com/spotify/helios/testing/Undeployer.java
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * Undeploy the job from all specified hosts, and delete the job. Any failures will be ignored,
+   * and we will keep trying each host. A list of errors encountered along the way will be returned
+   * to the caller.
+   * @param job the {@link TemporaryJob} to undeploy and delete
+   * @param hosts the hosts to undeploy from
+   * @return A list of {@link AssertionError}, if any, that occurred.
+   */
+  List<AssertionError> undeploy(Job job, List<String> hosts);
```
mattnworb: why `java.lang.AssertionError`? It is strange for any user-level code to throw/return/construct instances of `Error` rather than `Exception`.

helios-testing/src/test/java/com/spotify/helios/testing/DefaultUndeployerTest.java
```diff
@@ -0,0 +1,131 @@
...
+    //noinspection unchecked
+    when(client.undeploy(jobIdWithNameAndVersion(JOB_ID), anyString())).thenReturn(
+        undeployFuture1, undeployFuture2);
+
+    when(client.deleteJob(jobIdWithNameAndVersion(JOB_ID))).thenReturn(Futures.immediateFuture(
+        new JobDeleteResponse(JobDeleteResponse.Status.OK)));
+
+    final List<AssertionError> errors = new DefaultUndeployer(client).undeploy(JOB, HOSTS);
+
+    assertThat(errors.size(), equalTo(2));
```
mattnworb: minor but I find that `assertThat(errors, hasSize(2))` outputs more helpful info when it fails than this form, as it prints the whole collection and it's elements

helios-testing/src/test/java/com/spotify/helios/testing/DefaultUndeployerTest.java
```diff
@@ -0,0 +1,131 @@
...
+    final List<AssertionError> errors = new DefaultUndeployer(client).undeploy(JOB, HOSTS);
+
+    assertThat(errors.size(), equalTo(1));
+  }
+
+  private static JobId jobIdWithNameAndVersion(final JobId jobId) {
+    return argThat(new ArgumentMatcher<JobId>() {
+      @Override
+      public boolean matches(Object argument) {
+        if (argument instanceof JobId) {
```
mattnworb: you can extend `org.hamcrest.CustomTypeSafeMatcher` instead and it takes care of these instanceof checks for you

helios-testing/src/main/java/com/spotify/helios/testing/HeliosDeployment.java
gimaker: a) you can use try-with-resources, b) you're no longer bound to throwing check exceptions deriving from `IOException`

mattnworb: the first point is true with either interface as `Closeable extends AutoCloseable`

helios-testing/src/main/java/com/spotify/helios/testing/ExistingHeliosDeployment.java
davidxia: This calls into `Files.deleteIfExists()` which according to its javadocs:

```
     * @return  {@code true} if the file was deleted by this method; {@code
     *          false} if the file could not be deleted because it did not
     *          exist
 ```

Does it matter if it's false in this case?

File: helios-testing/src/main/java/com/spotify/helios/testing/HeliosDeployment.java (hunk 3 above)
Comment (davidxia) [inline helios-testing/src/main/java/com/spotify/helios/testing/HeliosDeployment.java]: I'll change to Closeable

File: helios-testing/src/main/java/com/spotify/helios/testing/HeliosDeploymentResource.java (hunk 4 above)
Comment (davidxia) [inline helios-testing/src/main/java/com/spotify/helios/testing/HeliosDeploymentResource.java]: Sorry this code is slightly outdated. Do we want to make the `deployment` attribute a `HeliosDeployment` instance? In that case, there might be more than one URI if it's an `ExistingHeliosDeployment`.

File: helios-testing/src/main/java/com/spotify/helios/testing/Jobs.java [hunk 10]
```diff
@@ -19,27 +19,17 @@
...
-
-import java.util.List;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
-import static java.lang.String.format;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
+// TODO (dxia) Does this class make sense? It's just a collection of static methods.
```
davidxia: Should we rename or move these methods out? If the latter, to where?

helios-testing/src/main/java/com/spotify/helios/testing/Undeployer.java
davidxia: I kept the `AssertionError` from previous code. Not sure why it was originally that. What kind of Exception should we use? Would it make sense to use a type of Exception even though they won't be thrown by helios code?

helios-testing/src/test/java/com/spotify/helios/testing/DefaultUndeployerTest.java
davidxia: Oh nice. My brain isn't working right now. Can you give an example? I couldn't figure out how to extend the class and return a `JobId` required by the calls above.

davidxia: Ah, nvm I figured it out.

davidxia:
@gimaker Want to take a stab at splitting up `TemporaryJobs` into `Rule` and `Closeable` or do we still need more refactoring?

davidxia:
Look at all those beautiful green checkmarks with verified badges!

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -240,11 +240,6 @@ private DockerHost containerDockerHost() {
     return dockerHost;
   }
 
-  @Override
-  public Set<URI> uris() {
```
davidxia: 👍

helios-testing/src/main/java/com/spotify/helios/testing/HeliosDeploymentResource.java
```diff
@@ -61,20 +54,17 @@ public HeliosDeploymentResource(final HeliosSoloDeployment deployment) {
...
       public Boolean call() throws Exception {
-        final URI uri = Iterables.getOnlyElement(deployment.uris());
-        final SocketAddress address = new InetSocketAddress(uri.getHost(), uri.getPort());
-        log.debug("attempting to connect to {}", address);
-
         try {
-          final Socket s = new Socket();
-          s.connect(address, 100);
-          log.info("successfully connected to address {} for {}", address, deployment);
+          client.listMasters().get();
```
davidxia: 👍

davidxia:
@gimaker Your changes LGTM. Shall we merge?

helios-testing/src/main/java/com/spotify/helios/testing/ExistingHeliosDeployment.java
mattnworb: I thought this was a `java.io.File` instance, nevermind

davidxia:
All commits have been cherry-picked to https://github.com/spotify/helios/pull/922

PR #927, opened 2016-05-23
HeliosSoloDeployment: don't use gateway IP in Docker for Mac

mattnworb:
When DOCKER_HOST is a unix socket path, then HeliosSoloDeployment ends up using the IP address of the gateway of the probe container as the address that deployed-containers can be reached at (for example, when TemporaryJob probes the container's port mappings).

This fails for Docker for Mac (and probably other environments as well) where the gateway address is 172.17.0.1 (when using the default bridged network settings) which is not reachable from the host.

This solution of testing explicitly for Docker for Mac feels a bit hacky as it does not address the root question. The root question that HeliosSoloDeployment asks is what address should be used to communciate with ports mapped from the job's container to the host running helios-solo.
In most cases the address of `DOCKER_HOST` is used, and there is an existing special case to use the probe container's gateway when the address of DOCKER_HOST is `localhost`. This special case fails with Docker for Mac and it's `DOCKER_HOST` of `unix:///var/run/docker.sock`, but there are other cases when the gateway *should* be used when DOCKER_HOST is a unix path (judging from the logic in `helios-env` and `helios-up`.

However it does not seem clear what the generalized conditions are for when the probe container's gateway address should be used for `HELIOS_HOST_ADDRESS`, hence this hack.

### DIFF
--- a/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
@@ -203,9 +203,8 @@ public class HeliosSoloDeployment implements HeliosDeployment {
     return "moby".equals(dockerInfo.name());
   }
 
-  private List<String> containerEnv(final Set builderEnv) {
-    final HashSet<String> env = new HashSet<>();
-    env.addAll(builderEnv);
+  private List<String> containerEnv(final Set<String> builderEnv) {
+    final HashSet<String> env = new HashSet<>(builderEnv);
     env.add("DOCKER_HOST=" + containerDockerHost.bindURI().toString());
     if (!isNullOrEmpty(containerDockerHost.dockerCertPath())) {
       env.add("DOCKER_CERT_PATH=/certs");
--- a/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
@@ -157,10 +157,14 @@ public class HeliosSoloDeployment implements HeliosDeployment {
    * be localhost or 127.0.0.1.</p>
    */
   private String determineHeliosHost(final Info dockerInfo) throws HeliosDeploymentException {
+    // note that checkDockerAndGetGateway is intentionally always called even if the return value
+    // is discarded, as it does important checks about the local docker installation
+    final String probeContainerGateway = checkDockerAndGetGateway();
+
     // conditions where the gateway IP address given to a container should be used
     if (dockerHostAddressIsLocalhost() && !isDockerForMac(dockerInfo)) {
       log.info("checking that docker can be reached from within a container");
-      return checkDockerAndGetGateway();
+      return probeContainerGateway;
     }
     // otherwise return the address of the docker host
     return dockerHost.address();
--- a/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
@@ -749,7 +749,7 @@ public class HeliosSoloDeployment implements HeliosDeployment {
      *
      * @return A Helios Solo deployment configured by this Builder.
      */
-    public HeliosDeployment build() {
+    public HeliosSoloDeployment build() {
       this.env = ImmutableSet.copyOf(this.env);
       return new HeliosSoloDeployment(this);
     }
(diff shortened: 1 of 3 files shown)


mattnworb:
@spotify/helios-team 

this could be improved - as mentioned above, the conditions in `determineHeliosHost()` don't feel exhaustive to me. It feels like this check should be generalized, but I am not sure how to express the general condition.

negz:
LGTM

mattnworb:
Tests were failing because some paths in one of the tests ends up calling `DockerHost.fromEnv()` and then the class being tested branches based upon those values, meaning the behavior in CircleCI is different than local. Fixed in ec2adf5.

mattnworb:
please review as this is the last hurdle I know of preventing helios-testing from working with Docker for Mac @gimaker @rohansingh @davidxia

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -146,17 +147,39 @@
...
+   * helios-solo container. This will be passed into the helios-solo container as the HOST_ADDRESS
+   * environment variable and is later used by TemporaryJob to figure out how to reach ports mapped
+   * by the container on the host.
+   * <p>
+   * The returned value is the address of the {@code dockerHost} unless the address is determined to
+   * be localhost or 127.0.0.1.</p>
+   */
+  private String determineHeliosHost(final Info dockerInfo) throws HeliosDeploymentException {
+    // conditions where the gateway IP address given to a container should be used
+    if (dockerHostAddressIsLocalhost() && !isDockerForMac(dockerInfo)) {
```
negz: We don't need to determine the gateway, but could there still be value in running the probe container when using Docker for Mac?

For context, `checkDockerAndGetGateway()` used to be two separate functions with distinct responsibilities - one to test whether we can communicate with Docker from inside the container and another to determine the gateway IP from inside the container. We merged them into the same function because they span up the same Docker image twice for different reasons.

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -146,17 +147,39 @@
...
   }
 
+  /**
+   * Determine what address to use when attempting to communicate with containers deployed via the
+   * helios-solo container. This will be passed into the helios-solo container as the HOST_ADDRESS
+   * environment variable and is later used by TemporaryJob to figure out how to reach ports mapped
+   * by the container on the host.
+   * <p>
+   * The returned value is the address of the {@code dockerHost} unless the address is determined to
+   * be localhost or 127.0.0.1.</p>
```
davidxia: Add a note about Docker for Mac too?

davidxia:
👍

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
mattnworb: Ah good point. I interpreted the fact that the return value from the method being ignored in one of these branches as meaning that the entire method call wasn't needed. I'll update this.

mattnworb: tackled in 393ee89

mattnworb:
ugh had to rebase, @davidxia @negz please take another look

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -176,18 +203,12 @@ private boolean isBoot2Docker(final Info dockerInfo) {
...
 
-  private Info dockerInfo() {
-    try {
-      return this.dockerClient.info();
-    } catch (DockerException | InterruptedException e) {
-      // There's not a lot we can do if Docker is unreachable.
-      throw Throwables.propagate(e);
-    }
+  private boolean isDockerForMac(final Info dockerInfo) {
+    return "moby".equals(dockerInfo.name());
```
davidxia: Will this always be the case?

mattnworb: potentially not but it is the best identifier I found for now. As I mentioned in the PR summary, I feel like there should be a broader condition to test for that I am not able to see the answer to ("what IP to use to reach a mapped port?") but I wanted to at least improve things for these users.

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -146,17 +147,43 @@
...
+   * Determine what address to use when attempting to communicate with containers deployed via the
+   * helios-solo container. This will be passed into the helios-solo container as the HOST_ADDRESS
+   * environment variable and is later used by TemporaryJob to figure out how to reach ports mapped
+   * by the container on the host.
+   * <p>
+   * The returned value is the address of the {@code dockerHost} unless the address is determined to
+   * be localhost or 127.0.0.1.</p>
+   */
+  private String determineHeliosHost(final Info dockerInfo) throws HeliosDeploymentException {
+    // note that checkDockerAndGetGateway is intentionally always called even if the return value
```
davidxia: Ah good note.

davidxia:
👍

PR #928, opened 2016-05-24
Upgrade docker-client dep from 3.6.7 to 5.0.1

davidxia:


helios-services/src/main/java/com/spotify/helios/agent/HealthCheckerFactory.java
```diff
@@ -23,7 +23,7 @@
 import com.google.common.collect.Iterables;
 
 import com.spotify.docker.client.DockerClient;
-import com.spotify.docker.client.DockerException;
+import com.spotify.docker.client.exceptions.DockerException;
```
mattnworb: should make sure these are sorted, simple search and replace won't work

mattnworb:
lgtm if you fix the import order

PR #929, opened 2016-05-24
Use getContainerInfo() to get container state during healthchecking

negz:
Reviewers: @mattnworb 
CC: @spotify/helios-devs 

Just squashing a few TODOs.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/agent/TaskRunner.java
+++ b/helios-services/src/main/java/com/spotify/helios/agent/TaskRunner.java
@@ -176,12 +176,18 @@ class TaskRunner extends InterruptingExecutionThreadService {
             .build().newScheduler();
 
         while (!healthChecker.get().check(containerId)) {
-          // TODO(negz): Why aren't we using getContainerInfo() here?
-          final ContainerState state = docker.inspectContainer(containerId).state();
+          final ContainerState state = getContainerState(containerId);
+          if (state == null) {
+            final String err = "container " + containerId + " was not found during health "
+                + "checking, or has no State object";
+            log.warn(err);
+            throw new RuntimeException(err);
+          }
           if (!state.running()) {
-            log.warn("container exited during health checking: {}: {}: {}",
-                     config, containerId, state.exitCode());
-            throw new RuntimeException("container exited during health checking");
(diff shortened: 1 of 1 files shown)


helios-services/src/main/java/com/spotify/helios/agent/TaskRunner.java
```diff
@@ -176,8 +176,12 @@ private int run0() throws InterruptedException, DockerException {
             .build().newScheduler();
 
         while (!healthChecker.get().check(containerId)) {
-          // TODO(negz): Why aren't we using getContainerInfo() here?
-          final ContainerState state = docker.inspectContainer(containerId).state();
+          final ContainerState state = getContainerState(containerId);
+          if (state == null) {
+            log.warn("container state was null during health checking: {}: {}",
+                     config, containerId);
+            throw new RuntimeException("container state was null during health checking");
```
mattnworb: a better error message for both the log line and exception would be something like `"container <containerId> was not found during health checking, or has no State object"` - as-is the ID of the container is not included and it's not totally clear what it means for the state to be null

mattnworb:
👍 for the intent though

negz:
@mattnworb PTAL

mattnworb:
👍 🙌

helios-services/src/main/java/com/spotify/helios/agent/TaskRunner.java
```diff
@@ -176,12 +176,18 @@ private int run0() throws InterruptedException, DockerException {
...
+            final String err = "container " + containerId + "was not found during health checking, "
+                + "or has no State object";
+            log.warn(err);
+            throw new RuntimeException(err);
+          }
           if (!state.running()) {
-            log.warn("container exited during health checking: {}: {}: {}",
-                     config, containerId, state.exitCode());
-            throw new RuntimeException("container exited during health checking");
+            final String err = "container " + containerId + "exited during health checking. "
```
daenney: This needs an additional space: `" exited during"...` otherwise you'll get the `containerId` glued to it.

negz: Thanks! I double checked for spaces but apparently needed to triple check. :(

PR #930, opened 2016-05-25
Use CustomTypeSafeMatcher in tests

davidxia:
instead of `argThat(new ArgumentMatcher<T>() {...})` since
`CustomTypeSafeMatcher` does `instanceof` checks.


davidxia:
@mattnworb

mattnworb:
👍

PR #931, opened 2016-05-25
default to not removing helios-solo containers after test

mattnworb:
The default value of removing the container after it's use makes it much
harder to troubleshoot issues within the helios-solo container itself.


mattnworb:
@negz @gimaker @davidxia

davidxia:
👍  Were we just afraid of cruft buildup originally?

negz:
@davidxia I forget, but it could have been me erroneously assuming that we had to rm the containers in order to ensure we pulled the latest image on the next run.

PR #932, opened 2016-05-26
Don't block rolling updates when new hosts fail to update.

negz:
Reviewer: @mattnworb 
CC: @spotify/helios-devs 

Currently it's possible that:
1. A user creates a deployment group and successfully performs a rolling update
2. New hosts are added to said group, triggering another rolling update
3. One or more of those hosts fails to update, and the group is marked FAILED
4. More hosts are added, but we don't trigger a rolling update because we're wary the last one failed.

This change prevents the scenario outlined at step 4 above, while ensuring we don't trigger an automatic rolling update for new hosts if the most recent manual rolling update failed.

This is implemented by tracking the reason (manual, or because of changed hosts) of the last rolling update in the DeploymentGroup object. This gives us an extra data point when deciding whether to automatically rolling update upon the addition of new hosts to a deployment group.


negz:
The new tests are super dirty copy pasta. I'm open to suggestions as to how they might be tightened up a bit.

helios-client/src/main/java/com/spotify/helios/common/descriptors/DeploymentGroup.java
```diff
@@ -199,6 +222,9 @@ public Builder setName(final String name) {
       return this;
     }
 
+    // TODO(negz): Annotate this as @JsonProperty("job")?
+    // AFAICT we're currently serialising this with the key "jobId", despite annotating it as
+    // @JsonProperty("job") in the constructor and javadoc.
```
mattnworb: It definitely serializes as `"jobId"`. 

I made a test for this and it seems like Jackson knows to modify the `jobId` field of the class after calling the constructor (through reflection somehow). So in the end, the data in the object is as expected if you translate it to json and back. 

Changing the annotations might cause incompatibilities with already stored versions 😞

mattnworb: also I think this comment is at the wrong place in the file, not in the middle of the builder class?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -433,6 +432,28 @@ public void removeDeploymentGroup(final String name) throws DeploymentGroupDoesN
...
+      return (!status.getState().equals(FAILED));
+    }
+
+    if (group.getRollingUpdateReason().equals(HOSTS_CHANGED) && status.getState().equals(FAILED)) {
+      // The last rolling update failed, but it was triggered by hosts changing. Try again.
+      return true;
+    }
+
+    // The last rolling update didn't fail, so there's no reason to expect this one will.
+    return (!status.getState().equals(FAILED));
```
mattnworb: the outer parenthesis here are unnecessary

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -433,6 +432,28 @@ public void removeDeploymentGroup(final String name) throws DeploymentGroupDoesN
...
+      // We're in an unknown state. Go hog wild.
+      return true;
+    }
+
+    if (group.getRollingUpdateReason() == null) {
+      // The last rolling update didn't fail, so there's no reason to expect this one will.
+      return (!status.getState().equals(FAILED));
+    }
+
+    if (group.getRollingUpdateReason().equals(HOSTS_CHANGED) && status.getState().equals(FAILED)) {
```
mattnworb: minor: you can compare enums with `==`

mattnworb:
LGTM, I think this approach will definitely work and solve the problem.

But would it be simpler conceptually if step 3 in the outline above simply didn't cause the group/state to be marked as FAILED?

i.e., when doing a rollout because hosts changed, don't change the `DeploymentGroupStatus.state` to `FAILED` if the reason is `HOSTS_CHANGED`

helios-client/src/main/java/com/spotify/helios/common/descriptors/DeploymentGroup.java
negz: Oh, right. Wrong `getJobId`. Yeah I was concerned about fixing it due to backwards compatibility. I could remove the TODO and just update the javadoc and leave a comment near the constructor annotation?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: Is that preferred? I knew you could, but thought for some reason I should default to `equals`.

helios-client/src/main/java/com/spotify/helios/common/descriptors/DeploymentGroup.java
mattnworb: 👍

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
mattnworb: it might just be a point of personal preference, but I find it's less characters to type and read 😎

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -459,14 +480,21 @@ public void updateDeploymentGroupHosts(final String groupName, final List<String
...
-        final DeploymentGroup deploymentGroup = getDeploymentGroup(groupName);
+        DeploymentGroup deploymentGroup = getDeploymentGroup(groupName);
         ImmutableList<Map<String, Object>> events = ImmutableList.of();
 
         if (deploymentGroup.getJobId() != null) {
-          final DeploymentGroupStatus deploymentGroupStatus = getDeploymentGroupStatus(groupName);
-          if (deploymentGroupStatus == null || deploymentGroupStatus.getState() != FAILED) {
-            final RollingUpdateOp op =
-                getInitRollingUpdateOps(deploymentGroup, hosts, HOSTS_CHANGED, client);
+          if (updateOnHostChange(deploymentGroup)) {
```
gimaker: Maybe only trigger roling-updates if the current state is `FAILED` or `SUCCEEDED` (not `ROLLING_OUT`)?

If I understand correctly, you can now time things to get perhaps unexpected results:
 - trigger a manual rolling-update of a bad, broken job
 - immediately after, before the rolling-update fails, a new hosts is added
 - a roll out (with reason=HOSTS_CHANGED) is triggered and fails
 - more hosts are added...
 - this triggers a rolling-update because the last reason had reason=HOSTS_CHANGED

gimaker: (This probably doesn't matter in practice)

gimaker: This also changes behaviour slightly in that a DG can go from `FAILED` to `ACTIVE` without user-intervention.

negz:
@mattnworb This way feels more intuitive to me personally.

That said, I suppose it depends whether you frame it as "the deployment group is failed" or "the rolling update you performed on the deployment group failed". The lack of separation between rolling updates and deployment groups really muddies the waters for me.

The alternatives:
* Break up the Deployment Group states per your proposal in Trello. We'd break `FAILED` into two new states, one being "this deployment group failed during a host update, so it's kind of unstable but it's cool to automatically deploy again" with the other being "this deployment group failed during a manual update, so it's unstable and we shouldn't allow any automatic updates until a human intervenes". I veered away from that proposal because I found it difficult to succinctly capture the above concepts in an `enum`. (`HOSTS_CHANGED_FAILED` and `MANUAL_FAILED`?)
* Set the `DeploymentGroupStatus` to `DONE` rather than `FAILED` at https://github.com/spotify/helios/blob/master/helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdateOpFactory.java#L164 if the `RollingUpdateReason` was hosts changed.

I think either of the above implementations would be best served by modeling the `RollingUpdateReason` in the `DeploymentGroup` per this change, so they'd share a bit of code anyhow.

@spotify/helios-devs  What do you guys think?

gimaker:
> But would it be simpler conceptually if step 3 in the outline above simply didn't cause the group/state to be marked as FAILED?

I agree with this (but I don't know how hairy it would be to implement).

I also think we should *not* trigger rolling-updates when hosts change *if there's already a manually triggered rolling-update in progress* (since it, depending on the exact implementation, leads to really subtle behaviour).

negz:
> I agree with this (but I don't know how hairy it would be to implement so take my opinion lightly).

@gimaker @mattnworb I believe it would be pretty trivial to implement, as it would share a bunch of implementation with this PR.

I prefer the implementation in this PR because, while both options are flawed, it seems better for a rolling update that leaves some hosts in a broken state to (or a deployment group with some hosts in a broken state) to indicate that by having state `FAILED`. Even if the next rolling update would be allowed. My understanding is that you're proposing we leave the deployment group in state `DONE` if a `HOSTS_CHANGED` rolling update fails.

negz:
> I also think we should not trigger rolling-updates when hosts change if there's already a manually triggered rolling-update in progress (since it, depending on the exact implementation, leads to really subtle behaviour).

@gimaker That seems reasonable, but it's tangential to the intent of this PR right? i.e. We allow new `HOSTS_CHANGED` rolling updates to start while `MANUAL` rolling update is `ROLLING_OUT` at the moment.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: Okay, changed. :)

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java

helios-client/src/main/java/com/spotify/helios/common/descriptors/DeploymentGroup.java
negz: Done.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: > This PR also changes behaviour slightly in that a DG can go from FAILED to ACTIVE without user-intervention.

I thought that was kind of the point. :)

mattnworb:
> I also think we should not trigger rolling-updates when hosts change if there's already a manually triggered rolling-update in progress (since it, depending on the exact implementation, leads to really subtle behaviour).

+1, in a separate PR

Although it would raise the question of how and when those new hosts would get the job version being rolled out, assuming that the initial batch of operations would include the list of hosts prior to the new hosts addition and therefore ignore the new host (not sure if this is accurate?).

mattnworb:
> That said, I suppose it depends whether you frame it as "the deployment group is failed" or "the rolling update you performed on the deployment group failed". 

Yep I think this accurately captures the different perspectives. In my mind I see it as the former since we have a "inspect deployment group" command which reports the state (which can be failed). But I agree the latter phrasing is the ideal way to look at things and for it to eventually be modeled as.

negz:
> +1, in a separate PR

@mattnworb I just noticed @gimaker's comment at https://github.com/spotify/helios/pull/932#discussion-diff-64818492R487. I think the behaviour described was actually introduced in this PR so maybe it is actually worth doing here?

Preventing `HOSTS_CHANGED` rolling updates when we're `ROLLING_OUT` is easy, but I have no idea how we'd get those hosts added later. Just hope the races were rare and ask people to do a `MANUAL` rolling update if they hit that edge case?

gimaker:
> That seems reasonable, but it's tangential to the intent of this PR right? i.e. We allow new HOSTS_CHANGED rolling updates to start while MANUAL rolling update is ROLLING_OUT at the moment.

Tangential, yes -- but I think it'd avoid certain odd cases that are otherwise present with this PR.

negz:
> I have no idea how we'd get those hosts added later. Just hope the races were rare and ask people to do a MANUAL rolling update if they hit that edge case?

@gimaker just pointed out that `updateDeploymentGroupHosts()` is called by a reactor once per second, so we could just return early without saving the new hosts if a rolling update was in progress and it would try again.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -433,11 +433,72 @@ public void removeDeploymentGroup(final String name) throws DeploymentGroupDoesN
...
+  }
+
   @Override
   public void updateDeploymentGroupHosts(final String groupName, final List<String> hosts)
       throws DeploymentGroupDoesNotExistException {
     log.debug("updating deployment-group hosts: name={}", groupName);
     final ZooKeeperClient client = provider.get("updateDeploymentGroupHosts");
+    final DeploymentGroupStatus status = getDeploymentGroupStatus(groupName);
+    if (!allowHostChange(status)) {
+      return;
```
gimaker: nitpick: maybe add a comment explaining that this will be re-run and the hosts updated on the next iteration.

negz: Do you think it's really worth explaining here when it's also explained in the `allowHostChange` javadoc? I considered it but it seemed like it might be excessive.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -459,14 +520,21 @@ public void updateDeploymentGroupHosts(final String groupName, final List<String
...
-        final DeploymentGroup deploymentGroup = getDeploymentGroup(groupName);
+        DeploymentGroup deploymentGroup = getDeploymentGroup(groupName);
         ImmutableList<Map<String, Object>> events = ImmutableList.of();
 
         if (deploymentGroup.getJobId() != null) {
-          final DeploymentGroupStatus deploymentGroupStatus = getDeploymentGroupStatus(groupName);
-          if (deploymentGroupStatus == null || deploymentGroupStatus.getState() != FAILED) {
-            final RollingUpdateOp op =
-                getInitRollingUpdateOps(deploymentGroup, hosts, HOSTS_CHANGED, client);
+          if (updateOnHostChange(deploymentGroup, status)) {
```
gimaker: I think there's still a race condition here, e.g.:
 * Get DG and DG status [in the context of this code]
 * Manual rolling-update triggered [concurrently with this code running]
 * `updateOnHostChange()` is run, and returns true
 * --> Current [manual] rolling-update is overwritten

You could add a ZK `check()` in the transaction to ensure that the current version of the DG config and/or status node has not changed if you want to avoid the race.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
gimaker: I don't have a strong opinion.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: Good point. I've taken a pass in https://github.com/spotify/helios/pull/932/commits/425a04f9e3749545af2d1a4281d2db0613d94a68. Does that look about right?

I feel like `updateDeploymentGroupHosts` is getting a bit bloated, so I'm open to refactoring if anyone has any ideas.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -433,40 +433,116 @@ public void removeDeploymentGroup(final String name) throws DeploymentGroupDoesN
...
+        final Node chn = client.getNode(Paths.statusDeploymentGroupHosts(groupName));
+        curHostsVersion = Optional.of(chn.getStat().getVersion());
+        curHosts = Json.read(chn.getBytes(), new TypeReference<List<String>>() {});
       } catch (JsonMappingException e) {
         curHosts = Collections.emptyList();
-      } catch (NoNodeException e) {
-        // DG was deleted -- abort.
+      }
+
+      final Node sn = client.getNode(Paths.statusDeploymentGroup(groupName));
```
gimaker: I believe this will throw a `NoNodeException` if the nodes does not exist. You probably want to catch and act on that. The node does not exist --> `status` will be `null` below.

gimaker: I think that checking only the config node is enough to avoid the race condition.

negz: Won't the NoNodeException get caught waaaay down at line 558?

gimaker: Yes... but is that what you want? `allowHostChange()` returns true if `status` is null, allowing a roll out to be started if there is not known status. With the exception thrown, `status` can never be null --> roll outs will never be triggered when the status is unknown.

negz: Oh, gotcha.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: I'll leave it as is.

PR #934, opened 2016-05-31
fix helios-solo scripts for Docker for Mac

mattnworb:
When `DOCKER_HOST` is not set, the `helios-env` script assumes that if the local OS is Linux then the Unix socket at `/var/run/docker.sock` can be used to communicate with the Docker daemon.

Docker for Mac also uses this socket and path and does not need `DOCKER_HOST` to be set to work, so add a test to see if a bare `docker info` command exits with a success and use the same Unix socket path when it exits successfully.

Also fixed some other low-hanging fruit in these scripts, specifically the mixing of tabs and spaces for indentation 😎

### DIFF
--- a/solo/docker/start.sh
+++ b/solo/docker/start.sh
@@ -12,11 +12,11 @@ SKYDNS_PATH=$(echo $HELIOS_NAME|python -c "import sys;h=sys.stdin.read().strip()
 
 # Write skydns configuration and retry for 30 seconds until successful
 for i in {1..30}; do
-	if curl --retry 30 -XPUT http://127.0.0.1:4001/v2/keys/skydns/config \
-		-d value="{\"dns_addr\":\"0.0.0.0:5353\", \"ttl\":3600, \"nameservers\": $NAMESERVERS, \"domain\":\"local.\"}"; then
-		break
-	fi
-	sleep 1
+  if curl --retry 30 -XPUT http://127.0.0.1:4001/v2/keys/skydns/config \
+    -d value="{\"dns_addr\":\"0.0.0.0:5353\", \"ttl\":3600, \"nameservers\": $NAMESERVERS, \"domain\":\"local.\"}"; then
+    break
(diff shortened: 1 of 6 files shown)


mattnworb:
@spotify/helios-team 

probably easier to review by ignoring whitespace with https://github.com/spotify/helios/pull/934/files?w=1

davidxia:
:+1:

negz:
LGTM

(Though I'm still not a fan of maintaining these scripts *and* HSD.)

mattnworb:
@negz I think there is still a lot of value in having the standalone `helios-solo` homebrew and Debian packages. They can be used outside of the Maven-build-and-test context for things like running jobs locally during development or to do one-off tests of how helios or your jobs would behave or otherwise test things without involving a real Helios cluster. IMHO the packages or the scripts around starting/creating/stopping/using the helios-solo-container should not go anywhere.

negz:
I won't argue that they don't add value - my main concern is around the cognitive and (potential) support burden of maintaining two similar yet different implementations of the same concept. I wonder whether the value they add is worth that.

davidxia:
It seems worth it to me for the reasons Matt pointed out. Any reduction in
duplicate logic if possible would be good though.

On Monday, June 6, 2016, Nic Cope <notifications@github.com> wrote:

> I won't argue that they don't add value - my main concern is around the
> cognitive and (potential) support burden of maintaining two similar yet
> different implementations of the same concept. I wonder whether the value
> they add is worth that.
>
> —
> You are receiving this because you commented.
> <https://github.com/spotify/helios/pull/934#issuecomment-223960123>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAdVbWzM8xOQVJzUNLne3ovM6Ql_6fGCks5qJCHygaJpZM4IqspS>
> .
>


-- 
David Xia
Software Engineer
dxia@spotify.com

PR #937, opened 2016-05-31
always update status to STOPPING when stopping a job

mattnworb:
Supervisor stores a status update of STOPPED after the container has
been shut down. For jobs with grace periods, it always stores a status
update of STOPPING prior to unregistering the job from service discovery
and sleeping before the container is stopped.

It would make more sense to unconditionally store the status update of
STOPPING prior to the container stop request and the STOPPED state,
since stopping the container may take a number of seconds for the
container to clean itself up or otherwise respond to SIGTERM. This would
lead to more consistent output from the `helios history` command.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/agent/Supervisor.java
+++ b/helios-services/src/main/java/com/spotify/helios/agent/Supervisor.java
@@ -346,12 +346,13 @@ public class Supervisor {
         return;
       }
 
+      statusUpdater.setState(STOPPING);
+      statusUpdater.update();
+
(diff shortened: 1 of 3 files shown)


davidxia:
👍 
💯 on refactoring the tests

mattnworb:
whew tests are ✅  now, after I accidentally introduced an off-by-one bug in my use of an iterator.

davidxia:
Good catch. Lgtm

negz:
LGTM

PR #942, opened 2016-07-06
Add JobHistoryReaper

davidxia:
This reaper, disabled by default, runs in the master and will
remove job histories whose corresponding jobs no longer exist.
It runs once a day when enabled.

This reaper is meant to address the scenario where jobs are deleted but
their histories are left behind in ZooKeeper. See JobHistoryReaper's
JavaDocs for more details.


davidxia:
cc @spotify/helios-team

mattnworb:
👍

PR #944, opened 2016-07-07
HeliosSoloDeployment: use external host address for HELIOS_HOST_ADDRESS

mattnworb:
.. when local environment is Docker for Mac.

This resolves a conflict introduced in #927 where the
HELIOS_HOST_ADDRESS passed into the helios-solo container ends up as
`localhost`.

This value is then used when a TemporaryJob is deployed and
`.address(port)` is called, because TemporaryJob queries the environment
variables of the helios agent.

This causes an issue if one tries to pass this "public address of the
TemporaryJob" value into another TemporaryJob instance (for example, to
tell one TemporaryJob how to reach another), because `localhost` in the
container will route differently than it does on the host.

The change in #927 originally intended to resolve an issue where the
`probeContainerGateway` value was being used, which in DFM has a value
like `172.17.0.1` which is not routable on the host. This caused a
problem if the user tried to use the `TemporaryJob.address(port)` value
_on the host_ to communicate with the container.

The value returned by `address()` needs to be usable both outside the
container and inside a container, so the fix here is to use the public
address of localhost _on the host_.

### DIFF
--- a/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
@@ -55,6 +55,8 @@ import org.jetbrains.annotations.NotNull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.net.InetAddress;
+import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


mattnworb:
@spotify/helios-team @lndbrg

lndbrg:
😄  👍 :shipit:

mattnworb:
@rohansingh @negz @gimaker please review

gimaker:
:+1:

mattnworb:
@lndbrg this should be available in helios-testing:0.8.804 now.

PR #945, opened 2016-07-07
Increase log level for dead agent and job history reapers

davidxia:
I want to know if the reapers have started.


davidxia:
@negz

PR #946, opened 2016-07-07

mattnworb:
user's project.

### DIFF
--- a/helios-client/pom.xml
+++ b/helios-client/pom.xml
@@ -16,10 +16,6 @@
   </properties>
 
   <dependencies>
-    <dependency>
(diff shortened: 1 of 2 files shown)


davidxia:
I believe @drewcsillag authored this, but I don't think we use it.

mattnworb:
I am ambivalent about leaving it in the other modules but I want to avoid dragging other annotation processors into anyone using helios-client.

davidxia:
@mattnworb We need to explicitly depend on jsr311-api in helios-client if we remove docgenerator: https://github.com/spotify/helios/pull/947

PR #947, opened 2016-07-07

davidxia:
user's project.

### DIFF
--- a/helios-tools/pom.xml
+++ b/helios-tools/pom.xml
@@ -58,6 +58,10 @@
       <groupId>com.typesafe</groupId>
       <artifactId>config</artifactId>
     </dependency>
+    <dependency>
+      <groupId>javax.ws.rs</groupId>
+      <artifactId>javax.ws.rs-api</artifactId>
+    </dependency>
 
     <!--test deps-->
     <dependency>
--- a/helios-client/pom.xml
+++ b/helios-client/pom.xml
@@ -16,10 +16,6 @@
   </properties>
 
   <dependencies>
-    <dependency>
-      <groupId>com.spotify.docgenerator</groupId>
-      <artifactId>scanner</artifactId>
-    </dependency>
     <!--compile deps-->
     <dependency>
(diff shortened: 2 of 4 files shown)


helios-tools/pom.xml
```diff
@@ -58,6 +58,10 @@
       <groupId>com.typesafe</groupId>
       <artifactId>config</artifactId>
     </dependency>
+    <dependency>
+      <groupId>javax.ws.rs</groupId>
+      <artifactId>jsr311-api</artifactId>
```
mattnworb: I could be wrong, but if this dependency is for things like the `@GET` annotation then I think the better dependency is `javax.xml.ws:jaxws-api`

mattnworb: or `javax.ws.rs:javax.ws.rs-api` 

there seems to be 1000 variations of that API jar

davidxia: I think you're right. I'll replace it.

On Fri, Jul 8, 2016 at 9:24 AM, Matt Brown <notifications@github.com> wrote:

> In helios-tools/pom.xml
> <https://github.com/spotify/helios/pull/947#discussion_r70072613>:
>
> > @@ -58,6 +58,10 @@
> >        <groupId>com.typesafe</groupId>
> >        <artifactId>config</artifactId>
> >      </dependency>
> > +    <dependency>
> > +      <groupId>javax.ws.rs</groupId>
> > +      <artifactId>jsr311-api</artifactId>
>
> I could be wrong, but if this dependency is for things like the @GET
> annotation then I think the better dependency is javax.xml.ws:jaxws-api
>
> —
> <https://github.com/spotify/helios/pull/947/files/b2aee7e1503abf81db1f7041984354ed2169eeea#r70072613>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAdVbco5-cBGJvobMwLW097igY1pRh7wks5qTk-rgaJpZM4JHdpM>
> .
>



-- 
David Xia
Software Engineer
dxia@spotify.com

Issue #462, opened 2015-04-01
Authentication against private registry.hub.docker.com repositories


nexus49:
Hi there, 
I have been searching for a way to provide auth credentials to pull an image from a private repository. (not necessarily a private registry)

Did i miss that this is somehow possible? I saw that the docker-client is able to provide credentials during pull, but did not see that being used as part of helios. 

Cheers,
Bastian

davidxia:
@nexus49 Our [docker-client](https://github.com/spotify/docker-client) supports private registries (see [docs here](https://github.com/spotify/docker-client/pull/152)), but helios isn't setup to use that functionality yet.

We haven't built this yet bc we haven't needed it and we're quite busy with other things at the moment. But we're open to a PR. The place to change is [here](https://github.com/spotify/helios/blob/master/helios-services/src/main/java/com/spotify/helios/agent/AgentService.java#L208). `dockerClient` needs to be instantiated with the private registry's auth credentials.

lvidarte:
Hi,
I'm trying helios and the agents can't pull images from my private registry.
Is there any advance on this?
Thanks.

    $> helios status
    JOB ID             HOST        GOAL     STATE                     CONTAINER ID    PORTS              
    api:0.6:ab1967d    staging.    START    FAILED (IMAGE_MISSING)                    http=80:9001

PR #948, opened 2016-07-11
New argument "--docker-auth" [#462]

elifarley:
This change is [Reviewable](https://reviewable.io/reviews/spotify/helios/948)


elifarley:
This PR depends on changes introduced in spotify/docker-client#473.

mattnworb:
In the currently-used version of docker-client, support for authentication via docker's config is automatically enabled, which would have the same effect as this PR. Thanks for the contribution though, and sorry for the long delay time in responding.

PR #949, opened 2016-07-13
When a DG fails due to port allocation issues, hint what the problem …

gimaker:
…might be

This commit adds the note in the parentheses below:

  "Allocation of port <port> for job <dg-job> collides with job <other-job>
  on host <host> (the conflicting job was deployed manually or by a
  different deployment group)"


gimaker:
@mattnworb @negz @rohansingh @davidxia

mattnworb:
👍

PR #950, opened 2016-07-13
Serialize Docker pulls for versions with known issues

rohansingh:
Docker versions starting with 1.6.0 have issues handling concurrent pulls,
especially of images that share base layers. The issue is fixed in 1.9+.

Since concurrent pulls can leave Docker in a state where containers won't
run until the daemon is restarted, we want to avoid them entirely for these
older versions.


rohansingh:
@spotify/helios-team

gimaker:
:+1:

mattnworb:
nice 👍

PR #951, opened 2016-07-19
Update docs for deployment group host selectors

davidxia:
Cream peeps would find this useful.

### DIFF
--- a/docs/deployment_groups.md
+++ b/docs/deployment_groups.md
@@ -11,15 +11,23 @@ motivation.
 
 ### Create
 
-Create a deployment group by specifying the [host labels][1] to filter on. The command below will
-create a deployment group whose hosts match all the specified labels.
+A host selector is an expression of the form [label] [operator] [operand].
+Host selectors are used by deployment groups to choose which hosts to act on. The command below will
+create a deployment group whose hosts match all the specified host selectors.
 
     $ helios create-deployment-group foo-group foo=bar baz=qux
 
     Creating deployment group: {"labels":{"baz":"qux","foo":"bar"},"name":"foo-group"}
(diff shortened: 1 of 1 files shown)


davidxia:
@spotify/helios-team

docs/deployment_groups.md
```diff
@@ -11,15 +11,23 @@ motivation.
...
     Creating deployment group: {"labels":{"baz":"qux","foo":"bar"},"name":"foo-group"}
     {"status":"CREATED"}
 
-Creating a deployment with the same name and labels as an existing one will succeed with a
+Here's a list of the available types of host selectors:
+
+  * foo = bar
+  * foo != bar
+  * foo in (bar, qux)
+  * foo notin (bar, qux)
```
mattnworb: This should probably clarify that the expression in `(bar, quz)` is not limited to two values

mattnworb: It might be cleared to just list the operators rather than the left-hand and right-hand side

- `=`
- `!=`
- `in (..)`
- `notin (..)`

docs/deployment_groups.md
```diff
@@ -11,15 +11,23 @@ motivation.
 
 ### Create
 
-Create a deployment group by specifying the [host labels][1] to filter on. The command below will
-create a deployment group whose hosts match all the specified labels.
+A host selector is an expression of the form <label-pattern> <operand> <operand pattern>.
```
gimaker: What's up with all the "patterns"?

The left hand side is always a label, no "patterns" allowed. The right hand side is always an operand.

davidxia:
Updated

mattnworb:
👍🏻

PR #955, opened 2016-07-24
Support all host selector operators in host cmd

davidxia:
We previously only supported the "=" host selector operator in
`helios hosts --labels`. This change supports using "!=", "in",
and "notin" operators.

### DIFF
--- a/helios-system-tests/src/main/java/com/spotify/helios/system/LabelTest.java
+++ b/helios-system-tests/src/main/java/com/spotify/helios/system/LabelTest.java
@@ -24,83 +24,108 @@ import com.spotify.helios.Polling;
 import com.spotify.helios.common.Json;
 import com.spotify.helios.common.descriptors.HostStatus;
 
+import com.google.common.collect.ImmutableSet;
+
 import org.junit.Test;
 
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.Callable;
 
-import static com.spotify.helios.common.descriptors.HostStatus.Status.UP;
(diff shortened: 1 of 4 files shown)


davidxia:
@spotify/helios-team

helios-system-tests/src/main/java/com/spotify/helios/system/LabelTest.java
```diff
@@ -24,83 +24,122 @@
...
+    assertThat(notInFooBarHosts, not(containsStrings(ImmutableSet.of(testHost1, testHost2))));
+
+    final String nonHosts = cli("hosts", "--labels", "role=doesnt_exist");
+    assertThat(nonHosts, not(containsStrings(ImmutableSet.of(testHost1, testHost2))));
+
+    final String nonHosts2 = cli("hosts", "--labels", "role=doesnt_exist", "xyz=123");
+    assertThat(nonHosts2, not(containsStrings(ImmutableSet.of(testHost1, testHost2))));
+  }
+
+  private CustomTypeSafeMatcher<String> containsStrings(final Set<String> substrings) {
```
mattnworb: i think this could also be something like `allOf(containsString(".."), ..)` if you knew the set of strings at build-time

mattnworb:
👍 

it would be really sweet if we could do this filtering server-side (we have a need for something like this in our new project)

davidxia:
@mattnworb Agreed. Lots of CLI logic could be moved server-side to simplify the client and reduce wasted bytes sent over the network.

PR #956, opened 2016-07-25
Default to not loading JDWP agent for debugging

davidxia:
Only laod when running in the helios project as it's probably
only used in that case.


davidxia:
@mattnworb 
cc @spotify/helios-team

mattnworb:
👍

PR #957, opened 2016-07-26
filter host list by selectors/pattern on server-side

mattnworb:
This PR moves the logic that currently only exists within the CLI for filtering the list of hosts by "name pattern" or list of labels into the server/master instead.

This allows users of HeliosClient to be able to list hosts, filtering by the pattern and/or label list, without duplicating the logic within the CLI.

Each of the three components is updated here:
- server-side filtering within the `/hosts/` endpoint
- HeliosClient support for these filters in `listHosts(..)` method overloads
- HostListCommand in the CLI will use the new HeliosClient methods

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
+++ b/helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
@@ -17,26 +17,25 @@
 
 package com.spotify.helios.client;
 
-import com.google.common.base.Function;
-import com.google.common.base.Optional;
-import com.google.common.base.Supplier;
-import com.google.common.base.Suppliers;
-import com.google.common.base.Throwables;
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Lists;
-import com.google.common.collect.Maps;
-import com.google.common.escape.Escaper;
-import com.google.common.net.UrlEscapers;
-import com.google.common.util.concurrent.AsyncFunction;
-import com.google.common.util.concurrent.FutureFallback;
-import com.google.common.util.concurrent.ListenableFuture;
-import com.google.common.util.concurrent.ListeningScheduledExecutorService;
-import com.google.common.util.concurrent.MoreExecutors;
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Strings.isNullOrEmpty;
+import static com.google.common.util.concurrent.Futures.immediateFuture;
+import static com.google.common.util.concurrent.Futures.transform;
+import static com.google.common.util.concurrent.Futures.withFallback;
+import static com.spotify.helios.common.VersionCompatibility.HELIOS_SERVER_VERSION_HEADER;
+import static com.spotify.helios.common.VersionCompatibility.HELIOS_VERSION_STATUS_HEADER;
+import static java.lang.String.format;
+import static java.net.HttpURLConnection.HTTP_BAD_METHOD;
+import static java.net.HttpURLConnection.HTTP_BAD_REQUEST;
+import static java.net.HttpURLConnection.HTTP_FORBIDDEN;
+import static java.net.HttpURLConnection.HTTP_NOT_FOUND;
+import static java.net.HttpURLConnection.HTTP_OK;
+import static java.util.Arrays.asList;
+import static java.util.Collections.singletonList;
+import static java.util.concurrent.Executors.newScheduledThreadPool;
+import static java.util.concurrent.TimeUnit.SECONDS;
 
-import com.fasterxml.jackson.core.type.TypeReference;
-import com.fasterxml.jackson.databind.JavaType;
-import com.fasterxml.jackson.databind.type.TypeFactory;
 import com.spotify.helios.common.HeliosException;
 import com.spotify.helios.common.Json;
 import com.spotify.helios.common.Resolver;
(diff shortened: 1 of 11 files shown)


mattnworb:
@spotify/helios-team

helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
```diff
@@ -336,7 +342,32 @@ public Integer apply(final Response reply) {
...
   public ListenableFuture<List<String>> listHosts() {
-    return get(uri("/hosts/"), new TypeReference<List<String>>() {
+    return listHosts(ImmutableMultimap.<String, String>of());
+  }
+
+  public ListenableFuture<List<String>> listHosts(final String namePattern) {
+    return listHosts(ImmutableMultimap.of("namePattern", namePattern));
+  }
+
+  public ListenableFuture<List<String>> listHosts(final List<String> labels) {
```
gimaker: s/labels/hostSelectors

helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
```diff
@@ -336,7 +342,32 @@ public Integer apply(final Response reply) {
...
+    final Multimap<String, String> query = HashMultimap.create();
+    query.putAll("labels", labels);
+
+    return listHosts(query);
+  }
+
+  public ListenableFuture<List<String>> listHosts(final String namePattern,
+                                                  final List<String> labels) {
```
gimaker: The code would be less confusing to read if you used `List<HostSelector>` instead of `List<String>` here -- but I see how that would be tricky if you want to send them as query parameters (since host selectors need to encoded as json).

gimaker:
:-1: untill you change all occurrences of "label" to "hostSelector".

Agents have labels, which have values. E.g. a label `foo` with a value of `bar`. host selectors act on the labels and their values, e.g. `foo in (bar, baz)`.

My first thought when seeing `listHosts(List<String> labels)` was that calling `listHosts(asList('foo', 'bar'))` would return agents that had both the `foo` and `bar` labels (disregarding the value of the labels themselves).

mattnworb:
@gimaker makes sense, will update.

helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
gimaker: Although a query parameter like `%7B%22label%22%3A%22foo%22%2C%22operator%22%2C%22EQUALS%22%2C%22operand%22%2C%22bar%22%7D` isn't terrible unless a human is supposed to craft it :)

mattnworb: I would prefer to leave it as a List of String so that the HeliosClient user can pass in any form of selector without having to necessarily have the operator defined in it's code - since HostSelector and the parsing is done in a class included in HeliosClient. This way if there are new operators it won't require a new HeliosClient version.

We could though add overloads of this method for both parsed and unparsed parameters if anyone finds that useful

```java
listHost(List<String> unparsedSelectors);
listHost(List<HostSelector> selectors);
```

gimaker: I am a-okay with `List<String>`. Functionality-wise it makes sense. It just makes the code a tad harder to follow.

Using unparsed host selectors is also a lot more friendly to the user if you do a request with `curl`.

mattnworb:
@gimaker updated. I also changed the `helios hosts` command argument from `--labels` to `--selector` and updated the description.

helios-tools/src/main/java/com/spotify/helios/cli/command/HostListCommand.java
```diff
@@ -109,25 +104,36 @@ public String apply(final HostStatus.Status input) {
...
         .setDefault(new ArrayList<String>())
         .nargs("+")
-        .help("Host selector expression. Hosts matching this expression will be part of the " +
-              "deployment-group. Multiple conditions can be specified, separated by spaces (as " +
-              "separate arguments). If multiple conditions are given, all must be fulfilled. " +
-              "Operators supported are =, !=, in and notin. Example: foo=bar baz!=qux");
+        .help("Host selector expression. The list of hosts will be filtered to match only those "
+              + "whose labels match all of the supplied expressions. "
+              + "Multiple expressions can be specified, separated by spaces "
+              + "(e.g. `-s site=foo bar!=yes`). " +
```
gimaker: Does `-s site=foo bar!=yes` actually work? Don't you have to specify `-s` once for each selector? E.g. `-s site=foo -s bar!=yes`

helios-services/src/main/java/com/spotify/helios/master/resources/HostsResource.java
```diff
@@ -88,8 +93,35 @@ public HostsResource(final MasterModel model) {
...
+    if (!hostSelectors.isEmpty()) {
+
+      final Map<String, Map<String, String>> hostsAndLabels = hosts.stream()
+          .collect(Collectors.toMap(Function.identity(),
+                                    host -> model.getHostStatus(host).getLabels()
+                   )
+          );
+
+      final List<HostSelector> selectors = hostSelectors.stream()
+          .map(HostSelector::parse)
```
gimaker: note that `HostSelector::parse` returns `null` on bad input

I'm guessing this will yield a 500 with no error message on bad input?

helios-tools/src/main/java/com/spotify/helios/cli/command/HostListCommand.java
mattnworb: I _thought so_, but `-s foo=bar -s x=y` resulted in the second argument clobbering the first and only the `x=y` selector was sent in the HTTP request.

gimaker:
:+1: but you might want to add code (and a test) to handle cases where the specified host selector doesn't parse

helios-tools/src/main/java/com/spotify/helios/cli/command/HostListCommand.java
gimaker: That's probably because you removed `action(append())`. With the current implementation specifying `-l` seems to work. OTOH only specifying `-l` once also seems to work.

I just wonder how this interacts with positional arguments...

mattnworb:
Good point about unparseable selectors. I added logic for the server to return `400 Bad Request` in those cases, as it seems like the behavior without that was to ignore the bad selector and return the unfiltered list (or the list that matched any other filters in the request).

gimaker:
:+1:

mattnworb:
we have the positional argument problem in existing code already:

```
$ helios --version
Spotify Helios CLI 0.8.812
Tested on Docker 1.6.2
$ helios -d foo.spotify.net hosts --label foo=bar --label x=y brown
Bad host selector expression: 'brown'
```

I'll fix in a new PR.

PR #958, opened 2016-07-27
bugfix: HostListCommand selector argument is greedy

mattnworb:
The use of `nargs("+")` for an optional argument combined with a
positional argument with a default value leads to some unexpected
behavior. A command like:

```
helios hosts --selector foo=bar blah
```

results in the arguments being parsed as if there were two `selector`
arguments - one for `foo=bar` and one for `blah`.

The correct way to allow multiple values for an optional argument is to
instead use `action(append())` and not `nargs(..)`.

With this change, all of these combinations of arguments are interpreted
as expected:

```
helios hosts --selector foo=bar pattern
helios hosts pattern --selector foo=bar
helios hosts --selector foo=bar --selector three=four pattern
helios hosts pattern --selector foo=bar --selector three=four
```


mattnworb:
@gimaker @davidxia

mattnworb:
@spotify/helios-team

gimaker:
:+1:

PR #961, opened 2016-08-03

mattnworb:
user's project.

### DIFF
--- a/helios-api-documentation/pom.xml
+++ b/helios-api-documentation/pom.xml
@@ -154,25 +154,4 @@
       </build>
     </profile>
   </profiles>
-
-  <reporting>
-    <plugins>
-      <plugin>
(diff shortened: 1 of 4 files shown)


mattnworb:
@davidxia 

a 3rd attempt at this - completely removes all trace of this plugin. replaces #947

davidxia:
I had similar test failures. Haven't spent much time yet on figuring out why.

mattnworb:
@davidxia I found one legit failure from #947 having to do with the wrong version of javax-rs being included in the pom, so I wouldn't be surprised if the others are legit too. 

Pretty weird that this dependency would have had a key effect in what other dependencies get included in the project 😐

mattnworb:
@davidxia found the issue - ✅

davidxia:
Nice. 👍

PR #962, opened 2016-08-11
solo: Allow clients to retrieve the solo agent name

rohansingh:
This can be useful in case you want to manually interact with the Helios
deployment. Previously you would have to call `solo.client().listHosts`.


rohansingh:
@spotify/helios-team

mattnworb:
👍

PR #963, opened 2016-08-12
solo: Redirect container stdout/err to terminal

rohansingh:
Really handy for debugging things.

### DIFF
--- a/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloLogService.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloLogService.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright (c) 2016 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.helios.testing;
+
+import static com.google.common.base.Strings.isNullOrEmpty;
+import static com.spotify.docker.client.DockerClient.LogsParam.follow;
+import static com.spotify.docker.client.DockerClient.LogsParam.stderr;
+import static com.spotify.docker.client.DockerClient.LogsParam.stdout;
+
+import com.spotify.docker.client.DockerClient;
+import com.spotify.docker.client.LogStream;
+import com.spotify.helios.client.HeliosClient;
+import com.spotify.helios.common.descriptors.HostStatus;
+import com.spotify.helios.common.descriptors.JobId;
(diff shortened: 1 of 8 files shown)


rohansingh:
@spotify/helios-team

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloLogService.java
```diff
@@ -0,0 +1,111 @@
...
+              log.info("attaching stdout/stderr for job={}, container={}", status.getJob().getId(),
+                       containerId);
+              logStream.attach(System.out, System.err);
+            } catch (Exception e) {
+              // do nothing, we'll retry in a bit
+              log.debug("error streaming logs for job={}, container={} - {}",
+                        status.getJob().getId(), containerId, e);
+            }
+          }
+        }, 0, 100, TimeUnit.MILLISECONDS);
```
mattnworb: why does attaching to the container need to be scheduled at a fix rate? won't this cause the attaching to happen many times for each container (and log the INFO message above too)?

rohansingh: Right, it should be a fixed delay, not a fixed rate. Though also I'm now realizing that this will retry attaching a bunch if a given container exits.

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloLogService.java
```diff
@@ -0,0 +1,104 @@
...
+            try {
+              final LogStream logStream = dockerClient.logs(containerId, stdout(), stderr(),
+                                                            follow());
+              log.info("attaching stdout/stderr for job={}, container={}", jobId, containerId);
+              logStream.attach(System.out, System.err);
+            } catch (final Exception e) {
+              log.warn("error streaming log for job={}, container={} - {}", jobId, containerId, e);
+            }
+          }
+        }, 100, TimeUnit.MILLISECONDS);
```
mattnworb: not very important, but - does the delay matter at all? Is it enough to call `executor.submit(runnable)` and just know that the task will be executed asynchronously from this Service / run loop? (i.e. whenever the executor gets to it)

mattnworb:
I think it would be nice if there was a way to disable this behavior, in case your docker container was very noisy with it's logs.

Otherwise 👍  awesome idea

rohansingh:
@mattnworb Mind taking another look? It's configurable now so that you can provide your own `OutputStream` for stdout/stderr, or disable log streaming entirely.

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloLogService.java
```diff
@@ -0,0 +1,117 @@
...
+                logStream.attach(logStreamProvider.getStdoutStream(jobId, containerId),
+                                 logStreamProvider.getStderrStream(jobId, containerId));
+              } catch (final InterruptedException e) {
+                // we're shutting down, no need to log anything
+              } catch (final Exception e) {
+                log.warn("error streaming log for job={}, container={} - {}", jobId, containerId,
+                         e);
+              }
+            }
+          }, 0, TimeUnit.MILLISECONDS);
```
mattnworb: nitpicky but isn't `schedule(Runnable, long, Timeunit)` the same as `submit(Runnable)`?

mattnworb:
👍  i like the idea to swap in different outputstreams

rohansingh:
Addressed all feedback and added some tests, going to merge this now.

PR #964, opened 2016-08-16
Fix race condition in HeliosSoloDeploymentTest

rohansingh:
There was a race condition in `testUndeployLeftoverJobs` such that if the
`heliosClient.hostStatus()` mock was called too many times, it would return
an incorrect value. This only became an actual problem recently since the
new `HeliosSoloLogService` calls `hostStatus()` in a scheduled loop.


rohansingh:
@spotify/helios-team

rohansingh:
I'm going to go ahead and merge this just so the Jenkins build is unblocked.

mattnworb:
It might be easier conceptually if you used `thenAnswer` with `hostStatus(..)` instead so you could decide which value to return based on a condition like whether undeploy has been called already or not, but this works too 👍

PR #968, opened 2016-08-25
Add flag to agent to control writing job history to ZK

davidxia:
Default is true which is previous behavior.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/agent/AgentConfig.java
+++ b/helios-services/src/main/java/com/spotify/helios/agent/AgentConfig.java
@@ -66,6 +66,7 @@ public class AgentConfig extends Configuration {
   private String zooKeeperAclAgentPassword;
   private FastForwardConfig fastForwardConfig;
   private List<String> extraHosts;
+  private boolean jobHistoryDisabled;
 
   public boolean isInhibitMetrics() {
     return inhibitMetrics;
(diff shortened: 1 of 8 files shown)


mattnworb:
👍

davidxia:
@mattnworb @rohansingh I also added javadoc to `OldJobReaper` since it will be affected if this flag is enabled.

davidxia:
cc @saunakchakrabarti

davidxia:
Was thinking about adding a test for the flag. The relevant class is `JobHistoryTest` which is an IT. But there's not really a good way to test that no event will ever be sent. I didn't feel like sleeping a bunch and then checking that there's no events would be a worthwhile test.

mattnworb:
@davidxia I think it would be sufficient to add a unit test of `ZooKeeperAgentModel` that sees what it does when the TaskHistoryWriter instance is null

davidxia:
@mattnworb I renamed the flag and added a test in b9bb5f3.

helios-services/src/main/java/com/spotify/helios/agent/AgentConfig.java
```diff
@@ -359,6 +360,15 @@ public AgentConfig setFfwdConfig(FastForwardConfig config) {
     return this;
   }
 
+  public boolean getNoJobHistory() {
```
mattnworb: `isJobHistoryDisabled()` would be a more readable way to phrase this

helios-services/src/main/java/com/spotify/helios/agent/AgentParser.java
```diff
@@ -265,6 +267,11 @@ protected void addArgs(final ArgumentParser parser) {
         .help("ZooKeeper agent password (for ZooKeeper ACLs). If the "
               + ZK_AGENT_PASSWORD_ENVVAR
               + " environment variable is present this argument is ignored.");
+
+    noJobHistory = parser.addArgument("--no-job-history")
```
mattnworb: `--disable-job-history` might be more intuitive ?

mattnworb:
👍  regardless of how you want to handle the above comments

helios-services/src/main/java/com/spotify/helios/agent/AgentParser.java
davidxia: Yea, I was trying to save chars.

PR #969, opened 2016-08-26
Rate-limit reapers to avoid hammering ZooKeeper

davidxia:
The various reapers retrieve a list of items from ZK, usually jobs. They then do more ZK operations for each item.

The reapers all start at once when the master service is started.

thousands of jobs * 4 reapers * many operations per job * several masters = potentially tens of thousands of ZK operations all at once which is bad

Rate-limit the reapers to keep ZK happy.


davidxia:
@spotify/helios-team 

Not sure what to do about `ExpiredJobReaper` which runs every 30 seconds. Ideally, there'd be a push instead of constant polling.

helios-services/pom.xml
```diff
@@ -344,6 +344,10 @@
       <artifactId>riemann-java-client</artifactId>
       <version>0.2.10</version>
     </dependency>
+    <dependency>
+      <groupId>org.awaitility</groupId>
+      <artifactId>awaitility</artifactId>
+    </dependency>
```
mattnworb: i think this is missing scope=test

helios-services/src/main/java/com/spotify/helios/master/reaper/DeadAgentReaper.java
```diff
@@ -0,0 +1,116 @@
...
+  private static final Logger log = LoggerFactory.getLogger(DeadAgentReaper.class);
+
+  private final MasterModel masterModel;
+  private final long timeoutMillis;
+  private final Clock clock;
+
+  public DeadAgentReaper(final MasterModel masterModel,
+                         final long timeoutHours) {
+    this(masterModel, timeoutHours, SYSTEM_CLOCK, PERMITS_PER_SECOND,
+         ThreadLocalRandom.current().nextInt(DELAY));
```
mattnworb: minor: there is no real benefit (or cost, AFAIK) to using ThreadLocalRandom here over a simple `new Random().nextInt(DELAY)` - there is no contention here.

helios-services/src/main/java/com/spotify/helios/master/reaper/RateLimitedService.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  /**
+   * Collect items to process.
+   * @return An iterable of T.
+   */
+  abstract Iterable<T> collectItems();
+
+  /**
+   * Process the tiem.
```
mattnworb: typo in `tiem`

mattnworb:
+1 to rate-limiting these reapers. I think that makes a lot of sense.

I'd prefer if the rates could be controlled through configuration though, rather than hard-coding - to make it easy to have different reapers use different rates than each other if that is ever desired. 

I think there are two other things that would go a long way towards improving the load generated by these reapers. We could do one or the other and get some benefit, but I think doing both together make the most sense:

1. Only have one master at a time running reapers. [LeaderLatch in Curator](http://curator.apache.org/curator-recipes/leader-latch.html) makes this really easy to implement and is battle-tested. I know we have had some hesitation in the past to elect leaders for tasks among the helios masters, but this seems like a prime use case. 

2. Allow the reapers to be scheduled on some sort of cron-like schedule, rather than the hard-coded rate of every 30 minutes. This would allow for quick reconfiguration of the reaper schedule to run less often if/when it causes high load on Zookeeper. For a lot of these reapers, it seems irrelevant if they are run every 30 minutes or just once a day, perhaps at some period of low activity overnight in the cluster.

saunakchakrabarti:
Would the Guava rate-limiter be appropriate for this use case?  https://google.github.io/guava/releases/19.0/api/docs/index.html?com/google/common/util/concurrent/RateLimiter.html

Instead of specifying a specific poll interval, one can specify the the maximum number of messages per second (with concurrency considered).

davidxia:
@saunakchakrabarti Yes, I use it in `RateLimitedService`.

helios-services/pom.xml
davidxia: Thanks

davidxia:
> rates could be controlled through configuration

I thought about this but was hesitant to add a CLI switch for every reaper to helios-master. If we think it's worth it though, I can add it.

2. Just for context: All the reapers except for `ExpiredJobReaper` are time-insensitive in the sense that they don't have to run that often. `DeadAgentReaper` is the one that runs every half hour. This is more often than the others which are once a day to reduce the chance that users who recycle hosts have to manually deregister DOWN hosts to reclaim the hostname namespace.

helios-services/src/main/java/com/spotify/helios/master/reaper/DeadAgentReaper.java
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * De-registers dead agents. An agent that has been DOWN for more than X hours is considered
+ * dead.
+ */
+public class DeadAgentReaper extends RateLimitedService<String> {
+
+  private static final double PERMITS_PER_SECOND = 0.2; // one permit every 5 seconds
+  private static final Clock SYSTEM_CLOCK = new SystemClock();
+  private static final int DELAY = 30;
+  private static final TimeUnit TIME_UNIT = TimeUnit.MINUTES;
```
rohansingh: Why is this needed? If you want to store a type of time unit with a value, `Duration` is probably a better choice.

helios-services/src/main/java/com/spotify/helios/master/reaper/DeadAgentReaper.java
```diff
@@ -0,0 +1,116 @@
...
+  Iterable<String> collectItems() {
+    return masterModel.listHosts();
+  }
+
+  @Override
+  void processItem(final String agent) {
+    log.info("Deciding whether to reap dead agent.");
+    try {
+      final HostStatus hostStatus = masterModel.getHostStatus(agent);
+      if (hostStatus == null || hostStatus.getStatus() != HostStatus.Status.DOWN) {
```
rohansingh: Should probably be `!HostStatus.Status.Down.equals(...)`

helios-services/src/main/java/com/spotify/helios/master/reaper/DeadAgentReaper.java
```diff
@@ -53,56 +54,63 @@
 
   public DeadAgentReaper(final MasterModel masterModel,
                          final long timeoutHours) {
-    this(masterModel, timeoutHours, SYSTEM_CLOCK);
+    this(masterModel, timeoutHours, SYSTEM_CLOCK, PERMITS_PER_SECOND,
+         new Random().nextInt(DELAY));
```
dflemstr: Minor: `ThreadLocalRandom`?  Also this is logic in ctor, consider Effective Java item #1: create a static factory method.

helios-services/src/main/java/com/spotify/helios/master/reaper/DeadAgentReaper.java
```diff
@@ -53,56 +54,63 @@
...
   DeadAgentReaper(final MasterModel masterModel,
                   final long timeoutHours,
-                  final Clock clock) {
+                  final Clock clock,
+                  final double permitsPerSecond,
+                  final int initialDelay) {
+    super(permitsPerSecond, initialDelay, DELAY, TIME_UNIT);
     this.masterModel = masterModel;
     checkArgument(timeoutHours > 0);
     this.timeoutMillis = TimeUnit.HOURS.toMillis(timeoutHours);
```
dflemstr: Consider using `java.time.Duration` instead of `long`

helios-services/src/main/java/com/spotify/helios/master/reaper/DeadAgentReaper.java
davidxia: Why? `hostStatus.getStatus()` is an enum so I can just use `!=` right? I still have to do `hostStatus == null` to prevent an NPE in `hostStatus.getStatus()`.

helios-services/src/main/java/com/spotify/helios/master/reaper/DeadAgentReaper.java
davidxia: Haha, see https://github.com/spotify/helios/pull/969#discussion_r76539981

helios-services/src/main/java/com/spotify/helios/master/reaper/DeadAgentReaper.java
davidxia: I need `TimeUnit` for `RateLimitedService` because `ScheduledExecutorService.scheduleWithFixedDelay()` takes `TimeUnit`.

helios-services/src/main/java/com/spotify/helios/master/reaper/DeadAgentReaper.java
dflemstr: Ha, somehow I ended up in some parallel universe without those comments :D Sorry, trying to get used to this style of reviews

davidxia:
Going to merge this unless someone has a good suggestion for how not to duplicate the rate limiting code and tests.

PR #970, opened 2016-08-30
Add meters for ZooKeeper operations

davidxia:
Currently there are only gauges for outstanding ZooKeeper
operations. Add meters to see the rate of various types of
ZooKeeper operations. We can sum these meters to get the rate of all
operations.


mattnworb:
A [Timer combines a Histogram and a Meter](https://github.com/dropwizard/metrics/blob/v2.1.5/metrics-core/src/main/java/com/yammer/metrics/core/Timer.java#L10-L14), so the call to `metrics.updateTimer(..)` and `registry.timer(..).update(...)` should already be recording the rate of those events.

PR #971, opened 2016-08-30
Make reapers not override RateLimitedService.schedule

davidxia:


davidxia:
@spotify/helios-team

mattnworb:
does this mean that before this PR the result of `Random.nextInt(delay)` for the `initialDelay` was completely unused?

mattnworb:
a bit tangental but I find it very confusing that the constructor parameters to the subclasses (like `DeadAgentReaper`) are in a completely different order than the superclass of RateLimitedService

davidxia:
@mattnworb Yes :(

I tried to change the constructors of the subclasses very little. `DeadAgentReaper`'s public constructor's parameters are the same. I appended parameters to its package-local constructor, but aren't the parameters in the signature that are then passed to the parent class in the same order, i.e. `permitsPerSecond` and `initialDelay`?

PR #973, opened 2016-08-31
fix NPE in Supervisor.Stop when job has gracePeriod

mattnworb:
It's possible for a Supervisor to receive a goal of UNDEPLOY for a job that
it has not been told to START - perhaps because the agent was restarted
during the undeploy of a previous job (during the grace period sleeping).

Since the TaskRunner is only set after receiving the START goal, the call
to `runner.unregister()` when the job being undeployed throws a
NullPointerException.

Add a test to verify behavior.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/agent/Supervisor.java
+++ b/helios-services/src/main/java/com/spotify/helios/agent/Supervisor.java
@@ -349,14 +349,16 @@ public class Supervisor {
       statusUpdater.setState(STOPPING);
       statusUpdater.update();
 
-      final Integer gracePeriod = job.getGracePeriod();
-      if (gracePeriod != null && gracePeriod > 0) {
-        log.info("Unregistering from service discovery for {} seconds before stopping",
-                 gracePeriod);
+      if (runner != null) {
+        final Integer gracePeriod = job.getGracePeriod();
+        if (gracePeriod != null && gracePeriod > 0) {
+          log.info("Unregistering from service discovery for {} seconds before stopping",
+              gracePeriod);
 
-        if (runner.unregister()) {
-          log.info("Unregistered. Now sleeping for {} seconds.", gracePeriod);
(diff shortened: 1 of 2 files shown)


helios-services/src/main/java/com/spotify/helios/agent/Supervisor.java
```diff
@@ -349,14 +349,16 @@ public void perform(final boolean done) throws InterruptedException {
...
 
-      final Integer gracePeriod = job.getGracePeriod();
-      if (gracePeriod != null && gracePeriod > 0) {
-        log.info("Unregistering from service discovery for {} seconds before stopping",
-                 gracePeriod);
-
-        if (runner.unregister()) {
-          log.info("Unregistered. Now sleeping for {} seconds.", gracePeriod);
-          sleeper.sleep(TimeUnit.MILLISECONDS.convert(gracePeriod, TimeUnit.SECONDS));
+      if (runner != null) {
```
mattnworb: here is the actual fix

helios-services/src/test/java/com/spotify/helios/agent/SupervisorTest.java
```diff
@@ -396,12 +365,78 @@ public void verifyRuntimeExceptionSetsTaskStatusToFailed() throws Exception {
...
-    return new Answer<Object>() {
+    return (Answer<Object>) invocation -> future.get();
+  }
+
+  /**
+   * Verifies a fix for a NPE that is thrown when the Supervisor receives a goal of UNDEPLOY for a
+   * job with gracePeriod that has never been STARTed.
+   */
+  @Test
+  public void verifySupervisorHandlesUndeployingOfNotRunningContainerWithGracePeriod()
```
mattnworb: and here is the actual test - the rest of the changes in SupervisorTest are refactoring to make this test possible (to use a different Job instance)

mattnworb:
@spotify/helios-team

gimaker:
:+1:

PR #975, opened 2016-09-07
prevent HeliosSoloLogService from continuing to run after tests with HeliosSoloDeployment are done

mattnworb:
In a test class that uses HeliosSoloDeployment as a `@Rule` (i.e. one
helios-solo instance per test method), I found that the test was hanging
for a long period of time after the test itself seemed to be finished
(judging from the class-under-test's log output).

Thread dumps showed that the thread running HeliosSoloLogService was
blocked at a call to `heliosClient.listHosts().get()`. This change adds
timeouts when using `Future.get()` to avoid any possibility of blocking
forever.

Also change the Service itself to use daemon threads with it's executor,
as we do not wish to keep the JVM running under any instance if the
HeliosSoloLogService is still running when the main thread has exited.

### DIFF
--- a/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloLogService.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloLogService.java
@@ -31,13 +31,19 @@ import com.spotify.helios.common.descriptors.TaskStatus;
 
 import com.google.common.collect.Maps;
 import com.google.common.util.concurrent.AbstractScheduledService;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.jetbrains.annotations.NotNull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executors;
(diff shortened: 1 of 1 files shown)


mattnworb:
@rohansingh @spotify/helios-team

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloLogService.java
```diff
@@ -58,12 +64,26 @@
...
+  private static <T> T get(Future<T> future)
+      throws InterruptedException, ExecutionException, TimeoutException {
+    return future.get(1, TimeUnit.SECONDS);
+  }
+
+  @Override
+  protected ScheduledExecutorService executor() {
+    final ThreadFactory threadFactory = new ThreadFactoryBuilder()
+        .setNameFormat(serviceName())
+        .setDaemon(false)
```
davidxia: Shouldn't this be set to true as daemon threads do not prevent the JVM from exiting when the program finishes.

mattnworb: oops, that was my intent. thanks for catching

davidxia:
LGTM

PR #976, opened 2016-09-08
helios-testing: include error messages when throwing TimeoutException

mattnworb:
When a call to `Polling.await(..)` fails within helios-testing / TemporaryJobs, it is very hard to troubleshoot what went wrong when there is no message in the thrown exception.

### DIFF
--- a/helios-testing/src/main/java/com/spotify/helios/testing/Polling.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/Polling.java
@@ -17,18 +17,18 @@
 
 package com.spotify.helios.testing;
 
-import java.util.concurrent.Callable;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
-
 import static com.google.common.base.Throwables.propagate;
 import static com.google.common.base.Throwables.propagateIfInstanceOf;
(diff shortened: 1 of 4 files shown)


mattnworb:
@spotify/helios-team

negz:
LGTM

helios-testing/src/main/java/com/spotify/helios/testing/Polling.java
```diff
@@ -37,13 +37,14 @@
       }
       Thread.sleep(500);
     }
-    throw new TimeoutException();
+    throw new TimeoutException(String.format(message, timeout, timeUnit.toString().toLowerCase()));
```
davidxia: Sorry I only got a chance to see this now.

Why not just use `message` instead of formatting it? Or `message` + `String.format("Timed out after %d %s", timeout, timeUnit.toString().toLowerCase())`. It's not obvious to callers that the message they pass in will be string formatted.

mattnworb: I figured I could avoid a tiny bit of concatenation at the call site - although really what I was saving was the line on the caller side being even longer. I figured since the only callers are within this same test code that the lack-of-obviousness would be ok

PR #977, opened 2016-09-12
Configurable container killing

negz:
@mattnworb

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
@@ -17,6 +17,19 @@
 
 package com.spotify.helios.common.descriptors;
 
+import static com.google.common.base.Charsets.UTF_8;
+import static com.google.common.base.MoreObjects.firstNonNull;
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Throwables.propagate;
+import static com.spotify.helios.common.Hash.sha1digest;
+import static java.util.Collections.emptyList;
+import static java.util.Collections.emptyMap;
+import static java.util.Collections.emptySet;
+
+import com.spotify.helios.common.Json;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.Optional;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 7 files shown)


helios-services/src/main/java/com/spotify/helios/agent/Supervisor.java
```diff
@@ -375,7 +385,13 @@ public void perform(final boolean done) throws InterruptedException {
...
           .build().newScheduler();
 
-      // Kill the container after stopping the runner
+
+      // TODO(negz): Use Docker stop instead of kill?
+      // I assume this loop is intended to handle the case where the above runner.stop() fails to
+      // communicate with the Docker API, because runner.stop() already sends a SIGKILL if
+      // necessary. What good could sending more SIGKILLs do if the first signal was sent
+      // successfully? If we *are* working around Docker API failures why not send more SIGTERMs
+      // to give the container the chance to shutdown gracefully?
```
negz: I'm tempted to address this TODO as a separate commit in this PR, but I'm not quite sure about the current implementation.

Do we ask the runner to stop the container, but then fall back to killing it in the supervisor because we want to handle the case where the container is somehow running but its runner is not? If that's not a concern it might be nicer to move the retry logic into the runner and just have it send a bunch of `docker.stopContainer()`s with the requested SIGKILL deadline.

helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
```diff
@@ -188,6 +190,8 @@
    * @param addCapabilities Linux capabilities to add for the container. Optional.
    * @param dropCapabilities Linux capabilities to drop for the container. Optional.
    * @see <a href="https://docs.docker.com/reference/run/#network-settings">Docker run reference</a>
+   * @param secondsToWaitBeforeKill The time to ask Docker to wait after sending a SIGTERM to the
+   *    container's main process before sending it a SIGKILL. Optional. Defaults to 120s.
```
mattnworb: nitpick but it does not default to 120s in regards to the value of the field in this Job class. The agent will default to using 120s if the Job. secondsToWaitBeforeKill is empty/null. I rather leave the value undocumented in this class so we can change the agent if we have to later without changing this part (so it's really just an issue of moving the documentation).

helios-services/src/main/java/com/spotify/helios/agent/Supervisor.java
mattnworb: To be honest I am not sure about the intent of this loop here - maybe @rohansingh or @rculbertson might remember.

mattnworb:
LGTM 👍

helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
negz: Ack, SGTM.

PR #979, opened 2016-09-14
Add seconds to wait before kill to JobInspectCommand

davidxia:


davidxia:
@negz @mattnworb

mattnworb:
nice catch 👍

PR #980, opened 2016-09-15

negz:
Reviewer: @mattnworb 
Interested parties: @spotify/helios-devs 

This change also has the side effect of not writing rolling update tasks for machines that *did not* change since the last check. Previously we would queue tasks to update hosts that were already running the target job, and those tasks would be treated as no-ops.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/master/HostChanges.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/HostChanges.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2016 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.helios.master;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Sets;
+
+import java.util.List;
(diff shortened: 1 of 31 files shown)


helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUndeployPlanner.java
```diff
@@ -0,0 +1,71 @@
...
+
+  @Override
+  public List<RolloutTask> plan(final Map<String, HostStatus> hostsAndStatuses) {
+    // we only care about hosts that are UP
+    final List<String> hosts = Lists.newArrayList();
+    for (final Map.Entry<String, HostStatus> entry : hostsAndStatuses.entrySet()) {
+      if (entry.getValue().getStatus().equals(HostStatus.Status.UP)) {
+        hosts.add(entry.getKey());
+      }
+    }
```
mattnworb: YMMV on whether this is more or less readable, but this could also be 

```java
final List<String> hosts = hostsAndStatuses.entrySet().stream()
  // we only care about hosts that are UP
  .filter(entry -> entry.getValue.getStatus() == HostStatus.Status.UP)
  .map(entry -> entry.getKey())
  .collect(Collectors.toList());
```

helios-services/src/main/java/com/spotify/helios/master/HostChanges.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.util.List;
+import java.util.Set;
+
+class HostChanges  {
+  private final List<String> added;
+  private final List<String> removed;
+
+  HostChanges(final Set<String> newHosts, final Set<String> currentHosts) {
+    this.added = ImmutableList.copyOf(Sets.difference(newHosts, currentHosts).immutableCopy());
```
negz: Ah good to know. I felt like I might have been doing a little too much copying here.

helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUndeployPlanner.java
negz: Thanks. This was copy-pasta but your proposed style fits with the rest of the class better.

PR #982, opened 2016-09-16

negz:
Reverts spotify/helios#981

This is a revert of a revert of https://github.com/spotify/helios/pull/980, which means all the unrelated imports changes are no longer broken into separate commits. :(

@gimaker please let me know where the race you spotted is.


gimaker:
So here's the suspect bug in #980:

Consider the following situation:
 * Some hosts (e.g. 20) are removed from a DG, which triggers a "rolling undeploy".
 * Immediately after this a user triggers a `rolling-update` from the CLI. This will overwrite the "rolling undeploy". At this point we no longer know that we were supposed to remove the job from those 20 removed hosts.

negz:
@gimaker Ack.

So (mostly to remind myself next week) the issue is that:
* The first (HOSTS_CHANGED) rolling update saves the updated DG member hosts to ZK and writes a bunch of rolling update and undeploy ops to ZK.
* Any subsequent rolling update would only generate update ops for the hosts that remained in the DG after the previous HOSTS_CHANGED rolling update, and would overwrite the previously written update and undeploy ops with the newly generated ones.

Note there's an existing check that does not allow new HOSTS_CHANGED rolling updates for DGs that are in state ROLLING_OUT, but we're still at risk because a DG that becomes FAILED due to a HOSTS_CHANGED rolling update is susceptible to the issue described above.

mattnworb:
Since we don't undeploy from removed hosts at all today, I think this should be seen as a improvement that could be made and not a blocker.

PR #984, opened 2016-09-16
add logically-equal comparison test for HostSelectors

mattnworb:
This adds a test to compare HostSelectors and Collections of
HostSelectors to find out if they are "logically equal". Currently the
only logically equal selectors are `foo = bar` and `foo in (bar)`.

This is intended to  remove an error message about conflicting host
selectors if a user tries to create a deployment group with selectors
`[foo in (bar), pool = a]` when a deployment group already exists
with selectors `[foo = bar, pool = a]`.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/HostSelector.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/HostSelector.java
@@ -22,12 +22,18 @@ import static java.lang.String.format;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.collect.ComparisonChain;
 import com.google.common.collect.Iterables;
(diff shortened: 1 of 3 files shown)


mattnworb:
@gimaker

gimaker:
Why do we need this? Is it really worth the trouble? If so then we should go all the way and make `foo notin (bar)` logically equivalent to `foo != bar`.

mattnworb:
@gimaker I thought it would be helpful for convenience reasons for someone who has a DG with a selector like `foo = bar` but they accidentally script (or have a tool that does) creation of `foo in (bar)`.

I am not sure if it is worth the extra code though.

You are right - I should add `!= x` and `not in (x)` as synonyms too.

dflemstr:
FWIW I think the "proper" solution for this would be to at the earliest possible moment (i.e. when parsing the label expression) normalize the selectors to one format, i.e. actually store `foo != x` as `foo notin (x)`

PR #985, opened 2016-09-21
WIP: Undeploy a deployment group's job from hosts that are removed

negz:
The git log for this is pretty messy due to attempting to build on reverted commits. I've cherry picked the reverted commits back into my branch, but every time I try to rebase on master said cherry picked commits disappear.

I'm thinking it might work to revert the previous revert, then rebase this branch against master so it contains only the new commits. I'm open to other git-taming suggestions.


negz:
See https://github.com/spotify/helios/pull/982 instead - I was able to make this fairly readable by cherry-picking my new commits into the revert-of-the-revert branch.

PR #986, opened 2016-09-21

negz:
Reviewers: @gimaker 
CC: @spotify/helios-devs 

This PR has a complicated history.

* It was originally reviewed and merged in https://github.com/spotify/helios/pull/980
* It was reverted based on a verbal comment in https://github.com/spotify/helios/pull/981
* The verbal comment was explained in https://github.com/spotify/helios/pull/982 (an abandoned revert of https://github.com/spotify/helios/pull/981)

This PR:
* Reintroduces the commits from https://github.com/spotify/helios/pull/980 by reverting https://github.com/spotify/helios/pull/981.
* Adds an integration test 
* Returns to the previous pattern of generating rolling update operations for *all* hosts in a HOSTS_CHANGED rolling update. This is necessary to avoid a race (see https://github.com/spotify/helios/commit/68a836eb47a1732d3fd37dfe3cbe3f6f00d9e1c4 for more details)
* Checks that we actually undeploy the deployment group's job from hosts that are removed during a `HOSTS_CHANGED` rolling update. The rolling update will fail if the job cannot be undeployed.
* Stores the set of machines removed during a `HOSTS_CHANGED` rolling update, removing each machine when we're confident the deployment group's job has been undeployed from it. This set of previously removed machines is mixed back into future `HOSTS_CHANGED` rolling updates to ensure we don't forget to undeploy them.


mattnworb:
just want to say: nice PR summary 👌

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -913,6 +935,76 @@ private RollingUpdateOp rollingUpdateDeploy(final ZooKeeperClient client,
...
+
+    try {
+      // TODO(negz): Are we susceptible to a race?
+      // Master A reads list
+      // Master B reads list
+      // Master A removes host X
+      // Master B removes host Y
+      // Master A commits list without X and with Y
+      // Master B commits list without Y and with X
+      // Does failing the transaction using check() fail the whole update, or will we retry this op?
```
negz: This should not happen, because two masters cannot work on the same rolling update at the same time, right?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -350,6 +350,8 @@ public void addDeploymentGroup(final DeploymentGroup deploymentGroup)
             create(Paths.configDeploymentGroup(deploymentGroup.getName()), deploymentGroup),
             create(Paths.statusDeploymentGroup(deploymentGroup.getName())),
             create(Paths.statusDeploymentGroupHosts(deploymentGroup.getName()),
+                   Json.asBytesUnchecked(emptyList())),
+            create(Paths.statusDeploymentGroupRemovedHosts(deploymentGroup.getName()),
```
negz: I guess this would need to be created for all existing DG's? Or I could just change the code to create it if it doesn't exist?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: Per discussion with @gimaker this can happen. I'll add a `check()`.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -913,6 +935,89 @@ private RollingUpdateOp rollingUpdateDeploy(final ZooKeeperClient client,
...
+      4. Master B removes host Y
+      5. Master A commits list with Y and without X
+      6. Master B commits list with X and without Y
+
+      We would end up thinking we'd successfully removed both hosts, but in fact would have only
+      removed host Y.
+
+      The check() will cause this RollingUpdateOp's ZK transaction to fail without incrementing the
+      task index, so this RollingUpdateOp will be retried by the next master that gets to it.
+      */
```
negz: @gimaker Please confirm that this giant comment is accurate. 😄

negz:
FYI - I only noticed a bug in the new integration test when I pushed this PR. Looking into it now.

negz:
Please review at your leisure. I'd particularly appreciate feedback on the implementation of the 'WIP' commit.

negz:
Ping @gimaker. Will you have time to look at this, or would you prefer I find another reviewer?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -490,66 +495,75 @@ private boolean updateOnHostChange(final DeploymentGroup group,
...
+
+      if (hosts.equals(curHosts)) {
+        return;
+      }
+
+      final List<String> removedHosts = removedHosts(curHosts, hosts, previouslyRemovedHosts);
+
+      final List<ZooKeeperOperation> ops = Lists.newArrayList();
+      ops.add(set(Paths.statusDeploymentGroupHosts(groupName), Json.asBytes(hosts)));
+      ops.add(set(Paths.statusDeploymentGroupRemovedHosts(groupName), Json.asBytes(removedHosts)));
```
mattnworb: does the list of hosts stored in this path for "removed hosts" grow unbounded over time? It appears that way from `previouslyRemovedHosts` being passed into the `removedHosts` function.

I would be concerned that someone who uses auto-scaling with their deployment group will end up with hundreds (or more) of removed hosts in this list over time.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -893,42 +938,176 @@ private RollingUpdateOp rollingUpdateDeploy(final ZooKeeperClient client,
...
+      We would end up thinking we'd successfully removed both hosts, but in fact would have only
+      removed host Y.
+
+      The check() will cause this RollingUpdateOp's ZK transaction to fail without incrementing the
+      task index, so this RollingUpdateOp will be retried by the next master that gets to it.
+      */
+      return opFactory.nextTask(ImmutableList.of(
+          check(Paths.statusDeploymentGroupRemovedHosts(deploymentGroup.getName()), version),
+          set(Paths.statusDeploymentGroupRemovedHosts(deploymentGroup.getName()),
+              Json.asBytes(hostsToUndeploy))));
```
mattnworb: this check-then-set is done in a transaction, right? Because I don't think it catches all possible races if not in a transaction.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
mattnworb: The best approach is for every piece of code that tries to check this path to be able to handle the case where it does not exist.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
davidxia: Nice comment.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
davidxia: We have a `CreateWithDataAndVersion` class. It might be nicer to make a `SetWithDataAndVersion` class instead of doing both `check()` with version followed by a `set()` without a version and executing both in a transaction.

gimaker:
@negz I'll have time to do a proper review tomorrow.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: > does the list of hosts stored in this path for "removed hosts" grow unbounded over time? 

@mattnworb Hosts are removed from this list if and when they are 'marked undeployed' per https://github.com/spotify/helios/pull/986/commits/58fedba3cae84ed8125fee61827057d2d9cf26b7#diff-8d500c531ffb1f7836e9d5b60585a002R979.

I *think* hosts will be 'marked undeployed' if they're unregistered from the Helios master, because the `getDeployment(host, jobid)` check at  https://github.com/spotify/helios/pull/986/commits/58fedba3cae84ed8125fee61827057d2d9cf26b7#diff-8d500c531ffb1f7836e9d5b60585a002R948 will come up blank and assume the job is gone, causing the 'await stopped' rollout task to proceed to the 'mark undeployed' task.

However any agents that are removed *without* unregistering will cause the list of stored hosts to grow unbounded, and potentially worse will cause any rolling update of the deployment groups they match to fail until the agent either comes back online or is unregistered.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: @mattnworb 

> The best approach is for every piece of code that tries to check this path to be able to handle the case where it does not exist.

Agreed. Will do.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: @davidxia 

> We have a CreateWithDataAndVersion class. It might be nicer to make a SetWithDataAndVersion class instead of doing both check() with version followed by a set() without a version and executing both in a transaction.

I agree, but I don't really want to do it in this PR. The current implementation follows a pattern we're already using - https://github.com/spotify/helios/blob/master/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java#L532

helios-system-tests/src/main/java/com/spotify/helios/system/SystemTestBase.java
```diff
@@ -1000,10 +1013,13 @@ public DeploymentGroupStatus call() throws Exception {
 
         if (response != null) {
           final DeploymentGroupStatus status = response.getDeploymentGroupStatus();
-          if (status.getState().equals(state)) {
+          final DeploymentGroupStatus.State actual = status.getState();
+          if (expected.equals(actual)) {
```
negz: Why? Is there some technical advantage, or is this a style convention?

helios-services/src/test/java/com/spotify/helios/master/DeploymentGroupTest.java
```diff
@@ -133,33 +147,58 @@ public void testUpdateDeploymentGroupHosts() throws Exception {
...
+      Paths.statusDeploymentGroupTasks(dg.getName()),
+      DeploymentGroupTasks.newBuilder()
+          .setRolloutTasks(tasks)
+          .setTaskIndex(0)
+          .setDeploymentGroup(changed)
+          .build()
     );
-    verify(client, times(2)).transaction(opCaptor.capture());
-    assertThat(opCaptor.getValue(), hasItem(setDeploymentGroupHostChanged));
+    assertThat(opCaptor.getValue(), hasItem(setDeploymentGroupTasks));
```
negz: It is, thanks. It's an artifact from when I was trying a different approach.

negz:
> The last commit 65fce64 confused me a bit. Is it needed for DG migrations? Any other cases?

@davidxia I can't quite infer whether the 'it' you're asking about means my commit (65fce64), or the existing code it works around, so I'll just try to better explain what that commit is doing.

https://github.com/spotify/helios/blob/master/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java#L908

Right now the above check will prevent a rolling update from generating the ZK ops that actually do the undeploying if it appears that a subsequent deployment would just deploy the same job that `rollingUpdateUndeploy()` is trying to undeploy. That makes a lot of sense in a world where an undeployment is always immediately followed by a deployment (i.e. a rolling update). This check would kick in today if someone tried to 'rolling update' a deployment group to the same job it was already running. The rolling update would 'succeed', but nothing would ever actually be undeployed.

In this PR we want to generate undeployment ops for agents that just plan left the deployment group. It just so happens that an agent that has left the deployment group can (and likely will) have a deployment that is owned by the deployment group, and of the same job as the deployment group. So, said agents can get caught out on the above check and never actually get undeployed.

Does that make more sense, or just confuse things further?

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
davidxia: Agreed, not for this PR.

helios-system-tests/src/main/java/com/spotify/helios/system/SystemTestBase.java
davidxia: * == never throws NullPointerException
* == is subject to type compatibility check at compile time
* It's faster

See more here http://stackoverflow.com/a/2937561/553994

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
mattnworb: I think it would definitely be a problem if in fact a host whose labels no longer match the DG ends up staying in this list until the host is unregistered (which could be never).

> However any agents that are removed without unregistering will cause the list of stored hosts to grow unbounded

by "removed" do you mean the host's labels no longer match the DG selector?

negz: Discussed this with @mattnworb in person. There's only one condition I can envision where this list will grow unbounded, which is if an unbounded amount of agents are removed from a DG (i.e. the agents are still running, but with labels that no longer match the DG) *and* we're unable to undeploy the job from those machines for some reason. In this case the failure to undeploy will cause the rolling update to fail, so hopefully the DG owner will notice and take action.

I was also concerned about the case where agents disappear permanently without being degregistered from Helios (i.e. the agents' hosts are decommissioned without deregistering). This *would* cause the list to grow unbounded, except that we have code to automatically deregister these AWOL agents after ~7 days, at which point the undeploy will 'succeed' because the deployment of the DG's job to the agent will have been deleted.

I'm going to add a test for the condition explained in the second paragraph, just to be sure.

davidxia:
Makes sense.

helios-system-tests/src/main/java/com/spotify/helios/system/SystemTestBase.java
```diff
@@ -1000,10 +1013,13 @@ public DeploymentGroupStatus call() throws Exception {
...
             return status;
-          } else if (status.getState().equals(DeploymentGroupStatus.State.FAILED)) {
-            assertEquals(state, status.getState());
+          }
+          if (FAILED.equals(actual) && !expected.equals(actual)) {
```
negz: Or rather re-order it, since we do actually want to catch that condition.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -893,42 +938,176 @@ private RollingUpdateOp rollingUpdateDeploy(final ZooKeeperClient client,
...
+    final TaskStatus taskStatus = getTaskStatus(client, host, deploymentGroup.getJobId());
+    final JobId jobId = deploymentGroup.getJobId();
+
+    if (taskStatus == null) {
+      // Agent has not written job status to zookeeper.
+
+      if (getDeployment(host, jobId) == null) {
+        // The job is not listed under /config/hosts. It may have been manually undeployed?
+        // We can't undeploy it if it's not there anymore, so move on.
+        return opFactory.nextTask();
```
gimaker: Undeploying the job == removing it from `/config/hosts/<host>/jobs` (see e.g.the `undeployJob()` mehtod), right?

So isnt' this branch always going to be true? I put a breakpoint on the line after this block, L958, and ran this in the debugger and it was never hit.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -893,42 +940,176 @@ private RollingUpdateOp rollingUpdateDeploy(final ZooKeeperClient client,
     }
   }
 
+  private RollingUpdateOp rollingUpdateAwaitStopped(final ZooKeeperClient client,
```
gimaker: Agian, I think the name should be `...Undeployed` rather than `...Stopped`, since we want the job undeployed and not stopped.

Also I think this entire method could be simplified a lot:

```
  private RollingUpdateOp rollingUpdateAwaitUndeployed(final ZooKeeperClient client,
                                                       final RollingUpdateOpFactory opFactory,
                                                       final DeploymentGroup deploymentGroup,
                                                       final String host) {
    // Yield until the job is undeployed on the given host, or the rollout times out.
    // Once the job is fully undeployed "task status" path for it, i.e.
    // /status/hosts/<host>/job/<job-id>
    final TaskStatus taskStatus = getTaskStatus(client, host, deploymentGroup.getJobId());
    if (taskStatus == null) {
      // The job is undeployed!
      return opFactory.nextTask();
    } else if (isRolloutTimedOut(client, deploymentGroup)) {
        return opFactory.error("timed out while retrieving job status", host,
                               RollingUpdateError.TIMED_OUT_WAITING_FOR_JOB_TO_UNDEPLOY);
    } else {
      return opFactory.yield();
    }
  }
```

(The `DeploymentGroupTest` system tests pass with this change)

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -893,42 +940,176 @@ private RollingUpdateOp rollingUpdateDeploy(final ZooKeeperClient client,
...
+      */
+      return opFactory.nextTask(ImmutableList.of(
+          check(Paths.statusDeploymentGroupRemovedHosts(deploymentGroup.getName()), version),
+          set(Paths.statusDeploymentGroupRemovedHosts(deploymentGroup.getName()),
+              Json.asBytes(hostsToUndeploy))));
+    } catch (KeeperException | IOException e) {
+      // We can't read the list of hosts to undeploy. Move on.
+      log.error("unable to mark host undeployed after removal from deployment group: "
+                + "deployment-group={}, host={}", deploymentGroup.getName(), host, e);
+      return opFactory.nextTask();
```
gimaker: Do you expected this to happen? If not, maybe make it an `opFactory.error()` instead?

It's not obvious to me that carrying on like normal in case of errors results in correct behaviour.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -893,42 +940,176 @@ private RollingUpdateOp rollingUpdateDeploy(final ZooKeeperClient client,
...
+                                                     final DeploymentGroup deploymentGroup,
+                                                     final String host) {
+    return rollingUpdateUndeploy(client, opFactory, deploymentGroup, host, true);
+  }
+
+  private RollingUpdateOp rollingUpdateUndeploy(final ZooKeeperClient client,
+                                                final RollingUpdateOpFactory opFactory,
+                                                final DeploymentGroup deploymentGroup,
+                                                final String host,
+                                                final boolean force) {
```
gimaker: What does `force` do? (add javadoc plz)

helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/Paths.java
```diff
@@ -29,6 +29,7 @@
   private static final String JOBS = "jobs";
   private static final String JOBREFS = "jobrefs";
   private static final String HOSTS = "hosts";
+  private static final String REMOVED_HOSTS = "removed";
```
gimaker: I find it quite hard to understand what these node store, at a glance. Is it all removed hosts, as the name suggests?
Or is it actually `hosts-pending-undeploy-of-dg-job`?

Maybe document what is stored in them, or name it to reflect the contents better?

helios-client/src/main/java/com/spotify/helios/common/descriptors/RolloutTask.java
```diff
@@ -17,18 +17,21 @@
...
-import static com.google.common.base.Preconditions.checkNotNull;
-
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class RolloutTask extends Descriptor {
 
   public enum Action {
     UNDEPLOY_OLD_JOBS,
     DEPLOY_NEW_JOB,
     AWAIT_RUNNING,
+    FORCE_UNDEPLOY_JOBS,
```
negz: If you're actually curious I attempted to explain it in https://github.com/spotify/helios/pull/986/commits, and by answering @davidxia's question on this PR. Point taken regarding documenting it in code though.

helios-client/src/main/java/com/spotify/helios/client/Response.java
```diff
@@ -17,12 +17,12 @@
 
 package com.spotify.helios.client;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
```
negz: Sorry. I would appreciate any tips on making IntelliJ less crap at managing imports. I'll disable my import-optimisation for now.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -893,42 +938,176 @@ private RollingUpdateOp rollingUpdateDeploy(final ZooKeeperClient client,
...
+    // because a migration was requested and it is of the deployment group's job.
+    if (deploymentGroup.getRolloutOptions().getMigrate() &&
+        deployment.getJobId().equals(deploymentGroup.getJobId())) {
+      return true;
+    }
+
+    return false;
+  }
+
+  private boolean redundantDeployment(final Deployment deployment,
```
negz: Having renamed this method to `redundantUndeployment` and added a docstring I strongly feel my current implementation is clearer than this suggestion. I think it makes it more obvious *why* we're doing these checks.

As someone new-ish to the codebase I had a really hard time parsing the existing code and what its purpose was. It would be obvious to me from your proposed docstring *what* the desired state was, but not *why*.

helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/Paths.java
negz: (Which will make it the only commented path amidst an entire file of paths that could really use explanatory comments.)

helios-client/src/main/java/com/spotify/helios/common/descriptors/RolloutTask.java
```diff
@@ -17,18 +17,21 @@
...
+    AWAIT_STOPPED,
```
negz: Done.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: Added javadoc and also renamed the var to something more self explanatory.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -893,42 +938,176 @@ private RollingUpdateOp rollingUpdateDeploy(final ZooKeeperClient client,
...
-        } catch (HostNotFoundException e) {
-          return opFactory.error(e, host, RollingUpdateError.HOST_NOT_FOUND);
-        } catch (JobNotDeployedException e) {
-          // probably somebody beat us to the punch of undeploying. that's fine.
-        }
+      if (!ownedByDeploymentGroup(deployment, deploymentGroup)) {
+        continue;
+      }
+
+      if (!force && redundantDeployment(deployment, deploymentGroup)) {
```

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -893,42 +938,176 @@ private RollingUpdateOp rollingUpdateDeploy(final ZooKeeperClient client,
...
+        return opFactory.error(e, host, RollingUpdateError.HOST_NOT_FOUND);
+      } catch (JobNotDeployedException e) {
+        // probably somebody beat us to the punch of undeploying. that's fine.
       }
     }
 
     return opFactory.nextTask(operations);
   }
 
+  private boolean ownedByDeploymentGroup(final Deployment deployment,
```
gimaker: :ok_hand:

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: Done. Nice catch.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
negz: Good point. I don't expect this to happen. The `KeeperError` check is for the `getNode`, and it shouldn't be possible to hit this code without `Paths.statusDeploymentGroupRemovedHosts()` having been written either at hosts changed time or deployment group creation time. I'll make it an error.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java

negz:
I believe I have addressed all concerns. 🎉 I'm going to wait until after the long weekend to merge this.

PR #988, opened 2016-09-22
Use javax.ws.rs version 2

davidxia:
Use javax.ws.rs-api instead of jsr311-api.


mattnworb:
why?

davidxia:
I'm just testing this out for Flemstrom. It conflicts with barge deps unless you use the newer library.

PR #989, opened 2016-09-24
Refactored Long Method

feroozkhanchintu:
Long Method was refactored

### DIFF
--- a/helios-tools/src/main/java/com/spotify/helios/cli/command/HostListCommand.java
+++ b/helios-tools/src/main/java/com/spotify/helios/cli/command/HostListCommand.java
@@ -78,8 +78,8 @@ public class HostListCommand extends ControlCommand {
 	public HostListCommand(final Subparser parser) {
 		super(parser);
 
-		Collection<String> statusChoices = Collections2.transform(
-				Arrays.asList(HostStatus.Status.values()), new Function<HostStatus.Status, String>() {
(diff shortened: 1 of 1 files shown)


helios-tools/src/main/java/com/spotify/helios/cli/command/HostListCommand.java
```diff
@@ -62,206 +51,233 @@
...
+		labelsArg = parser.addArgument("-l", "--labels").action(append()).setDefault(new ArrayList<String>()).nargs("+")
+				.help("Only include hosts that match all of these labels. Labels need to be in the format "
+						+ "key=value.");
+	}
+
+	@Override
+	int run(final Namespace options, final HeliosClient client, final PrintStream out, final boolean json,
+			final BufferedReader stdin) throws ExecutionException, InterruptedException {
+		final String pattern = options.getString(patternArg.getDest());
+		final List<String> hosts = FluentIterable.from(client.listHosts().get()).filter(containsPattern(pattern))
```
davidxia: @feroozkhanchintu Thanks for your PR.

Always calling `HeliosClient.listHosts()` might slow down this command if there are a lot of hosts but only a few hosts that match a specified pattern or selector. `HeliosClient.listHosts(pattern, selectors)` makes the server filter and results in less data being transmitted over the wire. (The ZK operation is still to get all the child nodes of `/config/hosts` though.)

I'm also a fan of keeping as much logic on the server side as possible.

PR #992, opened 2016-10-11
master: add MemoryUsageMetricSet and GarbageCollectorMetricSet

mattnworb:
It would be nice to keep these stats and have them emitted via the
metrics reporter. This will help for tracking heap size (of different
generations), time spent in GC, etc etc.


mattnworb:
@rohansingh @dflemstr @negz @davidxia

rohansingh:
Not sure if rearranging all those imports was entirely necessary but 👍

PR #995, opened 2016-10-12
Ensure removed hosts ZK node exists as an empty list

negz:
Reviewer: @mattnworb

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
@@ -530,7 +530,8 @@ public class ZooKeeperMasterModel implements MasterModel {
 
       // statusDeploymentGroupRemovedHosts may not exist for deployment groups created before it was
       // introduced.
-      client.ensurePath(Paths.statusDeploymentGroupRemovedHosts(groupName));
+      client.ensurePathAndSetData(Paths.statusDeploymentGroupRemovedHosts(groupName),
+                                  Json.asBytesUnchecked(emptyList()));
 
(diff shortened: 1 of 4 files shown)


helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -531,6 +531,8 @@ public void updateDeploymentGroupHosts(final String groupName, final List<String
       // statusDeploymentGroupRemovedHosts may not exist for deployment groups created before it was
       // introduced.
       client.ensurePath(Paths.statusDeploymentGroupRemovedHosts(groupName));
+      client.ensurePathAndSetData(Paths.statusDeploymentGroupRemovedHosts(groupName),
```
mattnworb: I think if the path is created on the line above, then this call will be a no-op and you won't write the `[]`

negz: Oh yeah. That line was supposed to go away. Fixed.

helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/ZooKeeperClient.java
```diff
@@ -45,6 +45,8 @@
 
   void ensurePath(String path, boolean excludingLast) throws KeeperException;
 
+  void ensurePathAndSetData(String path, byte[] data) throws KeeperException;
```
davidxia: Add a javadoc for this to differentiate it from the method above?

negz: Done.

PR #996, opened 2016-10-13
Don't throw exceptions when we're beaten to updating a DG's hosts

negz:
It's not an error condition, but spams the error logs.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
@@ -92,6 +92,7 @@ import com.google.common.collect.Ordering;
 import com.google.common.collect.Sets;
 
 import org.apache.zookeeper.KeeperException;
+import org.apache.zookeeper.KeeperException.BadVersionException;
 import org.apache.zookeeper.KeeperException.NoNodeException;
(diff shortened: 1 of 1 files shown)


helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -579,6 +580,11 @@ public void updateDeploymentGroupHosts(final String groupName, final List<String
 
       client.transaction(ops);
       emitEvents(DEPLOYMENT_GROUP_EVENTS_KAFKA_TOPIC, events);
+    } catch (BadVersionException e) {
+      // some other master beat us in processing this host update. not exceptional.
+      // ideally we would check the path in the exception, but curator doesn't provide a path
+      // for exceptions thrown as part of a transaction.
+      log.debug("error saving rolling updateDeploymentGroupHosts operations: {}", e);
```
mattnworb: a debug log will never really be seen - since we log at info above that we are starting some transaction with a list of operations, it would be good to have a matching log entry also at info about how the transaction was beaten / rolled back / whatever the zookeeper term is

negz: Sure. I just copied the existing pattern from `rollingUpdateStep()`. I'll log at `info` in both places.

PR #997, opened 2016-10-13
Bump minor version to 0.10.0!

negz:
Reviewers: @rohansingh 

Now with shiny new revision controlled release pipeline.

Generated with `for f in $(find . -name pom.xml); do sed -i bak 's/>0.9.0-/>0.10.0-/g' $f;done` because I can never remember how to use `find -exec`.


davidxia:
I guess not needed anymore. Close?

PR #999, opened 2016-10-20
Create SetDataAndVersion class and use in ZooKeeperMasterModel

davidxia:
So we don't need to call both CheckWithVersion and SetData in a transaction.
Just call SetData in a transaction.


davidxia:
Just a test

PR #1000, opened 2016-10-21
Use spotify/alpine instead of onescience/alpine

rohansingh:
The upstream image was renamed.


davidxia:
@rohansingh Is this OK now? Did you push a working image to docker hub?

davidxia:
I heard the renaming has been reverted, but this PR would still be good to have to prevent future things like this.

rohansingh:
@davidxia The image that is currently at `spotify/alpine` is just a manual retag+push of the `1science/alpine:3.3` image. We should change it to be an automated build of spotify/docker-alpine.

PR #1005, opened 2016-10-26
Use spotify upstream image instead of jplock/zookeeper

davidxia:


davidxia:
@mattnworb @rohansingh

PR #1006, opened 2016-10-26
Upgrade circleCI to use docker 1.10.0

davidxia:
so that we test on a docker version closer to 1.12
which is what we use to build and release.


mattnworb:
👍  if it works

davidxia:
doesn't seem to work

PR #1008, opened 2016-10-27
remove misleading log line in DeadAgentReaper

mattnworb:
The log statement reads as if it has already been decided that the agent
with the given name is down, but this log line is occuring before the
check to see if the agent is up or down.

Since we log something at INFO when reaping the dead agent, remove this
line altogether to reduce noise or confusion.


mattnworb:
@davidxia

davidxia:
I originally had this line just to check the master was running this reaper at all, but I won't miss it too much I suppose.

PR #1010, opened 2016-10-27
Make toString() better for various classes

davidxia:
The class name at the end is useless.

Before:

`RolloutTask{action=UNDEPLOY_OLD_JOBS,
target='gew1-accesspoint-a-r915.gew1.spotify.net'}
com.spotify.helios.common.descriptors.RolloutTask@8d6f646c`

After:
`RolloutTask{action=UNDEPLOY_OLD_JOBS,
target='gew1-accesspoint-a-r915.gew1.spotify.net'}`

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/Deployment.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/Deployment.java
@@ -127,7 +127,7 @@ public class Deployment extends Descriptor {
            ", deployerUser='" + deployerUser + '\'' +
            ", deployerMaster='" + deployerMaster + '\'' +
            ", deploymentGroupName='" + deploymentGroupName + '\'' +
-           "} " + super.toString();
+           '}';
   }
 
   @Override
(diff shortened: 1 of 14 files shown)


davidxia:
@mattnworb

helios-client/src/main/java/com/spotify/helios/common/descriptors/Descriptor.java
```diff
@@ -49,4 +49,9 @@ public String toJsonString() {
       throws IOException {
     return parse(value.getBytes(UTF_8), clazz);
   }
+
+  @Override
+  public String toString() {
+    return "Descriptor{}";
+  }
```
mattnworb: since this is an abstract class, I don't think there is any value in providing this default implementation.

For instance if someone adds a subclass and forgets to override `toString()`, it would make more sense for `SomeNewClass.toString()` to return `com.spotify.helios.foo.SomeNewClass@123456` or whatever the default is rather than `Descriptior{}`.

PR #1012, opened 2016-10-27
add retries to flaky test HeliosSoloIT.testServiceDiscovery

mattnworb:
The test can fail because it tries to connect immediately to the socket of
the alpine container. Need to give the `apk install` being run as the
command of the container time to finish.

### DIFF
--- a/helios-integration-tests/src/test/java/com/spotify/helios/HeliosSoloIT.java
+++ b/helios-integration-tests/src/test/java/com/spotify/helios/HeliosSoloIT.java
@@ -33,12 +33,15 @@ import com.spotify.helios.testing.TemporaryJobs;
 import com.google.common.net.HostAndPort;
 
 import org.apache.commons.io.IOUtils;
+import org.awaitility.Awaitility;
 import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 
 import java.net.Socket;
+import java.util.concurrent.Callable;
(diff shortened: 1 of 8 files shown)


mattnworb:
@spotify/helios-team

helios-integration-tests/src/test/java/com/spotify/helios/HeliosSoloIT.java
```diff
@@ -114,10 +118,15 @@ public void testServiceDiscovery() throws Exception {
 
     // Connect to alpine container to get the curl response. If we get back the nginx welcome page
     // we know that helios properly registered the nginx service in SkyDNS.
-    try (final Socket s = new Socket(alpineAddress.getHostText(), alpineAddress.getPort())) {
-      final String result = IOUtils.toString(s.getInputStream()).trim();
-      assertThat(result, containsString("Welcome to nginx!"));
-    }
+    final Callable<String> curlResponse = () -> {
```
davidxia: Perhaps just call this var `response` since it's not curl?

davidxia:
:100: if it works

helios-integration-tests/src/test/java/com/spotify/helios/HeliosSoloIT.java
mattnworb: i was trying to use a verb, but I'll find a better one :)

mattnworb:
The test failure is a different flaky test so I'll merge this fix.

PR #1013, opened 2016-10-28
add additional logging around HOSTS_CHANGED rolling-update

mattnworb:
This is intended to help troubleshoot issues related to unexpected
rollout tasks being generated when hosts are removed from a deployment
group.


mattnworb:
@davidxia

mattnworb:
This is a bit of a hail mary to help troubleshoot a bug report where the wrong set of tasks are generated when a host is removed from the group. I think the problem is coming from the combination of which hosts are derived in the `removedHosts` list from the inputs of `(curHosts, hosts, previouslyRemovedHosts)` but with the logging we have today it is hard to definitively say.

davidxia:
👍

PR #1014, opened 2016-10-28
RFC: Add 'secret volume' integration.

negz:
# Overview

Allow Helios jobs to be configured to request a 'secret volume' be created containing the specific secrets for that job. Currently this is implemented by calling out to a sidecar service that actually procures and stores the secrets. The API of said sidecar service is explained [here](https://github.com/negz/secret-volume).
# The status quo

Currently at Spotify we typically run a script to procure secrets for a host from [Talos](https://github.com/spotify/talos) and then expose the directory containing those secrets by adding a bind volume to the relevant Helios job, i.e.:

``` json
"volumes" : {
  "/container/secrets:ro" : "/host/secrets"
}
```

Talos is a wrapper around Puppet's Hiera, and is typically used to store the secrets for a host, rather than a Helios job (i.e. container). Our client-side secrets procurement tooling follows this pattern, making it awkward to procure a different set of secrets per Helios job. This means secrets management tends to be one of the reasons we only deploy one container per host at Spotify, even if we might like to, for example, deploy a service in one container and supporting software like [ffwd](https://github.com/spotify/ffwd) in another with a different set of secrets exposed to each container.
# Proposal

This RFC proposes that Helios be able to procure secrets for each job that it runs. The actual job of downloading and storing the secrets is offloaded to a sidecar application. `secret-volume` is a reference implementation of such an application.

The Helios agent makes a 'create' API call to `secret-volume` specifying a unique identifier, a secret source like Talos, and a keypair to authenticate to said source. `secret-volume` authenticates to the secret source on behalf of a Helios job and stores its secrets in a `tmpfs` filesystem mounted at `/secrets/:id` on the filesystem of the Helios agent's host.

To make use of this functionality it must be enabled in both the Helios agent and Helios job. To enable it for the Helios agent first run `secret-volume` on the agent's host, then point the Helios agent at it:

```
$ export ADDR="127.0.0.1:10002"
$ export SECRETS="/secrets"
$ ./secret-volume --addr=${ADDR} --parent=${SECRETS} --talos-srv="_talos._tcp.example.org"
$ java com.spotify.helios.agent.AgentMain --secret-volume-manager=${ADDR} --secret-volume-path=${SECRETS} 
```

Once `secret-volume` and the Agent are running each job needs to be configured to enable secrets procurement. This is done by adding a clause to the job config:

``` json
"secrets" {
  "source": "Talos",
  "path": "/secrets",
  "keypairFromHost": {
    "certificate": "/path/to/pem/cert/on/host",
    "privateKey": "/path/to/pem/key/on/host"
  }
}
```

The above clause requests that Helios authenticate to Talos using the provided keypair on the Agent's host. The secrets will be automatically exposed inside the container at the path `/secrets`. Helios requests `secret-volume` procure and store secrets each time the `TaskRunner` tries to start a Docker container for a job, and requests the secrets be destroyed when the `TaskRunner` stops said Docker container.
# The future...

The aforementioned `keyPairFromHost` directive is included for compatibility with the status quo at Spotify. If rolled out at Spotify today this change would still use the Puppet host certificate to authenticate to Talos for each container - i.e. it would be functionally the same as the status quo - but it opens up the possibility to decouple authentication to Talos from the host's Puppet certificate, or to procure secrets from a different source altogether.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/SecretVolumeRequest.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/SecretVolumeRequest.java
@@ -1,0 +1,172 @@
+/*
+ * Copyright (c) 2014 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.helios.common.descriptors;
+
+import static java.util.Collections.emptyMap;
+
+import com.spotify.helios.common.SecretVolumeException;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Represents a secret volume request to secret-volume, i.e. a secret-volume response with an added
+ * KeyPair.
+ *
+ * <pre>
+ *   {
+ *     "ID": "containerId",
+ *     "Source": "Talos",
+ *     "Tags": {},
+ *     "KeyPair": {
+ *        "Certificate": "PEM encoded cert",
+ *        "PrivateKey": "Pem encoded private key"
+ *     }
+ *   }
+ * </pre>
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class SecretVolumeRequest extends SecretVolume {
+
+  private final KeyPair keyPair;
+
+  public SecretVolumeRequest(
+      @JsonProperty("ID") final String id,
+      @JsonProperty("Source") final Secrets.Source source,
+      @JsonProperty("Tags") final Map<String, List<String>> tags,
+      @JsonProperty("KeyPair") final KeyPair keyPair) {
+    super(id, source, tags);
+    this.keyPair = keyPair;
+  }
+
+  public static SecretVolumeRequest create(final String id, final Secrets.Source source,
+                                           final String certData, final String keyData) {
+    // TODO(negz): Read tags from config, or agent labels?
+    final Map<String, List<String>> tags = emptyMap();
+    return new SecretVolumeRequest(id, source, tags, new KeyPair(certData, keyData));
+  }
+
+  private static String readToString(final String path) throws SecretVolumeException {
+    try {
+      return new String(Files.readAllBytes(Paths.get(path)), Charset.defaultCharset());
+    } catch (IOException e) {
+      throw new SecretVolumeException("cannot read secret volume keypair", e);
+    }
+  }
+
+  public KeyPair getKeyPair() {
+    return this.keyPair;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    final SecretVolumeRequest that = (SecretVolumeRequest) o;
+
+    return Objects.equals(this.id, that.id)
+           && Objects.equals(this.source, that.source)
+           && Objects.equals(this.tags, that.tags)
+           && Objects.equals(this.keyPair, that.keyPair);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(id, source, tags, keyPair);
+  }
+
+  @Override
+  public String toString() {
+    return "Secrets{"
+           + "source=" + source
+           + ", id=" + id
+           + ", tags=" + tags
+           + ", keyPair=" + keyPair
+           + "} " + super.toString();
+  }
+
+  public static class KeyPair {
+    private final String certificate;
+    private final String privateKey;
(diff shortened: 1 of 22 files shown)


helios-services/src/main/java/com/spotify/helios/agent/TaskRunner.java
```diff
@@ -249,20 +261,40 @@ private String createAndStartContainer()
...
+
+    final ImmutableList.Builder<String> binds = ImmutableList.builder();
+    if (hostConfigBuilder.binds() != null) {
+      binds.addAll(hostConfigBuilder.binds());
+    }
+
+    if (secretVolumeManager.isPresent() && config.secretsConfig() != null) {
+      log.info("Secret volume manager enabled for {}", config.containerName());
+      final String source = secretVolumeManager.get()
+          .create(config.containerName(), config.secretsConfig());
```
negz: My understanding is that if this call fails `startContainer()` will throw `SecretVolumeException` and the Supervisor will retry.

There's an open question as to what to do if this call fails due to secrets already existing for the requested container name. Currently `secret-volume` will return a 500 and thus this would fail forever until the secrets were manually destroyed (either via the `secret-volume` API or just unmounting them manually).

`secret-volume` could, instead of returning 500, either:
* Just return 200 OK, thus causing whatever secrets exist at the path to be exposed. This could be insecure and lead to stale secret data being exposed.
* Internally destroy the existing secrets at the requested path and attempt to replace them with newly procured secrets.

mattnworb:
I like that the PR is phrased as a proposal that we can evaluate, I think that makes the most sense to review and comment on before looking at the code that implements it.

Could you also write up the proposal as a new document in the `docs` directory so that this feature is documented from day one whenever it is merged? I imagine most of the text could be copy-and-pasted (with just some verb tense changes perhaps).

mattnworb:
A few initial questions:

1. Does the author of a Job really need to know or care about where the secrets come from? Would it be enough for them to just state in their job definition what path they want the secrets to be provided under? This would greatly simplify what the Helios users sees and needs to care about.

For example, consider how service registration works - the job author simply specifies a `"registration"` element as a series of port numbers and port mappings, and doesn't care about the actual mechanism used. Service registration is then provided as a plugin to the agent, so that we can support multiple types of registration (per Helios-installation).

I'd be in favor of making what is in the job definition be as concise as possible and having the implementation be specified just within the agent.

2. What happens if the job refers to a secrets source that the agent doesn't recognize?

3. What happens if the job refers to keypair files that don't exist on the host?

4. What would change with the proposed `"secrets"` element if a provider other than Talos is used?

helios-client/src/main/java/com/spotify/helios/common/descriptors/Secrets.java
```diff
@@ -0,0 +1,179 @@
...
+           + "source=" + source
+           + ", path=" + path
+           + ", keyPairFromHost=" + keyPairFromHost
+           + "} " + super.toString();
+  }
+
+  public enum Source {
+    UNKNOWN,
+    TALOS
+  }
```
mattnworb: if all possible names for the sources are defined within this class, then this means that the only secret providers that could be used have to be defined in the helios source code. in other words it is not very open for extension or pluggable. I think the plugin approach used by the service registration code would work great here to avoid tight coupling.

helios-services/src/main/java/com/spotify/helios/agent/AgentParser.java
```diff
@@ -272,6 +281,13 @@ protected void addArgs(final ArgumentParser parser) {
         .action(storeTrue())
         .setDefault(false)
         .help("If specified, the agent won't write job histories to ZooKeeper.");
+
+    secretVolumeManagerArg = parser.addArgument("--secret-volume-manager")
+        .help("Enable automatic procurement of secrets using this secret volume manager.");
```
mattnworb: it's not clear from this help text that this is meant to be a `host:port` string

helios-client/src/main/java/com/spotify/helios/common/descriptors/SecretVolumeRequest.java
```diff
@@ -0,0 +1,172 @@
...
+  public static SecretVolumeRequest create(final String id, final Secrets.Source source,
+                                           final String certData, final String keyData) {
+    // TODO(negz): Read tags from config, or agent labels?
+    final Map<String, List<String>> tags = emptyMap();
+    return new SecretVolumeRequest(id, source, tags, new KeyPair(certData, keyData));
+  }
+
+  private static String readToString(final String path) throws SecretVolumeException {
+    try {
+      return new String(Files.readAllBytes(Paths.get(path)), Charset.defaultCharset());
```
mattnworb: this seems to be the same exact code as `class FsSmallFileReader` - why does it need to be a static method too? 

I don't think this method is used in the class

helios-services/src/main/java/com/spotify/helios/agent/DefaultSecretVolumeManager.java
```diff
@@ -0,0 +1,142 @@
...
+        .build()
+        .toString();
+  }
+
+  @Override
+  public String create(final String id, final Secrets secretsCfg) throws SecretVolumeException {
+    log.debug("Creating secret volume id {}", id);
+    final SecretVolumeRequest svr = secretVolumeRequest(id, secretsCfg);
+
+    final HttpUrl url = HttpUrl.parse(baseUrl)
```
mattnworb: it would make more sense to keep `baseUrl` as the original `HttpUrl`, so that you don't keep re-parsing it on each request.

i.e. you take an InetSocketAddress and turn it into a HttpUrl then turn that into a String, just to turn that back into a HttpUrl here

negz:
@mattnworb I think it's a good idea to discuss this design, but I suspect a Google Doc might work better for that than top level PR comments. Let's use Github for code/implementation comments.

https://docs.google.com/document/d/14gnjg_qYHWg2wFMIRe79jRreWKRfK5GNnfYZIVvxNzY/edit#

I've created the above document (currently open to Spotify employees only) to discuss this change. I'll respond to your design questions there.

davidxia:
Is anyone still working on this?

mattnworb:
I don't think we are likely to pick this up again

PR #1016, opened 2016-10-31
Fix docker tag cmd in legacy helios-use script

davidxia:


davidxia:
@lepistone @rohansingh

lepistone:
Thanks, nice! If we keep this check for docker < 1.12, I think it deserves a comment to say what's going on. (and TIL how to parse strings in bash).

PR #1017, opened 2016-10-31
Prevent accidental undeploys by deployment groups

davidxia:
Fix `ZooKeeperMasterModel.getInitRollingUpdateOps()` so that if both
`updateHosts` and `undeployHosts` have host Strings in common, we don't
generate unnecessary rollout tasks that undeploy from a host.

Remove hosts that are in `updateHosts` from `undeployHosts`.

Just to be safe, we list the undeploy tasks first as well.


davidxia:
@mattnworb I fixed the relevant test. See last commit.

PR #1018, opened 2016-10-31
fix AssertionError in InMemoryLogStreamFollower

mattnworb:
docker-client's LogMessage class always returns the message content as
a read-only buffer (via `ByteBuffer.asReadOnlyBuffer()`).

ByteBuffer's `hasArray()` method will return false if the buffer is
read-only, which always trips the assertion in
InMemoryLogStreamFollower.

Add logic to deal with buffer's where `hasArray()` returns false by
copying the buffer's contents to a temporary array. This change leaves
in the original access of `ByteBuffer.array()` when `hasArray()` returns
true so that it can work with either type of ByteBuffer, to be immune to
any possible future changes in docker-client.

This also fixes a failure in HeliosSoloIt#testServiceDiscovery


mattnworb:
@davidxia @dflemstr @spotify/helios-team

davidxia:
💯

PR #1021, opened 2016-11-03
try to fix flaky testRemovingAgentTagUndeploysJob

mattnworb:
attempting to fix flakiness of testRemovingAgentTagUndeploysJob by
adjusting the logic of awaitHostStatusWithLabels in SystemTestBase

```
testRemovingAgentTagUndeploysJob - com.spotify.helios.system.DeploymentGroupTest
java.lang.AssertionError:
Expected: map containing ["another"->"label"]
     but: map was [<foo=bar>]
at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
at org.junit.Assert.assertThat(Assert.java:956)
at org.junit.Assert.assertThat(Assert.java:923)
at com.spotify.helios.system.DeploymentGroupTest.awaitUpWithLabel(DeploymentGroupTest.java:188)
at com.spotify.helios.system.DeploymentGroupTest.testRemovingAgentTagUndeploysJob(DeploymentGroupTest.java:243)
```

I am not positive if I actually changed anything of substance, besides adding better failure reasons when test assertions fail - but this commit has passed 3 of 4 runs in CircleCI (one timed out).

### DIFF
--- a/helios-system-tests/src/main/java/com/spotify/helios/system/SystemTestBase.java
+++ b/helios-system-tests/src/main/java/com/spotify/helios/system/SystemTestBase.java
@@ -39,6 +39,7 @@ import static java.util.concurrent.TimeUnit.MINUTES;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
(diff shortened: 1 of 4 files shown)


helios-system-tests/src/main/java/com/spotify/helios/system/SystemTestBase.java
```diff
@@ -922,25 +922,26 @@ public HostStatus call() throws Exception {
...
+          // labels are stored in ZK after the host has come up
+          || candidate.getLabels().size() == 0) {
+
+        return null;
       }
+      return candidate;
     });
+
+    assertThat("host " + host + " has status=" + status + " with labels=" + hostStatus.getLabels(),
+        hostStatus.getLabels(), is(labels));
```
mattnworb: moved the assertion into this method rather than assert something like this in each callsite as it was previously

PR #1022, opened 2016-11-03
Add option to set default report writer

mzeo:
Will fix issue #883

PR #1023, opened 2016-11-04
Fix for pull request #1022, set default report writer

mzeo:
ReportWriter does not have any public constructors, so
it was impossible to use the set default report writer option.

This adds protected constructor and protected writeEvent
method so ReportWriter can be inherited.


mzeo:
@mattnworb I made a another pull request that is more inline with how I think it should work instead of these minimal changes: https://github.com/spotify/helios/pull/1026/files

mzeo:
Superseded by https://github.com/spotify/helios/pull/1026

PR #1024, opened 2016-11-07
Add EventSender interface and GooglePubSubSender

zalenski:
@lndbrg @spotify/helios-team

### DIFF
--- a/docs/event_senders.md
+++ b/docs/event_senders.md
@@ -1,0 +1,30 @@
+Helios Event Senders
+===
+
+There is support to send events on Helios cluster updates.
+
+From helios masters, the following events are sent:
+
+Topic `HeliosDeploymentGroupEvents` with a `DeploymentGroupEvent` message,
+see `com.spotify.helios.rollingupdate.DeploymentGroupEventFactory`.
+
+From helios agents:
+
+Topic `HeliosTaskStatusEvents` with a `HeliosTaskStatusEvent` message,
+see `com.spotify.helios.common.descriptors.TaskStatusEvent`.
+
+
+There is support to send events over Kafka, and over Google PubSub.
+
+
+Kafka
+---
+
+Using the command line argument `--kafka <broker>`, Kafka publishing is enabled. The Kafka sender creates a KafkaRecord from the
+topic and message as is.
+
+GooglePubSub
+---
+
+Using the command line argument `--pubsub-topic-prefix <topic-prefix>`, Google PubSub publishing is enabled. The Google PubSub sender
+concatenates the topic prefix with the the helios event topic (with no separator), and sends the message as payload.
--- a/helios-services/pom.xml
+++ b/helios-services/pom.xml
@@ -231,6 +231,21 @@
       <version>0.8.2.1</version>
     </dependency>
     <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-pubsub</artifactId>
+      <version>0.5.1</version>
+      <exclusions>
+        <exclusion>
+          <groupId>javax.servlet</groupId>
+          <artifactId>servlet-api</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>com.google.protobuf</groupId>
+          <artifactId>protobuf-java</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
       <groupId>ch.qos.logback</groupId>
       <artifactId>logback-classic</artifactId>
     </dependency>
--- a/helios-services/pom.xml
+++ b/helios-services/pom.xml
@@ -239,10 +239,6 @@
           <groupId>javax.servlet</groupId>
           <artifactId>servlet-api</artifactId>
         </exclusion>
-        <exclusion>
-          <groupId>com.google.protobuf</groupId>
-          <artifactId>protobuf-java</artifactId>
-        </exclusion>
       </exclusions>
     </dependency>
     <dependency>
@@ -365,6 +361,17 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+
+  <dependencyManagement>
+    <dependencies>
+      <!-- google pubsub client requires protobuf v3 -->
+      <dependency>
+        <groupId>com.google.protobuf</groupId>
+        <artifactId>protobuf-java</artifactId>
+        <version>3.0.0</version>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
 
   <build>
     <plugins>
--- a/helios-services/src/main/java/com/spotify/helios/agent/ZooKeeperAgentModel.java
+++ b/helios-services/src/main/java/com/spotify/helios/agent/ZooKeeperAgentModel.java
@@ -26,9 +26,7 @@ import com.spotify.helios.common.descriptors.JobId;
 import com.spotify.helios.common.descriptors.Task;
 import com.spotify.helios.common.descriptors.TaskStatus;
 import com.spotify.helios.common.descriptors.TaskStatusEvent;
-import com.spotify.helios.servicescommon.KafkaClientProvider;
-import com.spotify.helios.servicescommon.KafkaRecord;
-import com.spotify.helios.servicescommon.KafkaSender;
+import com.spotify.helios.servicescommon.EventSender;
 import com.spotify.helios.servicescommon.coordination.Paths;
 import com.spotify.helios.servicescommon.coordination.PersistentPathChildrenCache;
 import com.spotify.helios.servicescommon.coordination.ZooKeeperClient;
@@ -41,6 +39,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.nio.file.Path;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CopyOnWriteArrayList;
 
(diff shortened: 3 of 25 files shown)


helios-services/src/main/java/com/spotify/helios/servicescommon/ServiceParser.java
```diff
@@ -185,6 +187,11 @@ public ServiceParser(final String programName, final String description, final S
         .setDefault(new ArrayList<String>())
         .help("Kafka brokers to bootstrap with");
 
+    pubsubArg = parser.addArgument("--pubsub")
```
lndbrg: maybe rename this arg to to something more logical?

lndbrg:
We should add this to the ZookeperAgentModel too @zalenski

helios-services/src/main/java/com/spotify/helios/master/MasterParser.java
```diff
@@ -86,6 +86,7 @@ public MasterParser(final String... args) throws ArgumentParserException {
         .setAdminEndpoint(adminAddress)
         .setHttpEndpoint(httpAddress)
         .setKafkaBrokers(getKafkaBrokers())
+        .setPubsubPrefixes(getPubsubOptions())
```
lndbrg: getPubsubOptions should probably be renamed to getPubsubPrefixes

zalenski:
> should add this to the ZookeperAgentModel too 

added to ZookeperAgentModel too

helios-services/src/main/java/com/spotify/helios/servicescommon/ServiceParser.java
```diff
@@ -185,6 +186,11 @@ public ServiceParser(final String programName, final String description, final S
         .setDefault(new ArrayList<String>())
         .help("Kafka brokers to bootstrap with");
 
+    pubsubTopicPrefixArg = parser.addArgument("--pubsub-topic-prefix")
+        .action(append())
+        .setDefault(new ArrayList<String>())
+        .help("Google pubsub topic prefixes to send to");
```
mattnworb: is the idea that the user will specify 0 or more prefixes like "foobar.helios." and then the helios agent or master will provide the rest of the topic name, like `"foobar.helios." + "HeliosTaskStatusEvents"`?

It would probably be helpful to document somewhere the full list of topics that the helios codebase will send messages to (when enabled)

helios-services/src/main/java/com/spotify/helios/servicescommon/ServiceParser.java
```diff
@@ -303,6 +309,11 @@ public String getZooKeeperAclAgentUser() {
     return kafkaBrokers.isEmpty() ? null : kafkaBrokers;
   }
 
+  public List<String> getPubsubPrefixes() {
+    final List<String> pubsubPrefixes = options.getList(pubsubTopicPrefixArg.getDest());
+    return pubsubPrefixes.isEmpty() ? null : pubsubPrefixes;
```
mattnworb: i see from above that this is following the existing pattern but it feels weird to me that this returns null rather than an empty list, especially when the provider.senders() method has logic to return an empty list when nothing is enabled.

helios-services/src/main/java/com/spotify/helios/master/MasterService.java
```diff
@@ -206,8 +214,15 @@ public MasterService(final MasterConfig config,
     // and wrap it in our KafkaSender.
     final KafkaSender kafkaSender = new KafkaSender(kafkaClientProvider.getDefaultProducer());
 
+    // Make GooglePubsub senders
+    final List<GooglePubSubSender> pubSubSenders = googlePubSubProvider.senders();
+
+    final List<EventSender> eventSenders = Stream.concat(
+        Stream.of(kafkaSender),
+        pubSubSenders.stream()).collect(Collectors.toList());
```
mattnworb: again, this follows the existing coding style, but I think it would make more sense if this was simply an empty list if neither the kakfa sender or pubsub sender were enabled - or a list of just one element if just one of them was enabled - rather than always having two instances which internally have to check if they are enabled or not.

(it is much simpler to simply never call a KafkaSender if kafka-sending is not enabled than to call a KafkaSender that has to wrap it's logic in `!= null` checks)

helios-services/src/main/java/com/spotify/helios/servicescommon/ServiceParser.java
zalenski: I'll change to empty list. It probably even breaks the code using it right now.

helios-services/src/main/java/com/spotify/helios/servicescommon/ServiceParser.java
zalenski: Yes, that would be the idea. I followed the current style of not documenting the topics… ;)

Is there a good place to put it? A new `.md` file in `docs/`?

mattnworb: > Is there a good place to put it? A new .md file in docs/?

that would be perfect, thanks

helios-services/src/main/java/com/spotify/helios/servicescommon/ServiceParser.java
zalenski: updated

helios-services/src/main/java/com/spotify/helios/servicescommon/ServiceParser.java
zalenski: Added a minimal `event_senders.md`

docs/event_senders.md
```diff
@@ -0,0 +1,28 @@
+Helios Event Senders
+===
```
mattnworb: awesome, thank you!

zalenski:
@mattnworb Do you recognise the error:
`java.lang.SecurityException: class "javax.servlet.Filter"'s signer information does not match signer information of other classes in the same package` ??

mattnworb:
@zalenski no that is new to me. I tried to check out your branch and run `mvn clean compile` and I don't see it. Can you post the full output to a gist maybe?

mattnworb:
@zalenski - it looks like com.google.cloud:google-cloud-pubsub:0.5.1 drags in a transitive dependency on javax.servlet:servlet-api:2.5 (via com.google.api-client:google-api-client-appengine, of all things - google-cloud-core seems to have a ton of auth-related dependencies). This causes the SecurityException because some of the servlet classes are loaded from org.eclipse.jetty.orbit:javax-servlet:3.0 that dropwizard depends on.

This goes away if you add an exclusion to the pubsub library:

```xml
    <dependency>
      <groupId>com.google.cloud</groupId>
      <artifactId>google-cloud-pubsub</artifactId>
      <version>0.5.1</version>
      <exclusions>
        <exclusion>
          <groupId>javax.servlet</groupId>
          <artifactId>servlet-api</artifactId>
        </exclusion>
      </exclusions>
    </dependency>
```

but then this causes a new error in those tests:

```
Caused by: java.lang.NoClassDefFoundError: com/google/protobuf/GeneratedMessageV3
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:361)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at com.google.pubsub.v1.PublisherGrpc.<clinit>(PublisherGrpc.java:41)
	at com.google.cloud.pubsub.spi.v1.PublisherSettings$Builder.<init>(PublisherSettings.java:462)
```

it seems that a depenency on riemann-java-client drags in a protobuf 2.5.0 library. You can probably exclude that as well as I do not think the riemann stuff is really even in use.

zalenski:
> can probably exclude that as well

@mattnworb yup, seems to work

zalenski:
`\o/` it builds

lndbrg:
:+1:

helios-services/pom.xml
```diff
@@ -231,6 +231,21 @@
...
+      <version>0.5.1</version>
+      <exclusions>
+        <exclusion>
+          <groupId>javax.servlet</groupId>
+          <artifactId>servlet-api</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>com.google.protobuf</groupId>
+          <artifactId>protobuf-java</artifactId>
+        </exclusion>
```
mattnworb: trying to run this locally and it looks to me like you actually need to add an explicit dependency on v3 of protobuf-java (or manage it via dependencyManagement), as whatever loading of pubsub classes occurs also loads the v3 protobuf classes 

`Caused by: java.lang.ClassNotFoundException: com.google.protobuf.GeneratedMessageV3`

davidxia:
I took a quick look and things look good. The design looks reasonable to me. Wondering if I'm not thinking of something obvious though.

helios-services/src/main/java/com/spotify/helios/agent/ZooKeeperAgentModel.java
```diff
@@ -167,7 +166,10 @@ public void setTaskStatus(final JobId jobId, final TaskStatus status)
       }
     }
     final TaskStatusEvent event = new TaskStatusEvent(status, System.currentTimeMillis(), agent);
-    kafkaSender.send(KafkaRecord.of(TaskStatusEvent.KAFKA_TOPIC, event.toJsonBytes()));
+    final byte[] message = event.toJsonBytes();
+    for (final EventSender sender : eventSenders) {
+      sender.send(TaskStatusEvent.KAFKA_TOPIC, message);
```
zalenski: done

helios-services/src/main/java/com/spotify/helios/servicescommon/GooglePubSubProvider.java
```diff
@@ -0,0 +1,55 @@
...
+public class GooglePubSubProvider {
+  private static final Logger log = LoggerFactory.getLogger(GooglePubSubProvider.class);
+
+  private final List<String> pubsubPrefixes;
+
+  public GooglePubSubProvider(final List<String> pubsubPrefixes) {
+    this.pubsubPrefixes = pubsubPrefixes;
+  }
+
+  public List<GooglePubSubSender> senders() {
```
zalenski: added some tests.
(`PubSubOptions.getService()` starts a real pubsub service straight away, so did a `senders(Supplier<PubSub> pubsubSupplier)` to test it)

lndbrg:
:+1:

mattnworb:
thanks again for all the changes and addressing the several rounds of feedback

zalenski:
`\o/` it's merged! :)

PR #1025, opened 2016-11-08
update dependencies to fix issues with google-cloud-pubsub

mattnworb:
the pubsub client requires v3 of protobuf and also uses methods in Guava that are only available after version 19 (such as`Futures.catching(..)`).

Note that it is possible that these changes break something else in the master, so let's hope the automated tests catch this (I have not yet run them myself).

### DIFF
--- a/helios-services/pom.xml
+++ b/helios-services/pom.xml
@@ -239,10 +239,6 @@
           <groupId>javax.servlet</groupId>
           <artifactId>servlet-api</artifactId>
         </exclusion>
-        <exclusion>
-          <groupId>com.google.protobuf</groupId>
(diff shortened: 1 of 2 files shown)


mattnworb:
@zalenski

lndbrg:
:+1:

mattnworb:
build is green, good thing we were only one version behind on guava

helios-services/pom.xml
```diff
@@ -366,6 +362,17 @@
     </dependency>
   </dependencies>
 
+  <dependencyManagement>
```
mattnworb: because the problematic dependency is only in this pom

PR #1026, opened 2016-11-11
Fix for pull request #1022, set default report writer

mzeo:
ReportWriter does not have any public constructors, so
it was impossible to use.

This changeset refactores TemporaryJobReports into one
abstract class which does not do any IO, and one TemporaryJobJsonReports
which is used by default.

TemporaryJobs builder was changed to allow setting TemporaryJobsReports
instead of just default writer for greater controll.

### DIFF
--- a/helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobJsonReports.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobJsonReports.java
@@ -46,7 +46,9 @@ public class TemporaryJobJsonReports extends TemporaryJobReports {
 
   public TemporaryJobJsonReports(final Path outputDir) {
     this.outputDir = outputDir;
-    this.outputDir.toFile().mkdirs();
+    if (!this.outputDir.toFile().mkdirs()) {
+      throw new IllegalStateException(format("failed to create directory \"%s\"", outputDir));
+    }
   }
 
   public ReportWriter getWriterForTest(final Description testDescription) {
--- a/helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobJsonReports.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobJsonReports.java
@@ -81,9 +81,8 @@ public class TemporaryJobJsonReports extends TemporaryJobReports {
         jg.writeStartArray();
       } catch (IOException e) {
         log.error("exception creating event log: {} - {}", logFile.getAbsolutePath(), e);
-      } finally {
-        this.jg = jg;
       }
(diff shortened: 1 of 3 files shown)


helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobJsonReports.java
```diff
@@ -0,0 +1,139 @@
...
+ */
+public class TemporaryJobJsonReports extends TemporaryJobReports {
+
+  private static final Logger log = LoggerFactory.getLogger(TemporaryJobJsonReports.class);
+
+  private final Path outputDir;
+
+  public TemporaryJobJsonReports(final Path outputDir) {
+    this.outputDir = outputDir;
+    this.outputDir.toFile().mkdirs();
```
mattnworb: might make sense to catch when this returns `false` and throw some sort of IllegalStateException

helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobJsonReports.java
```diff
@@ -0,0 +1,139 @@
...
+      final File logFile = outputDir.resolve(logFilename).toFile();
+
+      JsonGenerator jg = null;
+      try {
+        jg = new JsonFactory().createGenerator(logFile, JsonEncoding.UTF8);
+        jg.writeStartArray();
+      } catch (IOException e) {
+        log.error("exception creating event log: {} - {}", logFile.getAbsolutePath(), e);
+      } finally {
+        this.jg = jg;
```
mattnworb: I guess this comes from the original source, but fwiw this pattern with the finally looks strange to me. I think this could just be:

```java
try {
  this.jg = new JsonFactory().createGenerator(logFile, JsonEncoding.UTF8)
} catch (IOException e) {
  //log the exception, and set the final field
  this.jg = null;
}
```

or arguably this field doesn't need to be `final` if it can be null sometimes and set to an instance other times.

helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobJsonReports.java
mzeo: done

mzeo:
@mattnworb Seems like some pre test check failed.

mattnworb:
weird, looks like the settings in the CircleCI project got changed somehow. I've restored them and re-run the build. 

sorry for all the headaches!

mattnworb:
ok the test failures all look like random flukes now.

Thanks again for adding this!

mattnworb:
0.9.53 was released with this

PR #1027, opened 2016-11-19
Use Futures.catching, not Futures.withFallback

odsod:
Fixes a runtime incompatibility with Guava 20.0, which has removed
Futures.withFallback.


odsod:
Thanks @mattnworb!

PR #1030, opened 2016-12-01
disable GooglePubSubSender if pubsub cannot be reached

mattnworb:
Adds a `isHealthy()` method to the EventSender interface that is used to
filter out any unhealthy senders at startup.

This logic is done in a new class named EventSenderFactory, which
extracts the identical logic for constructing Lists of EventSenders from
the MasterService and the AgentService. To make this common class
possible, also extracted a class for the common configuration fields
between AgentConfig and MasterConfig. This new CommonConfiguration class
is incomplete - more fields can be moved to this class to avoid
duplication, but I left that for future commits.

I created EventSenderFactory in hopes of writing a test for the logic
that unhealthy senders are removed from the List, but the nature of this
class makes this test to impractical - since the
List-of-EventSender-building involves constructing new instances of
KafkaProvider/GooglePubSubProvider/etc and calling methods on instances
that those instances return, which is not really feasible to be mocked
out in a test.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/servicescommon/KafkaSender.java
+++ b/helios-services/src/main/java/com/spotify/helios/servicescommon/KafkaSender.java
@@ -39,7 +39,12 @@ public class KafkaSender implements EventSender {
     this.kafkaProducer = kafkaProducer;
   }
 
-  public void send(final KafkaRecord kafkaRecord) {
+  @Override
+  public boolean isHealthy() {
+    return true;
+  }
+
+  private void send(final KafkaRecord kafkaRecord) {
     if (kafkaProducer.isPresent()) {
       final ProducerRecord<String, byte[]> record =
           new ProducerRecord<>(kafkaRecord.getKafkaTopic(), kafkaRecord.getKafkaData());
--- a/helios-services/src/main/java/com/spotify/helios/agent/AgentConfig.java
+++ b/helios-services/src/main/java/com/spotify/helios/agent/AgentConfig.java
@@ -18,6 +18,7 @@
 package com.spotify.helios.agent;
 
 import com.spotify.docker.client.DockerHost;
+import com.spotify.helios.servicescommon.CommonConfiguration;
 import com.spotify.helios.servicescommon.FastForwardConfig;
 
 import java.net.InetSocketAddress;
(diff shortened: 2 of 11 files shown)


mattnworb:
@davidxia @rohansingh @lndbrg @zalenski 

this aims to avoid issues like https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1432

lndbrg:
i believe it would be beneficial to have a cache of senders that expires after n minutes and retries connection, in case the targeted event bus is down intermittently.

mattnworb:
I think there are two approaches we could take to improve this so that an intermittent network problem at startup does not cause the pubsub sender to be disabled for as long as the agent is up:

1. Add a periodic out-of-band check (similar to `getTopic` here) that checks connectivity, and enables/disables the sender based on the result.

2. Remove the healthcheck-at-startup added here and modify the failure listener in the FutureCallback added to the future returned by `pubsub.publishAsync(..)` to set a flag that disables sending for some configured period of time. When this time elapses, subsequent publish attempts would be made again.

I am leaning towards 1 as 2 would still cause publishing attempts if the agent had total unconnectivity to the pubsub service, and so the ThresholdBundler will still fill up with requests, but just at a slower rate.

lndbrg:
1 sounds like a good idea.

helios-services/src/main/java/com/spotify/helios/servicescommon/KafkaSender.java
```diff
@@ -39,7 +39,12 @@ public KafkaSender(final Optional<KafkaProducer<String, byte[]>> kafkaProducer)
     this.kafkaProducer = kafkaProducer;
   }
 
-  public void send(final KafkaRecord kafkaRecord) {
+  @Override
+  public boolean isHealthy() {
+    return true;
```
davidxia: Unrelated to PR, but should we always return true here? Is there value in checking we can talk to kafka by using `KafkaProducer.waitOnMetadata() for example?

mattnworb: since we don't have any issues with the kafka sender and bad behavior when messages can't be published, I opted to leave that out here to avoid changing anything that is working ok today.

PR #1033, opened 2016-12-05
Fix NPE in HostsResource

davidxia:
when trying to get host labels.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
@@ -1692,17 +1692,10 @@ public class ZooKeeperMasterModel implements MasterModel {
    */
   @Override
   public HostStatus getHostStatus(final String host) {
-    final Stat stat;
     final ZooKeeperClient client = provider.get("getHostStatus");
 
-    try {
-      stat = client.exists(Paths.configHostId(host));
-    } catch (KeeperException e) {
-      throw new HeliosRuntimeException("Failed to check host status", e);
-    }
-
-    if (stat == null) {
-      log.warn("Missing configuration for host {}", host);
+    if (!ZooKeeperRegistrarUtil.isHostRegistered(client, host)) {
+      log.warn("Host {} isn't registered in ZooKeeper.", host);
       return null;
     }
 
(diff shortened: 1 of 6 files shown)


davidxia:

davidxia:
I found an error by writing a test. `getLabels` should return empty map if hostStatus is null.

helios-services/src/main/java/com/spotify/helios/master/resources/HostsResource.java
```diff
@@ -122,11 +123,14 @@ public HostsResource(final MasterModel model) {
           })
           .collect(Collectors.toList());
 
+      final Function<String, Map<String, String>> getLabels =
+          host -> {
+            final HostStatus hostStatus = model.getHostStatus(host);
+            return hostStatus == null ? emptyMap() : hostStatus.getLabels();
```
mattnworb: would it be easier if `model.getHostStatus(host)` returned an `Optional<String>` instead? seems like every place that calls that method has to perform the null check, so might be nice to address that with the type system itself

helios-services/src/main/java/com/spotify/helios/master/resources/HostsResource.java
```diff
@@ -122,11 +123,14 @@ public HostsResource(final MasterModel model) {
           })
           .collect(Collectors.toList());
 
+      final Function<String, Map<String, String>> getLabels =
```
mattnworb: this might be easier to read as a private method rather than an inline function, since this doesn't close over anything in the current method body

helios-services/src/main/java/com/spotify/helios/master/resources/HostsResource.java
davidxia: You mean `Optional<HostStatus>`? Maybe...

davidxia:
This one is ready for review.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -1725,6 +1718,19 @@ public HostStatus getHostStatus(final String host) {
         .build();
   }
 
+  @Override
+  public Map<String, String> getHostLabels(final String host) {
+    final ZooKeeperClient client = provider.get("getHostStatus");
+
+    if (!ZooKeeperRegistrarUtil.isHostRegistered(client, host)) {
+      log.warn("Host {} isn't registered in ZooKeeper.", host);
```
mattnworb: would be helpful to have more context in log messages like this - like what method this is from, or what we were doing with this host in the first place, etc. In this case though is the log warning even useful?

davidxia: Personally, I don't think it's that useful.

helios-services/src/test/java/com/spotify/helios/master/resources/HostsResourceTest.java
```diff
@@ -80,6 +81,7 @@ public void setUp() {
           .build();
 
       when(model.getHostStatus(host)).thenReturn(hostStatus);
```
mattnworb: getHostStatus(host) should not be called anymore, so do you still need the `when()`? removing this catches the case where it is called unexpectedly

davidxia: Thanks

gimaker:
lgtm

PR #1034, opened 2016-12-06
Use com.spotify:ssh-agent-tls

davidxia:
which are open-source repos containing the client-side
TLS and HTTPS logic.


mattnworb:
assuming that this stuff was extracted so we can use it in our new project, I think it would make sense to wait to merge this into helios until we know the library is good and usable for the new project. This way you can avoid some back and forth of having to merge it, fix the library, bump this repo, etc.

we could also make these libraries be top-level modules in this project, and avoid having them be a wholly separate repo - since we deploy these jars to sonatype

davidxia:
Yes, we should wait until they work well for the new one.

davidxia:
@mattnworb @rohansingh We have a python client, but this still might be useful.

mattnworb:
I am indifferent to this personally - it makes sense to me to extract reusable code to a new library/repo if there are other things using it, but it might just add extra hurdles and steps if we have to make bugfixes if helios is the only user.

davidxia:
The code for this in helios seems pretty stable.

I think it'll be useful and cool for us to tell other teams (maybe even internal blog post or talk) who are thinking about adding authentication to their command line tools about this approach. I remember us comparing [crtauth](https://github.com/spotify/crtauth) with this approach in depth and deciding this approach was easier, more portable, and more secure.

Others might have ideas to contribute, and it'll be easier if it's in a separate repo.

davidxia:
@rohansingh ssh-agent-tls has been open-sourced. Can we merge this now?

mattnworb:
@davidxia just to double-check, there are no functional changes for Helios in this change, right? i.e. the name of the environment variables that you set to customize the location of the certs, etc ?

davidxia:
@mattnworb Thanks for checking. The [issuer DN](https://github.com/spotify/ssh-agent-tls/blob/master/src/main/java/com/spotify/sshagenttls/X509CertKeyCreator.java#L103) changes, but I don't think this makes any difference right?

The [cert cache dir](https://github.com/spotify/ssh-agent-tls/blob/9dac6298ff80a0378286a4cc63120a0bd7832aa7/src/main/java/com/spotify/sshagenttls/SshAgentHttpsHandler.java#L34) also changes, but this just means there'll be a one-time cache miss.

mattnworb:
@davidxia perhaps this comment belongs in a review of that new library, but if it always caches in the same directory, what happens if you have two applications both using that library? Won't those certs collide?

I would expect that calling code (like helios in this case) could customize those paths to provide isolation. i.e. a cert used by Helios may not make sense to also be used by some-other-CLI we might develop. Or would it?

davidxia:
We have open issues tracking both of these things [here](https://github.com/spotify/ssh-agent-tls/issues/4) and [here](https://github.com/spotify/ssh-agent-tls/issues/5), respectively.

davidxia:
Two issues above for ssh-agent-tls are now resolved.

davidxia:
Any thoughts on moving this forward?

mattnworb:
sounds good to me

PR #1036, opened 2016-12-09
Don't get host status when we don't need to

davidxia:
Getting the host status is an expensive operation because
it reads from many ZK paths. Don't use it when we don't need to.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
@@ -90,6 +90,7 @@ import com.google.common.collect.Maps;
 import com.google.common.collect.Ordering;
 import com.google.common.collect.Sets;
 
+import java.util.stream.Collectors;
 import org.apache.zookeeper.KeeperException;
 import org.apache.zookeeper.KeeperException.BadVersionException;
 import org.apache.zookeeper.KeeperException.NoNodeException;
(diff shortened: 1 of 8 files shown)


helios-services/src/main/java/com/spotify/helios/master/reaper/DeadAgentReaper.java
```diff
@@ -76,13 +75,12 @@ public DeadAgentReaper(final MasterModel masterModel,
   @Override
   void processItem(final String agent) {
     try {
-      final HostStatus hostStatus = masterModel.getHostStatus(agent);
-      if (hostStatus == null || hostStatus.getStatus() != HostStatus.Status.DOWN) {
-        // Host not found or host not DOWN -- nothing to do
+      if (!masterModel.isHostUp(agent)) {
+        // Host DOWN -- nothing to do
```
gimaker: You have inverted the logic here -- which I assume was unintended?

gimaker: Looks like perhaps this entire file should be left as-is.

davidxia: oh thanks

davidxia:
@gimaker The first commit is definitely good to do.

I'm less sure the second commit is as necessary. It changes to `MasterModel` interface just to have this reaper do less ZK operations. Thoughts?

helios-services/src/main/java/com/spotify/helios/master/reaper/DeadAgentReaper.java
davidxia: but why should the file be left as is?

davidxia:
cc @mattnworb @rohansingh

helios-services/src/main/java/com/spotify/helios/master/reaper/DeadAgentReaper.java
```diff
@@ -76,13 +75,12 @@ public DeadAgentReaper(final MasterModel masterModel,
   @Override
   void processItem(final String agent) {
     try {
-      final HostStatus hostStatus = masterModel.getHostStatus(agent);
-      if (hostStatus == null || hostStatus.getStatus() != HostStatus.Status.DOWN) {
-        // Host not found or host not DOWN -- nothing to do
+      if (masterModel.isHostUp(agent)) {
```
gimaker: Why change this logic at all?

AFAICT this is just doing the same thing as before, but adds a new method to do it.

gimaker: it does

davidxia:
I'll move the second commit into another PR for more feedback on whether it's worth adding the extra methods to the `MasterModel` interface.

davidxia:
Second commit moved to https://github.com/spotify/helios/pull/1037.

PR #1039, opened 2016-12-12
Undeploy jobs from hosts even if they are DOWN

rohansingh:
If a host is down, we still want the `RollingUndeployPlanner` to attempt
to undeploy from it, and then remove it from the list of removed hosts.

Otherwise, the list of removed hosts for a deployment group can grow
unboundedly.


rohansingh:
@mattnworb @gimaker @davidxia

rohansingh:
@spotify/helios-team

mattnworb:
👍  but would prefer if we could add a test for this too

PR #1040, opened 2016-12-12
monitor how often zookeeper connection state changes

mattnworb:
Adds meters for how often the state of the zookeeper connection changes.

Intended to help monitor issues where the masters or agents continually disconnect/reconnect from Zookeeper (for example, if the masters are in a loop where they continually send bad or too-large requests to ZK, which causes the connection to be dropped by the ZK server).


mattnworb:
@spotify/helios-team

PR #1041, opened 2016-12-12
[wip] GceAgentReaper initial draft

rohansingh:


helios-services/src/main/java/com/spotify/helios/master/reaper/GceAgentReaper.java
```diff
@@ -0,0 +1,155 @@
...
+        return;
+      }
+
+      final AgentInfo agentInfo = masterModel.getAgentInfo(host);
+      if (agentInfo == null) {
+        return;
+      }
+
+      try {
+        log.info("Reaping deleted GCE instance '{}'");
```
mattnworb: no arguments are passed here

helios-services/src/main/java/com/spotify/helios/master/reaper/GceAgentReaper.java
```diff
@@ -0,0 +1,155 @@
...
+        masterModel.deregisterHost(host);
+      } catch (Exception e) {
+        log.warn("Failed to reap deleted GCE instance '{}'", host, e);
+      }
+    } catch (Exception e) {
+      log.warn("Failed to parse pubsub message: {}", message, e);
```
mattnworb: This seems like a really broad catch - would be better to only have the pubsub field checks in this try. Is there not a simple boolean method that could be checked rather than using an exception?

mattnworb: Also, iirc Message.toString() logs a message JD and just a portion of the message as hex, not the full message contents, in case you are expecting that to be logged.

mattnworb: s/JD/ID ... commenting from a mobile device is less easy than it seems

mattnworb:
if there are multiple helios clusters in different GCP regions, will the pubsub topic contain `compute.instances.delete` events for just that region or for all? In other words, is there one pubsub topic per helios cluster, or will all of the masters be receiving the same stream of pubsub events?

The latter might present an issue if the single global topic contains a `compute.instances.delete` event for `region1-host1` if a master in `region2` is the one to pick up that message first. That master will ack it, fail to deregister the host since it doesn't exist in it's cluster, and the other cluster will never get that event. 

One way to fix this, if the pubsub topic is global, is to create one subscription per cluster rather than a single `GceAgentReaper`subscription that all masters share.

PR #1042, opened 2016-12-13
More robust cleanup of removed hosts for deployment groups

rohansingh:
Instead of the previous method where we would try to undeploy from hosts
that were no longer registered, instead just apply a straightforward filter
to the removed hosts list.

This is clearer and works without generating a bunch of unnecessary tasks.
Also, the previous version would only work if the set of UP hosts for the
deployment group changed. This version doesn't have that flaw.

This reverts commit 6412fc02d1c18915e3f6f31f1817b9cfc030e309.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
@@ -516,7 +516,10 @@ public class ZooKeeperMasterModel implements MasterModel {
 
     // Calculate the freshly removed hosts (current - new) and add in any previously removed hosts
     // that haven't been undeployed yet.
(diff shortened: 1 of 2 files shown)


rohansingh:
@spotify/helios-team

mattnworb:
so hosts that are no longer in the DG *and* happen to be down will stay in the `/removed` znode until they are deregistered *and* another rolling-update happens for that group (either manually or because of another host change)?

rohansingh:
@mattnworb Almost right. The hosts will be removed from the `/removed` znode when they are deregistered, regardless of another rolling-update happening. This is because the reactor to update the hosts lists executes periodically regardless of rolling updates.

mattnworb:
thanks for the clarification. I think the stated outcome makes sense. It's hard to evaluate the code on it's own - it is a big bummer that we don't have any real unit tests of any of this rolling-update logic.

gimaker:
:+1:

PR #1043, opened 2016-12-15
use same pubsub topic in healthchecks as the EventSender

mattnworb:
Previously the EventSenderFactory is passing a hardcoded `health.canary`
topic name to GooglePubSubSender$DefaultHealthChecker. The assumption
here was that if the client could successfully access the Cloud PubSub
API then it could check to see if any topic existed.

However it is possible for the PubSub credentials to be limited to only
publish and even *see* certain topics within a project. This would cause
the healthchecker to get `PERMISSION_DENIED` when checking if the
healthcheck topic existed.

To fix this, use the same topic name in the healthchecker as what will
eventually be published to.

Refactored the ZooKeeperAgentModel/MasterModel so that the topic name to
publish to is passed in as a constructor argument rather than being
hardcoded, so that the Agent/MasterService that constructs the Model
class can control that the same topic is used both in the healthchecker
and the eventSender.


mattnworb:
@spotify/helios-team  @rzvoncek @lndbrg

lndbrg:
:+1:

gimaker:
:+1:

PR #1044, opened 2016-12-15
Fixed check for volume source being null or empty

piccobit:
This bug was hidden by the second condition which is normally always false and was only discovered when I disabled the second condition to allow the usage of named volumes which don't start with a '/'.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
@@ -150,7 +150,7 @@ public class JobValidator {
         errors.add("Volume path is not absolute: " + path);
         continue;
       }
-      if (!isNullOrEmpty(source) && !source.startsWith("/")) {
+      if (isNullOrEmpty(source) && !source.startsWith("/")) {
         errors.add("Volume source is not absolute: " + source);
         continue;
       }
--- a/helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
@@ -150,7 +150,11 @@ public class JobValidator {
         errors.add("Volume path is not absolute: " + path);
         continue;
       }
-      if (isNullOrEmpty(source) && !source.startsWith("/")) {
+      if (isNullOrEmpty(source)) {
+        errors.add("Volume source is null or empty");
+        continue;
+      }
+      if (!source.startsWith("/")) {
         errors.add("Volume source is not absolute: " + source);
         continue;
       }


helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
```diff
@@ -150,7 +150,7 @@ public JobValidator(final boolean shouldValidateJobHash,
         errors.add("Volume path is not absolute: " + path);
         continue;
       }
-      if (!isNullOrEmpty(source) && !source.startsWith("/")) {
+      if (isNullOrEmpty(source) && !source.startsWith("/")) {
```
rohansingh: Are you sure this logic wasn't correct before? Wouldn't this now throw an NPE if `source` is null?

piccobit: @rohansingh Ah, yes, you're right! In our use case with the named volume I've removed the second condition completely and there it makes sense!

piccobit: Ok, I've separated the checks for the volume source to prevent the NPE.

helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
```diff
@@ -150,7 +150,11 @@ public JobValidator(final boolean shouldValidateJobHash,
         errors.add("Volume path is not absolute: " + path);
         continue;
       }
-      if (!isNullOrEmpty(source) && !source.startsWith("/")) {
```
davidxia: We can just change `&&` to `||`, but I guess more specific errors can't hurt.

helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
```diff
@@ -150,7 +150,11 @@ public JobValidator(final boolean shouldValidateJobHash,
...
+      if (isNullOrEmpty(source)) {
```
mattnworb: it seems like before, a null or empty source was allowed. Now this will cause an error. Am I correct on that and is that change intentional?

mattnworb:
we seem to mention in https://github.com/spotify/helios/blob/master/docs/user_manual.md#volumes that the right-hand side of the volume can be left off or empty

davidxia:
Ah, the original code makes sense then.

helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
mattnworb: @piccobit thanks - it would be helpful to see both changes in context.

PR #1048, opened 2016-12-18
Wait for correct number of host labels in test

davidxia:
Make `SystemTestBase.awaitHostStatusWithLabels()` wait for the correct
number of labels instead of just waiting until it's up with at least one
label. This fixes a race condition in
`DeploymentGroupTest.testRemovingAgentTagUndeploysJob()`.

### DIFF
--- a/helios-system-tests/src/main/java/com/spotify/helios/system/SystemTestBase.java
+++ b/helios-system-tests/src/main/java/com/spotify/helios/system/SystemTestBase.java
@@ -919,7 +919,7 @@ public abstract class SystemTestBase {
 
       if (candidate == null || candidate.getStatus() != status
           // labels are stored in ZK after the host has come up
-          || candidate.getLabels().size() == 0) {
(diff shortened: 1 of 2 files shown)


mattnworb:
@davidxia do the tests pass for this change?

helios-system-tests/src/main/java/com/spotify/helios/system/SystemTestBase.java
```diff
@@ -919,7 +919,7 @@ protected HostStatus awaitHostStatusWithLabels(final HeliosClient client,
 
       if (candidate == null || candidate.getStatus() != status
           // labels are stored in ZK after the host has come up
-          || candidate.getLabels().size() == 0) {
+          || candidate.getLabels().size() == labels.size()) {
```
davidxia: This should've been `!=`.

davidxia:
I've reverted this and made a new PR here. https://github.com/spotify/helios/pull/1054

PR #1049, opened 2016-12-19
Upgrade netty to 4.1.5 with exclusion for google-cloud

davidxia:
The version of netty pulled in by pubsub has a memory leak.

* https://groups.google.com/forum/#!topic/grpc-io/4wHa_fxaT-Q
* https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1433

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -263,6 +263,29 @@
         <artifactId>jsr305</artifactId>
         <version>2.0.3</version>
       </dependency>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>google-cloud-pubsub</artifactId>
+        <version>0.5.1</version>
+        <exclusions>
+          <exclusion>
+            <groupId>javax.servlet</groupId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -263,6 +263,26 @@
...
+          <exclusion>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-all</artifactId>
+          </exclusion>
+        </exclusions>
+      </dependency>
+      <dependency>
+        <groupId>io.netty</groupId>
+        <artifactId>netty-all</artifactId>
+        <version>4.1.5.Final</version>
```
mattnworb: should add a note in here about why we are using this version, linking to the two other issues, so that someone doesn't accidentally change it in the future without being aware.

I'm not sure that we need to care that the same version of netty-all is used in helios-services (where it really matters) and in the system tests (where it probably doesn't matter what version is used), but if the latter works with this change then that is fine too

PR #1050, opened 2016-12-19
Fix HostListcommand output for no hosts

davidxia:
`helios hosts --json` returned an empty JSON object,
    but `helios hosts foo --json` returned an empty JSON array.
    Make them both return empty objects.
    
    This fixes a race condition in `Labeltest.testCliHosts()`.


davidxia:
Test failure is unrelated.

davidxia:
Good point. It was a bug, I think. I've made both return empty JSON object.

davidxia:
Tests passed!

PR #1055, opened 2016-12-21
remove fallback for batch status fetching

mattnworb:
When the bulk jobStatus method was first added to the API (back in
commit b1d345f), a fallback was added in case the bulk query failed to
loop over the list of jobs and fetch the status of each individually.

Assuming this fallback was added to handle having a newer CLI version
talking to a version of the master that didn't have this API yet, I
think it is safe to remove this logic at this point.

If the bulk query fails, it would be better to just surface the error rather than resort to the much slower N+1 query.


davidxia:
I don't know why the fallback was kept, but I guess we'd never want it to fallback to n+1 queries even if the bulk response is huge? I guess ZK instead of HTTP will be the bottleneck in this case.

mattnworb:
had to fix some checkstyle conflicts - rebased

mattnworb:
@gimaker @rohansingh

PR #1056, opened 2016-12-21
remove riemann support

mattnworb:
The support for emitting events via Riemann is not well maintained and I
am not sure that anyone is even using it.


davidxia:
I'm personally not that familiar with this part of the code. I guess we just use ffwd for metrics now?

mattnworb:
rebased to fix conflicts

davidxia:
Do you still want to merge this?

mattnworb:
Yes. @rohansingh any thoughts on if we should keep this support around?

rohansingh:
👍 

We don't use `--riemann-host-port` internally, and I highly, highly doubt anyone is using it externally.

mattnworb:
rebased on master

PR #1058, opened 2016-12-27
Upgrade docker-client dep from 5.0.2 to 7.0.0

davidxia:


davidxia:
Depends on https://github.com/spotify/docker-client/pull/571

helios-services/pom.xml
```diff
@@ -319,6 +318,13 @@
       <version>0.8</version>
     </dependency>
 
+    <!-- versions that we have to manage to avoid requireUpperBoundDeps violations -->
```
mattnworb: this should probably be in a `<dependencyManagement>` section

mattnworb: ah I see httpcore is in dependencyManagement in the parent pom, so why not manage both there?

helios-testing/pom.xml
```diff
@@ -41,7 +41,7 @@
     <dependency>
       <groupId>com.spotify</groupId>
       <artifactId>docker-client</artifactId>
-      <version>5.0.2</version>
+      <version>6.2.2</version>
```
mattnworb: we should manage this in the parent pom too IMHO

helios-services/pom.xml
davidxia: You mean httpclient? Yea I was going to do this originally but hesitated because I didn't want to change too much. I'll move it.

davidxia:
Yea just testing it passes tests.

rohansingh:
👍  once docker-client 7.0.0 is released and we use that (instead of 7.0.0-rc3).

PR #1060, opened 2016-12-29
Improve rolling-update logging

davidxia:
Logging for large deployment groups is unreadable because there
are so many `RolloutTasks` in a `DeploymentGroupTask`.

Instead of logging the entire `DeploymentGroupTask` which includes
the entire list of `RolloutTask`s, just log the current `RolloutTask`.


davidxia:
cc @gimaker @rohansingh

PR #1061, opened 2017-01-03
name the threads used by HeliosClient

mattnworb:
for improved debug-ability

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
+++ b/helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
@@ -36,7 +36,6 @@ import static java.net.HttpURLConnection.HTTP_NOT_FOUND;
 import static java.net.HttpURLConnection.HTTP_OK;
 import static java.util.Arrays.asList;
 import static java.util.Collections.singletonList;
-import static java.util.concurrent.Executors.newScheduledThreadPool;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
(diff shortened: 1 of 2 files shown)


helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
```diff
@@ -655,9 +661,18 @@ public HeliosClient build() {
     }
 
     private static ListeningScheduledExecutorService defaultExecutorService() {
-      final ScheduledExecutorService executor = MoreExecutors.getExitingScheduledExecutorService(
-          (ScheduledThreadPoolExecutor) newScheduledThreadPool(4), 0, SECONDS);
-      return MoreExecutors.listeningDecorator(executor);
+      final int clientCount = clientCounter.incrementAndGet();
+
+      final ThreadFactory threadFactory = new ThreadFactoryBuilder()
+          .setNameFormat("helios-client-" + clientCount + "-thread-%d")
```
davidxia: I'm not 100% sure, but it seems like `Executors.defaultThreadFactory()` will do this type of naming. Perhaps just create one instance for all `HeliosClient`s instead of creating a new instance of a `ThreadFactory` for every `HeliosClient` instance? Or will that not be enough in some way?

mattnworb:
@gimaker @rohansingh

helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
davidxia: Ah sorry. I see now.

PR #1062, opened 2017-01-03
add a property to allow skipping of debian package building

mattnworb:
This might be convenient when trying to do `mvn install` to be able to
test one of the modules in another project without having to wait while
the Debian packages build.


mattnworb:
@davidxia `-Dproperty.name=value`

PR #1063, opened 2017-01-03
TemporaryJobs: move old job cleanup to apply()

mattnworb:
Instead of attempting to clean up new jobs in the TemporaryJobs
constructor, instead perform the cleanup (just one time) when the
Statement is evaluated.

In other words, we attempt to cleanup old jobs when the test method is
being executed (if the user is using `@Rule`).

This avoids conflicts when used with HeliosSoloDeployment in a RuleChain
(as the docs suggest) like:

```
@ClassRule
public static final RuleChain chain = RuleChain
    .outerRule(soloResource)
    .around(temporaryJobs);
```

If the old job cleanup is performed in the constructor, then
TemporaryJobs will start attempting to query the Helios cluster before
it has been fully initialized by the HeliosDeploymentResource named
`soloResource` in the above example.

### DIFF
--- a/helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/TemporaryJobs.java
@@ -98,9 +98,10 @@ public class TemporaryJobs implements TestRule {
   private final Map<String, String> env;
   private final List<TemporaryJob> jobs = Lists.newCopyOnWriteArrayList();
   private final Deployer deployer;
-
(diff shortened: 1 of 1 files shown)


mattnworb:

gimaker:
Nice catch!

`removeOldJobs()` is called both in the ctor and in `apply()` - is that intentional? The PR description suggests that it should no longer be called in the ctor

mattnworb:
@gimaker yep that was a mistake - fixed in 6606572

PR #1065, opened 2017-01-09
Make helios support IP in docker port mapping

davidxia:


helios-client/src/main/java/com/spotify/helios/common/descriptors/PortMapping.java
```diff
@@ -41,33 +44,54 @@
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class PortMapping extends Descriptor {
 
+  public static final String WILDCARD_ADDRESS = "0.0.0.0";
   public static final String TCP = "tcp";
   public static final String UDP = "udp";
 
+  // Nullable for backwards compatibility
+  // TODO (dxia) Should we validate this string?
```
mattnworb: I think this could be useful - if so you can use `boolean com.google.common.net.InetAddresses.isInetAddress(String)`, which avoids any DNS lookups

helios-tools/src/main/java/com/spotify/helios/cli/command/PortMappingParser.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+
+class PortMappingParser {
+
+  private static final Pattern PATTERN =
+      compile("(?<n>[_\\-\\w]+)=((?<ip>(localhost|([0-9]{1,3}.){3}[0-9]{1,3})):)"
+              + "?(?<i>\\d+)(:(?<e>\\d+))?(/(?<p>\\w+))?");
```
mattnworb: do we definitely still need a regex here? A regex that checks if something is a valid IP address feels prone to failure to me. Would it be easier (and more correct) to split the string up by various tokens and then pass the ip address portion to the guava `boolean isInetAddress(string)` function?

helios-client/src/main/java/com/spotify/helios/common/descriptors/PortMapping.java
davidxia: Yea I started doing this.

helios-tools/src/main/java/com/spotify/helios/cli/command/PortMappingParser.java
davidxia: agreed

davidxia: Hm, this might be hard to do and be backwards compatible. We need to deal with strings like `http=8080:80` and `http=0.0.0.0:8080:80`. What kind of regex would match both with the appropriate named capture groups?

mattnworb: I think you're right, that a regex is better

davidxia:
@mattnworb This is ready for another look. Did I forget to update anything?

docs/user_manual.md
```diff
@@ -272,7 +273,10 @@ For example, `{"foo": {"internalPort": 4711}}`  will  map the internal port
 4711 of the container to an arbitrary external port on the host. 
   
 Specifying `{"foo": {"internalPort": 4711, "externalPort": 80}}` will map
-internal port 4711  of the container to port 80 on the host. 
+internal port 4711  of the container to port 80 on the host.
+
+The `ip` attribute is optional and specifies the IP address on which the port
+will be exposed. Any IP string literal is valid.
```
mattnworb: I think it would help to clarify that this is referring to an IP address of the host that the Helios job will run on, and there is no way for Helios to verify that this IP address actually exists or will work at deploy-time, and putting bad values here will cause your deploy to fail.

mattnworb: And in other words this is likely only useful if you want to deploy a job that exposes a port mapping that only listens on localhost

davidxia: Good first point. On the second, a machine could have multiple IPs though.

mattnworb:
👍

davidxia:
@gimaker as another pair of eyes since this can break things if we don't handle old data in ZK correctly.

docs/user_manual.md
```diff
@@ -272,7 +273,11 @@ For example, `{"foo": {"internalPort": 4711}}`  will  map the internal port
 4711 of the container to an arbitrary external port on the host. 
   
 Specifying `{"foo": {"internalPort": 4711, "externalPort": 80}}` will map
-internal port 4711  of the container to port 80 on the host. 
+internal port 4711  of the container to port 80 on the host.
+
+The `ip` attribute is optional and specifies the IP address on which the port
+will be exposed. Although Helios will check the IP string is a valid IP address,
+it won't verify the address is routable or whether it'll work at deploy-time.
```
gimaker: Consider rewording this to talk about which interface to listen on. E.g. "... and specifies the IP of the interface to expose the port on."

helios-client/src/main/java/com/spotify/helios/common/descriptors/PortMapping.java
```diff
@@ -41,31 +46,46 @@
...
 
+  private final String ip;
   private final int internalPort;
   private final Integer externalPort;
   private final String protocol;
 
-  public PortMapping(@JsonProperty("internalPort") final int internalPort,
-                     @JsonProperty("externalPort") final Integer externalPort,
-                     @JsonProperty("protocol") final String protocol) {
+  public PortMapping(@JsonProperty("ip") @Nullable final String ip,
```
gimaker: Is this public API that's used anywhere? If yes, add an overload that uses a default `ip`

docs/user_manual.md
davidxia: I've wondered what's the correct terminology here. Does one say a port is exposed on an IP and that an IP is bound to an interface?

davidxia: I'm also wondering why we should mention the word "interface" specifically. Isn't saying IP enough?

helios-client/src/main/java/com/spotify/helios/common/descriptors/PortMapping.java
davidxia: Don't I have that on line 62?

davidxia:
@gimaker How's it look overall? Any backwards incompatibility?

helios-client/src/main/java/com/spotify/helios/common/descriptors/PortMapping.java
gimaker: Can only see `PortMapping(final int internalPort, final Integer externalPort)` and `PortMapping(final int internalPort)` - but on second thought it seems unlikely that is used apart from json parsing

docs/user_manual.md
gimaker: I think it's perfectly fine as it is now. Personally, I find it clearer if you speak about interfaces but that's just me.

For inspiration, this is what the unbound docs say:

```
       interface: <ip address[@port]>
              Interface  to  use  to connect to the network. This interface is
              listened to for queries from clients, and answers to clients are
              given  from  it.  Can be given multiple times to work on several
              interfaces. If none are given the default is to listen to local-
              host.   The  interfaces  are not changed on a reload (kill -HUP)
              but only on restart.  A port number can be specified with  @port
              (without spaces between interface and port number), if not spec-
              ified the default port (from port) is used.

       ip-address: <ip address[@port]>
              Same as interface: (for easy of compatibility with nsd.conf).
```

and nginx (no mention of interfaces here!):

```
Syntax:	listen address[:port] ...
...
Sets the address and port for IP, or the path for a UNIX-domain socket on which the server will accept requests. Both address and port, or only address or only port can be specified. An address may also be a hostname, for example:
```

davidxia: Ah, thanks for the references.

PR #1066, opened 2017-01-10
replace spotify.github.io/helios-apt with bintray

mattnworb:


mattnworb:
@rohansingh @davidxia

mattnworb:
I updated the commit - apt seems to require the filename to end in `.list`, and doesn't allow `.pref`.

PR #1068, opened 2017-01-11
Fix inversion of namespace and repo name checkers

rwstephenson:


mattnworb:
Thanks for the patch. I believe this causes some of the existing unit tests to fail though, and it would be nice to add a new test for the specific image name that had trouble here. I'll push this commit as a new branch directly to this fork of the repo to make sure that CircleCI can build it ok (it won't build from forks for security reasons).

mattnworb:
created #1069 to work on this in

PR #1069, opened 2017-01-11
Fix image names with hyphens being reported as invalid

### LINKED ISSUE #1070: Image name b.gcr.io/cloudsql-docker/gce-proxy:1.05 mistakenly treated as invalid by JobValidator
reported by @rwstephenson

```
Invalid namespace name (cloudsql-docker), only [a-z0-9_] are allowed, size between 4 and 30
```

docker pulls this image just fine though:

```
$> docker pull b.gcr.io/cloudsql-docker/gce-proxy:1.05
1.05: Pulling from cloudsql-docker/gce-proxy
134472cccba7: Pull complete
Digest: sha256:3b6ebcfc6801d616ceded25b0a3e9c559d97e538a7f6d2a0a6097f7dfeca2bfc
Status: Downloaded newer image for b.gcr.io/cloudsql-docker/gce-proxy:1.05
```

mattnworb:
Replaces #1068. Will fix #1070.

It seems like at some point Docker changed the format of image names to
be less restrictive: the image name can have multiple slash-separated
name components (with no bound on the number of components):

> The image name which is made up of slash-separated name components...
> Name components may contain lowercase characters, digits and
> separators. A separator is defined as a period, one or two
> underscores, or one or more dashes. A name component may not start or
> end with a separator.

This commits removes the regexes in JobValidator that check the
"namespace" and "repoName" separately along with the limits on how long
each portion can be. It now validates that each slash-seperated name
component contains only the allowed characters and that the entire
repository name is below the limit of 255 characters.

See:
docker/docker@ea98cf7
https://github.com/docker/distribution/blob/master/reference/reference.go
https://github.com/docker/distribution/blob/master/reference/regexp.go

### DIFF
--- a/helios-client/src/test/java/com/spotify/helios/common/JobValidatorTest.java
+++ b/helios-client/src/test/java/com/spotify/helios/common/JobValidatorTest.java
@@ -40,7 +40,9 @@ import static com.spotify.helios.common.descriptors.Job.EMPTY_SECONDS_TO_WAIT;
 import static com.spotify.helios.common.descriptors.Job.EMPTY_SECURITY_OPT;
 import static com.spotify.helios.common.descriptors.Job.EMPTY_TOKEN;
 import static com.spotify.helios.common.descriptors.Job.EMPTY_VOLUMES;
+import static org.hamcrest.CoreMatchers.not;
 import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.empty;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.core.Is.is;
@@ -456,7 +458,7 @@ public class JobValidatorTest {
   @Test
   public void testImageNamespaceWithHyphens() {
     final Job job = VALID_JOB.toBuilder()
-        .setImage("b.gcr.io/cloudsql-docker/gce-proxy:1.05 ")
+        .setImage("b.gcr.io/cloudsql-docker/gce-proxy:1.05")
         .build();
 
     assertThat(validator.validate(job), is(empty()));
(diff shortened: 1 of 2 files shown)


mattnworb:
Note that I didn't implement checking that the "separator character" fits the exact definition of `A separator is defined as a period, one or two
underscores, or one or more dashes` as that seems very restrictive.

mattnworb:
@rohansingh @gimaker @davidxia @rwstephenson

helios-client/src/test/java/com/spotify/helios/common/JobValidatorTest.java
```diff
@@ -451,4 +452,23 @@ public void testWhitelistedCapabilities_withWhitelist() throws Exception {
     final Set<String> errors2 = validator2.validate(job);
     assertEquals(0, errors2.size());
   }
+
+  @Test
+  public void testImageNamespaceWithHyphens() {
+    final Job job = VALID_JOB.toBuilder()
+        .setImage("b.gcr.io/cloudsql-docker/gce-proxy:1.05 ")
```
davidxia: A space after the version is OK?

davidxia:
Do we have any docs on image name that need to be updated?

helios-client/src/test/java/com/spotify/helios/common/JobValidatorTest.java
mattnworb: No, I don't think so, but that was an accident https://github.com/docker/distribution/blob/master/reference/regexp.go#L36

mattnworb: added a commit to catch this, good catch

mattnworb:
@davidxia I can't find documentation in Helios that refers to what a valid image string or tag is (which is good - we shouldn't document what another system considers valid).

PR #1071, opened 2017-01-12
log less in GooglePubSubSender healthchecks

mattnworb:
instead of logging a message every time the healthcheck is successful,
only log the "is healthy" message when the state is transitioning from
not-healthy to healthy.


mattnworb:

PR #1072, opened 2017-01-12
make sure we always log why a job's healthcheck fails

mattnworb:
For exec healthchecks, nothing was logged at all if a DockerException or
InterruptedException was caught.

HTTP healthchecks don't log anything if they get a
MalformedUrlException, and TCP healthchecks don't log any failure at all
either.

Also make the logs more consistent by always mentioning the type of
healthcheck and the containerId.


mattnworb:

rohansingh:
👍

PR #1074, opened 2017-01-12
HealthCheckTest: remove docker driver restriction

mattnworb:
the test seems to always be skipped because of this clause


mattnworb:
ah so here is why this test was previously checking that the driver is native, on Circle it gets:

> Unsupported: Exec is not supported by the lxc driver

davidxia:
Let's try checking if it's not run by Circle. `pom.xml` inserts the CIRCLECI env var for circle builds, so you can do something like `assumeThat(getenv("CIRCLECI"), isEmptyOrNullString());`

I think we have the same issues with docker-client.

davidxia:
@mattnworb What's the right way of getting the execution driver if that method is deprecated?

mattnworb:
docker doesn't have an execution driver anymore, see https://docs.docker.com/engine/reference/api/docker_remote_api/#v124-api-changes

davidxia:
Similar one here https://github.com/spotify/docker-client/pull/587

davidxia:
CircleCI build is green. It didn't update the status here though.

PR #1075, opened 2017-01-12
don't assign default values to PortMapping.ip

mattnworb:
We can't assign default values from within the POJO class to any
"descriptors" that are reachable from the Job class. When the Job
builder computes the JobId.hash, it takes all of the Job parameters that
are not meta parameters (username, create time), serializes them to
JSON, and then computes a sha1 hash for that string.

If we change things so that a client running an older version of the
code does not have the same default values for fields that the server
does, then the JobId.hash computed on the server-side will be different
than what the client computed.

One way that this will surface itself is via users of helios-testing and
TemporaryJobs; since HeliosSoloDeployment always pulls
`spotify/helios-solo:latest`, regardless of what version of
helios-testing is being used.

Another possible way this problem would express itself is if an older
helios-cli user tried to create a job against a newer helios master.

The PortMapping.ip field was added in #1065

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/PortMapping.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/PortMapping.java
@@ -46,7 +46,6 @@ import org.jetbrains.annotations.Nullable;
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class PortMapping extends Descriptor {
 
-  public static final String WILDCARD_ADDRESS = "0.0.0.0";
   public static final String TCP = "tcp";
   public static final String UDP = "udp";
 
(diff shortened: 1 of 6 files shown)


mattnworb:
You can reproduce this bug by running tests in a repo with helios-testing:0.9.45 that uses HeliosSoloDeployment. Jobs will fail to deploy with errors like

> java.lang.AssertionError: Failed to create job tmp-20170112-a979ac64_spotify_foobar_latest:608b8e43:9843a1140ed3e3d14e207a794cb7707ae38a39cf - CreateJobResponse{status=INVALID_JOB_DEFINITION, errors=[Id hash mismatch: 9843a1140ed3e3d14e207a794cb7707ae38a39cf != aac98ec92542ca337e1797f857721da9856c85a7], id='tmp-20170112-a979ac64_spotify_foobar_latest:608b8e43:9843a1140ed3e3d14e207a794cb7707ae38a39cf'

mattnworb:
I verified this change works by running the tests with the same repo that is using helios-testing:0.9.45 against a master running with this code; the attempt by TemporaryJobs to create the job succeeds ✅

gimaker:
:+1:

davidxia:
my god. Helios makes it so hard to be backwards compatible.

davidxia:
I guess a null value will work when it's passed to docker client?

mattnworb:
Yep, that is what the old code did: https://github.com/spotify/helios/pull/1065/files#diff-f91e971f38066cb5d88314fc8cb0b61d

davidxia:
Great. Thanks for catching this.

PR #1076, opened 2017-01-13
add a test for backwards-incompatible changes to the Job class

mattnworb:
Add a test that makes sure that the computed hash of a Job instance
stays the same over time, by failing if someone were to add new
non-null/non-empty fields to the Job class (or it's fields), thus
changing the hash result.


mattnworb:

mattnworb:
this would catch bugs like what was introduced in #1065 (which was fixed in #1075)

rohansingh:
Good idea. 👍

rohansingh:
Would be fun to cherry-pick this change onto the top of #1065 and verify that the test actually fails.

davidxia:
💯

PR #1077, opened 2017-01-13
Upgrade docker-client dep from 5.0.2 to 7.0.2

davidxia:
This reverts commit db39f5e
and upgrades docker-client to 7.0.2.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/agent/HealthCheckerFactory.java
+++ b/helios-services/src/main/java/com/spotify/helios/agent/HealthCheckerFactory.java
@@ -30,10 +30,13 @@ import com.spotify.docker.client.DockerClient;
 import com.spotify.docker.client.DockerHost;
 import com.spotify.docker.client.LogStream;
 import com.spotify.docker.client.exceptions.DockerException;
+import com.spotify.docker.client.messages.ExecCreation;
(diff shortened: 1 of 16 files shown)


helios-services/src/main/java/com/spotify/helios/agent/HealthCheckerFactory.java
```diff
@@ -90,14 +93,17 @@ public boolean check(final String containerId) {
...
             DockerClient.ExecCreateParam.attachStdout(),
             DockerClient.ExecCreateParam.attachStderr());
+        final String execId = execCreation.id();
 
-        final String output;
+        String output = "";
         try (LogStream stream = docker.execStart(execId)) {
           output = stream.readFully();
+        } catch (IOException e) {
+          // thrown from implicit call to close(). Ignore.
```
mattnworb: why do we ignore this exception? 

fwiw it seems like docker-client's DefaultLogStream.close() can not throw an IOException, so perhaps the LogStream interface should redefine close() to not throw a checked exception.

davidxia: I guess we shouldn't ignore? Redefining close() seems like a better idea though.

rohansingh:
👍

PR #1080, opened 2017-01-23
simplify ZooKeeperHealthChecker by removing PathChildrenCache

mattnworb:
It is not clear why this healthcheck creates and starts a PathChildrenCache for the purposes of knowing if zookeeper is connected or not - the PathChildrenCache proactively attempts to keep all children of a ZK path locally cached. This involves registering watches on those paths, responding to the events, keeping the data in-memory, etc.

Keeping the /status/hosts tree cached on the masters would be a pretty heavyweight operation - the cache kept by the healthcheck is updated every single time any host in the cluster has a status change, has a job deployed, etc.

Since the only thing that this healthcheck really cares about is the state of the zookeeper *connection* itself, and not the state of a tree, replace this with a ConnectionStateListener.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/ZooKeeperHealthChecker.java
+++ b/helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/ZooKeeperHealthChecker.java
@@ -23,115 +23,104 @@ package com.spotify.helios.servicescommon.coordination;
 import com.codahale.metrics.health.HealthCheck;
 import com.spotify.helios.servicescommon.RiemannFacade;
 import io.dropwizard.lifecycle.Managed;
+import java.sql.Connection;
+import java.time.Instant;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
-import org.apache.curator.framework.CuratorFramework;
-import org.apache.curator.framework.recipes.cache.PathChildrenCache;
-import org.apache.curator.framework.recipes.cache.PathChildrenCacheEvent;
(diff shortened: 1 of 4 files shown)


mattnworb:

mattnworb:
We have been seeing some issues in production where this healthcheck is reporting that the connection is suspended even though normal API operations on the master (listing masters, listing hosts, etc) work fine. I have a feeling, but no proof, that it has to do with the use of PathChildrenCache here.

rohansingh:
@danielnorberg Any ideas on why this originally used a `PathChildrenCache`? Is the connection state listener not trustworthy or something?

rohansingh:
Actually never mind, looks like this was originally written by @drewcsillag. Don't see any specific comments or commit messages though.

Might as well use the `ConnectionStateListener`. 👍

helios-services/src/main/java/com/spotify/helios/servicescommon/coordination/ZooKeeperHealthChecker.java
```diff
@@ -23,115 +23,104 @@
...
-    scheduler.scheduleAtFixedRate(this, 0, interval, timeUnit);
+    this.connectionStateListener
+        .stateChanged(zooKeeperClient.getCuratorFramework(), interpretedState);
   }
 
   @Override
   public void stop() throws Exception {
-    scheduler.shutdownNow();
+    // probably irrelevant but remove the listener to be safe
+    this.zooKeeperClient.getConnectionStateListenable().removeListener(connectionStateListener);
```
davidxia: Should we keep `scheduler.shutdownNow();`?

mattnworb: I meant to add that back, thanks for catching

PR #1081, opened 2017-01-25
helios-solo: skydns workaround to allow DNS responses up to 32768 bytes

gimaker:
TL;DR; skydns does not handle TCP well. We already worked around this
in https://github.com/spotify/helios/pull/900. See thar PR for more context.
However, that change only fixed the problem to an extent as we still have
the same issue once the responses are >4096 bytes.

This change extends that workaround to allow us to survive responses up to
32768 bytes in size. This change does not fix the issue, but should make it
more rare.


gimaker:
@davidxia @rohansingh @mattnworb

mattnworb:
Nice work. Could you also document somewhere (for instance `docs/helios_solo.md`) what error messages you can look for from helios-solo or skydns when this problem occurs? For that one user in the future who encounters a giant 33kb response.

gimaker:
@mattnworb Note added to docs.

davidxia:
tests pass

PR #1082, opened 2017-01-26
JobValidator: fix NPE if Job has malformed registration

mattnworb:
Inspired by a real-life example where someone accidentally included the
`volumes` section in the `registration` section. The JobValidator does
not check that each `registration` entry has a non-null `getPorts()`
Map.


mattnworb:
@spotify/helios-team @marcusb

PR #1083, opened 2017-02-02
don't log every time OldJobReaper runs

mattnworb:
rather than logging a message about what it is examining, just log if a
job is being reaped. This aims to reduce noise in the logs.


mattnworb:

PR #1084, opened 2017-02-14
simplify ZooKeeperHealthChecker

mattnworb:
Instead of deriving if the zookeeper client is connected or not based on
listening to ConnectionState change events, simply ask the
CuratorZookeeperClient if it is connected or not.

This is an attempt to alleviate issues we see in production where this
healthcheck on a master node will report something like `connection
state is SUSPENDED, state changed at 2017-02-08T17:07:31.029Z` - for
many days (6 at current count) - even though normal API operations on
the master that involve querying zookeeper are successful.

If there is some problem where the ConnectionStateListenable is not
firing in certain cases, of course it is possible that the
`isConnected()` method would suffer from the same problem - but it also
might be more accurate, with less moving pieces in the Healthcheck code.


mattnworb:

gimaker:
:+1: even if the new approach suffers from the same problem it's still less code and less complicated

PR #1085, opened 2017-02-16
Make helios-client guava 20 compatible.

daniel-kristjansson:
This changes some guava transform calls to transformAsync so we're compatible with guava 20.

We still pull in guava 19 for now.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
+++ b/helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
@@ -26,6 +26,7 @@ import static com.google.common.base.Strings.isNullOrEmpty;
 import static com.google.common.util.concurrent.Futures.catching;
 import static com.google.common.util.concurrent.Futures.immediateFuture;
 import static com.google.common.util.concurrent.Futures.transform;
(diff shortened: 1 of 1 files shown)


davidxia:
why didn't tests run?

davidxia:
Tests don't seem to run for forked branches.

mattnworb:
@davidxia there is a setting in CircleCI for only building pull requests that come from the same repo, which makes sense security-wise in case your CircleCI build has any sensitive info in it like encrypted env vars, etc. I don't think we have any of those so the setting could probably be disabled.

davidxia:
@mattnworb Thanks. I've changed that.

davidxia:
@rohansingh Just curious. Any idea why we don't use `AsyncFunction` in `HelisoClient` in places like https://github.com/spotify/helios/blob/master/helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java#L273?

mattnworb:
@davidxia `AsyncFunction` is a function that returns a `ListenableFuture<T>`. If you don't have a future to return, then you just use `Function`.

PR #1088, opened 2017-03-17
Expose docker connection pool size as agent argument

christoe:
This PR makes it possible to set the Docker client connection pool size in a Helios Agent command line argument. The default value has been left unchanged.

We need this as we're running enough containers on each host to exhaust the default number (100).

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/agent/AgentParser.java
+++ b/helios-services/src/main/java/com/spotify/helios/agent/AgentParser.java
@@ -70,6 +70,7 @@ public class AgentParser extends ServiceParser {
   private Argument zkAclMasterDigest;
   private Argument zkAclAgentPassword;
   private Argument disableJobHistory;
+  private Argument connectionPoolSize;
 
   public AgentParser(final String... args) throws ArgumentParserException {
     super("helios-agent", "Spotify Helios Agent", args);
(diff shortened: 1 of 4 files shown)


helios-services/src/main/java/com/spotify/helios/agent/AgentParser.java
```diff
@@ -273,6 +275,11 @@ protected void addArgs(final ArgumentParser parser) {
         .action(storeTrue())
         .setDefault(false)
         .help("If specified, the agent won't write job histories to ZooKeeper.");
+
+    connectionPoolSize = parser.addArgument("--connection-pool-size")
```
mattnworb: can you call the flag (and the field names) `--docker-connection-pool-size` instead to be very clear about what the connection pool is for?

helios-services/src/main/java/com/spotify/helios/agent/AgentService.java
```diff
@@ -347,10 +348,11 @@ public AgentService(final AgentConfig config, final Environment environment)
   }
 
   private DockerClient createDockerClient(final AgentConfig config) {
-    final DockerClient dockerClient;
-    if (isNullOrEmpty(config.getDockerHost().dockerCertPath())) {
-      dockerClient = new PollingDockerClient(config.getDockerHost().uri());
-    } else {
+    final DefaultDockerClient.Builder builder = DefaultDockerClient.builder()
+            .uri(config.getDockerHost().uri())
+            .connectionPoolSize(config.getConnectionPoolSize());
```
mattnworb: I think it would be better to not call `.connectionPoolSize(int)` at all if the flag is not specified for helios-agent, that way the default value when not specified still resides within the DefaultDockerClient.Builder class as it does today.

christoe:
@mattnworb: Valid comments. I have pushed the requested changes.

PR #1093, opened 2017-03-20
additional logging around host values used for Docker for Mac

mattnworb:
attemping to log more information around why an address of `127.0.0.1`
is returned by `temporaryJob.address("port")` sometimes on Docker for
Mac.

Also move some logging statements closer to where they are actually
done, and remove an old TODO


mattnworb:
@spotify/helios-team

PR #1095, opened 2017-03-28
master: add 'hostPattern' parameter to GET /jobs

mattnworb:
Add the ability to filter on the server-side the output from `GET /jobs`
to jobs deployed to specific hosts.

This is intended to improve the performance of the
`helios status --host` command, which currently runs throught the
following steps on the client-side in order to output the status of jobs
deployed on hosts that match the `--host` name pattern:

1. Get all jobs in the cluster (GET /jobs)
2. Get the job status of all these jobs (POST /statuses)
3. Filter the response to jobs deployed on hosts that match the filter

In a cluster with thousands of hosts, this is quite inefficient and
results in a lot of data being transferred that the CLI ends up
discarding. As an example, in a cluster with ~3000 jobs, the first
request ends up transferring more than 3MB of JSON, while the second
request transfers 14MB of JSON.

This can be improved by moving the filtering onto the server-side, so
that the CLI can simply ask for jobs that are deployed on a given host
name pattern in step 1 above. This also allows the server to not bother
to make thousands of zookeeper calls when finding the status of all
jobs.

I'll create a separate PR for using this new functionality in the HeliosClient
and in the CLI, so that we can merge and release these separately.

### DIFF
--- a/helios-services/src/test/java/com/spotify/helios/master/resources/JobsResourceTest.java
+++ b/helios-services/src/test/java/com/spotify/helios/master/resources/JobsResourceTest.java
@@ -1,0 +1,121 @@
+/*-
+ * -\-\-
+ * Helios Services
+ * --
+ * Copyright (C) 2016 - 2017 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.helios.master.resources;
+
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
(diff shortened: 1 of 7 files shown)


mattnworb:
@spotify/helios-team

helios-services/src/test/java/com/spotify/helios/master/resources/JobsResourceTest.java
```diff
@@ -0,0 +1,142 @@
+/*-
+ * -\-\-
```
davidxia: Double copyright header.

helios-services/src/test/java/com/spotify/helios/master/resources/JobsResourceTest.java
```diff
@@ -0,0 +1,142 @@
...
+    final HostStatus hostStatus = mockHostStatus(ImmutableMap.of(
+        jobId1, Deployment.of(jobId1, Goal.START),
+        jobId2, Deployment.of(jobId1, Goal.START)
+    ));
+    when(model.getHostStatus("foobar.example.net")).thenReturn(hostStatus);
+
+    when(model.getJob(jobId1)).thenReturn(job1);
+    when(model.getJob(jobId2)).thenReturn(job2);
+
+    assertThat(resource.list("", namePattern), is(
```
davidxia: Test a host pattern that returns an empty list of jobs too?

mattnworb: an empty map should be treated the same, not sure that adds much value

helios-services/src/test/java/com/spotify/helios/master/resources/JobsResourceTest.java
mattnworb: fixed, thanks

PR #1096, opened 2017-03-28
use more efficient API for job status with host filter in CLI and client

mattnworb:
Uses the new API method introduced in #1095 in HeliosClient and in
JobStatusCommand so that `helios status --host foo` does not need to
fetch all jobs, the job status for all jobs, and then filter by hostname
within the CLI. The filtering is all server-side as of #1095.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
+++ b/helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
@@ -44,6 +44,7 @@ import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.type.TypeFactory;
 import com.google.common.base.Function;
 import com.google.common.base.Optional;
+import com.google.common.base.Strings;
 import com.google.common.base.Supplier;
 import com.google.common.base.Suppliers;
 import com.google.common.base.Throwables;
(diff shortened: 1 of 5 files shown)


mattnworb:
To be merged/released after #1095 

@spotify/helios-team

mattnworb:
note that the test `testFilteringJobAndHostStatuses - com.spotify.helios.system.MultipleHostsTest` fails because this commit is on top of master and not #1095 so the test does not have the new API logic to test against; I'll rebase this once #1095 is merged.

mattnworb:
rebased on top of master now that #1095 is merged

helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
```diff
@@ -121,6 +122,9 @@
   private final RequestDispatcher dispatcher;
   private final AtomicBoolean versionWarningLogged = new AtomicBoolean();
 
+  private final TypeReference<Map<JobId, Job>> jobIdMap = new TypeReference<Map<JobId, Job>>() {
```
gimaker: nitpick: you could just make this local (it's used in single place afaict)

mattnworb: good point; originally I did not have the three overloads of `jobs()` calling each other, so it was repeated three times in there before I condensed the logic down.

PR #1099, opened 2017-03-31
Upgrade logback for faster CLI startup

rohansingh:
logback calls `InetAddress.getLocalHost()`, which is painfully slow on
some versions of macOS (i.e., 4-5 seconds). Since this happens when
logging is being configured, it potentially adds seconds to every
`helios` CLI invocation.

To address this, we upgrade to a newer version of logback that includes
a fix to evaluate the hostname lazily and not block startup:
https://jira.qos.ch/browse/LOGBACK-1221

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -145,7 +145,7 @@
       <dependency>
         <groupId>org.slf4j</groupId>
         <artifactId>slf4j-api</artifactId>
-        <version>1.7.21</version>
+        <version>1.7.22</version>
       </dependency>
       <dependency>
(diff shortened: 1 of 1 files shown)


rohansingh:
@spotify/helios-team

rohansingh:
The relevant fix is in https://github.com/qos-ch/logback/commit/3f79c0e1488be9.

mattnworb:
👍 

( would love to know why this is the case though: `it may stuck for seconds in InetAddress.getLocalHost()`)

PR #1100, opened 2017-04-03
JobsResource: Fix bug in new `list` implementation

rohansingh:
Our new, more efficient implementation of `JobsResource.list` would fail
if the same job was deployed to multiple hosts.


rohansingh:
@spotify/helios-team

caipre:
👍

PR #1102, opened 2017-04-06
replace deprecated Throwables.propagate(Throwable)

mattnworb:
This method was deprecated in Guava 20; see
https://github.com/google/guava/wiki/Why-we-deprecated-Throwables.propagate

No functional changes here (I hope), it just removes a ton of deprecation warnings.


mattnworb:
@spotify/helios-team

PR #1103, opened 2017-04-07
add rolling-update --ignore-failures, skip over AWAIT_RUNNING

mattnworb:
Internally at Spotify we have a use case where a deployment group with hundreds of hosts in the group frequently has trouble completing a `rolling-update` command successfully to deploy a new version of a job because of various issues with individual hosts out of the hundreds (and sometimes thousands) of hosts in the group that prevent a new job from successfully running. Issues we have seen have to do with unhealthy dockerd instances, etc.

This deployment group represents a critical service where the service owners want to ensure that the deploymentGroup's status is never `FAILED`, as it is necessary for SLA reasons to make sure that new/auto-scaled instances get the job assigned to the deployment group deployed to it.

To that end we have decided to add a `--ignore-failures` flag to the `helios rolling-update` command that should cause the rollout of the new image to continue to all hosts in the deployment group even if the deployment/rollout fails on any single host in the group. We picture that very few services will ever want to use this flag besides this particular use case.

This PR is one possible way that we could implement this behavior in the heliosmaster: by completely removing the `RolloutTask.Action.AWAIT_RUNNING` step from the list of rollout tasks when the rollout plan is generated (see `RollingUpdatePlanner`). 

This would mean in practice that a `helios rolling-update --ignore-failures` quickly rolls through every host in the group, undeploying the previous job and deploying the new job, without waiting for any host to even start the image pull / container creation / etc process.

There are probably other ways we could implement this beyond removing `RolloutTask.Action.AWAIT_RUNNING` from the list of rollout tasks; I'd be happy to discuss and review various other options.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdatePlanner.java
+++ b/helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdatePlanner.java
@@ -25,6 +25,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.spotify.helios.common.descriptors.DeploymentGroup;
+import com.spotify.helios.common.descriptors.RolloutOptions;
 import com.spotify.helios.common.descriptors.RolloutTask;
 import java.util.List;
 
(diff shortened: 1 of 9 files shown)


mattnworb:
@spotify/helios-team

mattnworb:
One possible issue I see with this approach now that I have written up a description of how it would work is that I am not sure if any problems/failures to _undeploy_ the old job would still fail the group if `--ignore-failures` is present. This change skips over the AWAIT_RUNNING part of the plan, but doesn't skip over the parts of the rollout logic that set DeploymentGroupStatus.state to FAILED.

helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdatePlanner.java
```diff
@@ -44,37 +45,45 @@ public static RollingUpdatePlanner of(final DeploymentGroup deploymentGroup) {
...
     for (final String host : hosts) {
       result.add(RolloutTask.of(RolloutTask.Action.UNDEPLOY_OLD_JOBS, host));
       result.add(RolloutTask.of(RolloutTask.Action.DEPLOY_NEW_JOB, host));
     }
-    for (final String host : hosts) {
-      result.add(RolloutTask.of(RolloutTask.Action.AWAIT_RUNNING, host));
+    // TODO (mbrown): is AWAIT_RUNNING what we want to skip?
+    if (!ignoreFailures) {
+      for (final String host : hosts) {
+        result.add(RolloutTask.of(RolloutTask.Action.AWAIT_RUNNING, host));
```
mattnworb: need to add tests of this logic / flag

mattnworb:
perhaps a more complete approach would leave most of the current logic in place (list of tasks would still include AWAIT_RUNNING) but [`RollingUpdateOpFactory.error()` would actually call `nextTask()` instead of failing/aborting/setting the group's status to FAILED](https://github.com/spotify/helios/blob/0831a61237ec64f8011e20d972c64569035f1206/helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdateOpFactory.java#L156-L165).

mattnworb:
on second thought the above comment probably wouldn't work as [there are places](https://github.com/spotify/helios/blob/0831a61237ec64f8011e20d972c64569035f1206/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java#L941-L952) where `nextTask(List<ZookeeperOperations)` is called (by ZookeeperMasterModel) which relies on the caller knowing what ZK operations should be done as a part of the next task in the plan - the OpFactory isn't fully in charge of what happens "next".

mattnworb:
With the approach in this PR, the rolling-update would fail under the following:

- when processing RolloutTask.Action.DEPLOY, if
   - the job no longer exists
   - token is invalid
   - host is not found 
   - a JobPortAllocationConflictException occurs (i.e. try to allocate a static port already assigned to a different job on the same host)

as far as I can tell all other places where `opFactory.error(..)` is called are from code paths where the master is processing a `RolloutTask.Action.AWAIT_RUNNING`

helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdatePlanner.java
mattnworb: this is done now

helios-tools/src/main/java/com/spotify/helios/cli/command/RollingUpdateCommand.java
```diff
@@ -131,6 +132,15 @@ public Long get() {
...
         .help("Insecure access token meant to prevent accidental changes to your job "
               + "(e.g. undeploys).");
+
+    ignoreFailuresArg = parser.addArgument("--ignore-failures")
+        .setDefault(false)
+        .action(storeTrue())
+        .help("When specified, the rolling update will continue on to all hosts in the deployment "
+              + "group without waiting for the job to reach the RUNNING state on each host. "
+              + "Be *very* careful about using this option, as it has the potential to completely "
+              + "take down your service by rolling out a broken job to all of the hosts in your "
```
gimaker: Perhaps mention that this also effectively makes the rollout "fully parallel"? Specifying `--ignore--failures` will, aside from ignoring errors, also cause similar behaviours as when using e.g. `--par 10000`

mattnworb: good point

mattnworb:
@gimaker would you have any preference between this approach and the AWAIT_TERMINAL one you mention? I think the latter might involve more complexity inside helios-master, but might be better from the user perspective.

gimaker:
> would you have any preference between this approach and the AWAIT_TERMINAL one you mention? I think the latter might involve more complexity inside helios-master, but might be better from the user perspective.

@mattnworb Either approach is fine with me. I have a weak preference for the "AWAIT_TERMINAL" approach (a shitty name that'd need changing, since it doesn't accurately capture what it'd do) because with that approach the `--par` flag still kind of makes sense (== it'd wait for the image to pull (or fail to pull) before continuing to the next host).

mattnworb:
just thinking out loud, but what should a theoretical AWAIT_TERMINAL action do if (for example) the image failed to pull before the timeout elapsed? Seems like we would not want to fail the rollout then, as a host with a broker docker instance will probably never get out of `TaskStatus.State.PULLING_IMAGE`.

gimaker:
@mattnworb Fair point. Having to deal with that edge case does add extra complexity. Perhaps it's best to keep it simple and stick with the current approach - at least we understand the implications of it (?).

mattnworb:
btw if we were to go with this approach instead of #1105 I would like to rename the flag to be `helios rolling-update --do-not-await-running` to be more accurate

gimaker:
> btw if we were to go with this approach instead of #1105 I would like to rename the flag to be helios rolling-update --do-not-await-running to be more accurate

:+1:

helios-tools/src/main/java/com/spotify/helios/cli/command/RollingUpdateCommand.java
davidxia: Would it help the user to enumerate all the checks that are skipped that are usually checked for AWAIT_RUNNING? e.g. retrieving job status, job unexpectedly undeployed, etc.

mattnworb: I think that renaming this `--do-not-await-running` as mentioned above would make it more clear

mattnworb:
closing in favor of #1105

davidxia:
did you mean to actually close this PR?

PR #1105, opened 2017-04-10
do not set DeploymentGroup to FAILED when ignoreFailure is true

mattnworb:
When rolloutOptions.ignoreFailures is true, the RollingUpdateOpFactory will
quietly transition to the next task in the deployment group's rollout plan
(the DeploymentGroupTasks object) when one of the overlaods for `error(..)`
is called.

This will have the effect of moving the rolling update to the next task in
the plan when the group/rollout otherwise would have FAILED, only if the
ignoreFailures option is set to true.

At first glance one problem with this method of internally returning the
result of `nextTask()` when `error(..)` is that sometimes `nextTasks` is
called with a list of ZookeeperOperations supplied as a parameter by the
calling code; however, this only occurs when the ZookeeperMasterModel is
first processing an action of DEPLOY, MARK_UNDEPLOYED or UNDEPLOYED. The
overload of `nextTasks()` with no argument is called whenever the task
action is AWAIT_RUNNING and the task reaches running, therefore it should
be safe (unless the callers are refactored).

This is an alternate approach to #1103 to try to achieve the same goals but
with less downside.

### DIFF
--- a/helios-services/src/test/java/com/spotify/helios/rollingupdate/RollingUpdateOpFactoryTest.java
+++ b/helios-services/src/test/java/com/spotify/helios/rollingupdate/RollingUpdateOpFactoryTest.java
@@ -22,7 +22,11 @@ package com.spotify.helios.rollingupdate;
 
 import static com.spotify.helios.common.descriptors.DeploymentGroup.RollingUpdateReason.HOSTS_CHANGED;
 import static com.spotify.helios.common.descriptors.DeploymentGroup.RollingUpdateReason.MANUAL;
+import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.hamcrest.Matchers.empty;
+import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
 import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.anyString;
 import static org.mockito.Matchers.eq;
@@ -30,6 +34,7 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
@@ -44,9 +49,13 @@ import com.spotify.helios.servicescommon.coordination.SetData;
 import com.spotify.helios.servicescommon.coordination.ZooKeeperClient;
 import com.spotify.helios.servicescommon.coordination.ZooKeeperOperation;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
+import java.util.Set;
 import org.apache.zookeeper.data.Stat;
+import org.hamcrest.CustomTypeSafeMatcher;
+import org.hamcrest.Matcher;
 import org.junit.Test;
 
 public class RollingUpdateOpFactoryTest {
@@ -234,7 +243,7 @@ public class RollingUpdateOpFactoryTest {
         deploymentGroupTasks, eventFactory);
     final RollingUpdateOp op = opFactory.nextTask();
 
-    // A nexTask op with no ZK operations should result advancing the task index
+    // A nextTask op with no ZK operations should result advancing the task index
     assertEquals(1, op.operations().size());
     assertEquals(new SetData("/status/deployment-group-tasks/my_group",
                              deploymentGroupTasks.toBuilder()
@@ -379,5 +388,44 @@ public class RollingUpdateOpFactoryTest {
 
     assertEquals(0, op.operations().size());
     assertEquals(0, op.events().size());
+  }
+
+  @Test
+  public void testErrorWhenIgnoreFailuresIsTrue() {
+    final DeploymentGroup deploymentGroup = DeploymentGroup.newBuilder()
+        .setName("ignore_failure_group")
+        .setRolloutOptions(RolloutOptions.newBuilder()
+            .setIgnoreFailures(true)
+            .build()
+        )
+        .setRollingUpdateReason(MANUAL)
+        .build();
+
+    // the current task is the AWAIT_RUNNING one
+    final DeploymentGroupTasks tasks = DeploymentGroupTasks.newBuilder()
+        .setTaskIndex(2)
+        .setRolloutTasks(ImmutableList.of(
+            RolloutTask.of(RolloutTask.Action.UNDEPLOY_OLD_JOBS, "host1"),
+            RolloutTask.of(RolloutTask.Action.DEPLOY_NEW_JOB, "host1"),
+            RolloutTask.of(RolloutTask.Action.AWAIT_RUNNING, "host1")
+        ))
+        .setDeploymentGroup(deploymentGroup)
+        .build();
+
+    final RollingUpdateOpFactory opFactory = new RollingUpdateOpFactory(tasks, eventFactory);
+
+    final RollingUpdateOp nextOp = opFactory.error("something went wrong", "host1",
+        RollingUpdateError.TIMED_OUT_WAITING_FOR_JOB_TO_REACH_RUNNING);
+
+    assertThat(nextOp.operations(), containsInAnyOrder(
+        new SetData("/status/deployment-groups/ignore_failure_group",
+            DeploymentGroupStatus.newBuilder()
+                .setState(DeploymentGroupStatus.State.DONE)
+                .setError(null)
+                .build()
+                .toJsonBytes()
+        ),
+        new Delete("/status/deployment-group-tasks/ignore_failure_group")
+    ));
   }
 }
--- a/helios-client/src/main/java/com/spotify/helios/common/descriptors/RolloutOptions.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/descriptors/RolloutOptions.java
@@ -38,7 +38,8 @@ import org.jetbrains.annotations.Nullable;
  *   "parallelism": 2,
  *   "timeout": 1000,
  *   "overlap": true,
- *   "token": "insecure-access-token"
+ *   "token": "insecure-access-token",
+ *   "ignoreFailures": false
  * }
  * </pre>
  */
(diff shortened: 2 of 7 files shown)


mattnworb:
the first commit (to add this to the CLI and client) is the same; only the server-side implementation is changed compared to #1103 

@gimaker

mattnworb:
@spotify/helios-team

helios-services/src/test/java/com/spotify/helios/rollingupdate/RollingUpdateOpFactoryTest.java
```diff
@@ -380,4 +389,43 @@ public void testYield() {
...
+        .setDeploymentGroup(deploymentGroup)
+        .build();
+
+    final RollingUpdateOpFactory opFactory = new RollingUpdateOpFactory(tasks, eventFactory);
+
+    final RollingUpdateOp nextOp = opFactory.error("something went wrong", "host1",
+        RollingUpdateError.TIMED_OUT_WAITING_FOR_JOB_TO_REACH_RUNNING);
+
+    assertThat(nextOp.operations(), containsInAnyOrder(
+        new SetData("/status/deployment-groups/ignore_failure_group",
```
davidxia: You can use `Paths.statusDeploymentGroup("ignore_failure_group")` here and `Paths.statusDeploymentGroupTasks(` for line 428.

helios-services/src/test/java/com/spotify/helios/rollingupdate/RollingUpdateOpFactoryTest.java
```diff
@@ -380,4 +389,43 @@ public void testYield() {
...
+        .setRollingUpdateReason(MANUAL)
+        .build();
+
+    // the current task is the AWAIT_RUNNING one
+    final DeploymentGroupTasks tasks = DeploymentGroupTasks.newBuilder()
+        .setTaskIndex(2)
+        .setRolloutTasks(ImmutableList.of(
+            RolloutTask.of(RolloutTask.Action.UNDEPLOY_OLD_JOBS, "host1"),
+            RolloutTask.of(RolloutTask.Action.DEPLOY_NEW_JOB, "host1"),
+            RolloutTask.of(RolloutTask.Action.AWAIT_RUNNING, "host1")
```
davidxia: Would it worth testing this too? Maybe it's too similar?

```java
    // the current task is the AWAIT_RUNNING one
    final DeploymentGroupTasks tasks = DeploymentGroupTasks.newBuilder()
        .setTaskIndex(2)
        .setRolloutTasks(ImmutableList.of(
            RolloutTask.of(RolloutTask.Action.UNDEPLOY_OLD_JOBS, "host1"),
            RolloutTask.of(RolloutTask.Action.DEPLOY_NEW_JOB, "host1"),
            RolloutTask.of(RolloutTask.Action.AWAIT_RUNNING, "host1"),
            RolloutTask.of(RolloutTask.Action.UNDEPLOY_OLD_JOBS, "host2"),
            RolloutTask.of(RolloutTask.Action.DEPLOY_NEW_JOB, "host2"),
            RolloutTask.of(RolloutTask.Action.AWAIT_RUNNING, "host2")
        ))
        .setDeploymentGroup(deploymentGroup)
        .build();

    final RollingUpdateOpFactory opFactory = new RollingUpdateOpFactory(tasks, eventFactory);

    final RollingUpdateOp nextOp = opFactory.error("something went wrong", "host1",
        RollingUpdateError.TIMED_OUT_WAITING_FOR_JOB_TO_REACH_RUNNING);

    assertThat(nextOp.operations(), contains(
        new SetData(Paths.statusDeploymentGroupTasks("ignore_failure_group"),
            tasks.toBuilder()
                .setTaskIndex(3)
                .build()
                .toJsonBytes()
        )
 ```

**Comment by davidxia** [comment]
Add ignore-failures to user output as well.

* https://github.com/spotify/helios/blob/051622b2fb1bea279c40f655cb23892580345c9e/helios-tools/src/main/java/com/spotify/helios/cli/command/RollingUpdateCommand.java#L176
* https://github.com/spotify/helios/blob/051622b2fb1bea279c40f655cb23892580345c9e/helios-tools/src/main/java/com/spotify/helios/cli/command/RollingUpdateCommand.java#L186

**Comment by davidxia** [comment]
One difference I noticed between this and #1103 is that here a failing health check will need to [exceed the default rolling-update timeout duration](https://github.com/spotify/helios/blob/051622b2fb1bea279c40f655cb23892580345c9e/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java#L902) for AWAIT_RUNNING before the rollout moves onto the next host.

In #1103 a failing health check won't stall the rollout because the AWAIT_RUNNING task is [never inserted into the rollout plan](https://github.com/spotify/helios/pull/1103/files#diff-1016753d8158cfc30942a0034eeb7f75R69).

As a user, it'd be nice not to have to wait for health checks if I don't care if they fail anyways.

**Comment by davidxia** [comment]
Another difference is that now [all these errors](https://github.com/spotify/helios/blob/master/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java#L964-L973) will also be ignored. Which might be fine. But then the [description for the switch](https://github.com/spotify/helios/pull/1105/files#diff-09ebe80c4b7d83ee0a1323aed194d6baR142) should be updated to say instead of "take down your service by rolling out a broken job" that the job might not even be attempted to be deployed because, for example, empty token.

**Comment by mattnworb** [comment]
Testing this locally:

```
# two hosts in my cluster
$ bin/helios -z http://localhost:5801 hosts
running in helios project, using /Users/mattbrown/code/helios/helios-tools/target/helios-tools-0.9.0-SNAPSHOT-shaded.jar
HOST              STATUS           DEPLOYED    RUNNING    CPUS    MEM     LOAD AVG    MEM USAGE    OS                  HELIOS            DOCKER               LABELS
host2.            Up 1 minute      0           0          4       8 gb    1.85        1.00         Mac OS X 10.12.4    0.9.0-SNAPSHOT    17.03.1-ce (1.27)    foo=bar
mattbrown-air.    Up 30 seconds    0           0          4       8 gb    1.72        1.00         Mac OS X 10.12.4    0.9.0-SNAPSHOT    17.03.1-ce (1.27)    foo=bar

# create a DG for the two hosts
[mattbrown@mattbrown-air helios]$ bin/helios -z http://localhost:5801 create-deployment-group my-group foo=bar
running in helios project, using /Users/mattbrown/code/helios/helios-tools/target/helios-tools-0.9.0-SNAPSHOT-shaded.jar
Creating deployment group: {"hostSelectors":[{"label":"foo","operand":"bar","operator":"EQUALS"}],"name":"my-group"}
{"status":"CREATED"}

# create a job for an image that does not exist (as an example of something that would otherwise fail the rollout/group)
$ bin/helios -z http://localhost:5801 create failing-job:1 registry.spotify.net/dev/null:does-not-exist
running in helios project, using /Users/mattbrown/code/helios/helios-tools/target/helios-tools-0.9.0-SNAPSHOT-shaded.jar
Creating job: {"addCapabilities":[],"command":[],"created":null,"creatingUser":null,"dropCapabilities":[],"env":{},"expires":null,"gracePeriod":null,"healthCheck":null,"hostname":null,"id":"failing-job:1","image":"registry.spotify.net/dev/null:does-not-exist","metadata":{},"networkMode":null,"ports":{},"registration":{},"registrationDomain":"","resources":null,"secondsToWaitBeforeKill":null,"securityOpt":[],"token":"","volumes":{}}
Done.
failing-job:1:9044bc35b5183426216de2668c564c2267be97cc

# do a normal rollout of this job - the first host times out after 300 seconds because the image is not found, and then the group is failed
$ bin/helios -z http://localhost:5801 rolling-update failing-job:1 my-group
running in helios project, using /Users/mattbrown/code/helios/helios-tools/target/helios-tools-0.9.0-SNAPSHOT-shaded.jar
Rolling update started: my-group -> failing-job:1:9044bc3 (parallelism=1, timeout=300, overlap=false, token=)


Failed: host2: timed out waiting for job failing-job:1:9044bc35b5183426216de2668c564c2267be97cc to reach state RUNNING due to missing Docker image (terminal job state FAILED, previous states: PULLING_IMAGE->FAILED->PULLING_IMAGE->FAILED->PULLING_IMAGE->FAILED)
Duration: 303.24 s

# do the same operation with --ignore-failures
$ bin/helios -z http://localhost:5801 rolling-update --ignore-failures failing-job:1 my-group
running in helios project, using /Users/mattbrown/code/helios/helios-tools/target/helios-tools-0.9.0-SNAPSHOT-shaded.jar
Rolling update started: my-group -> failing-job:1:9044bc3 (parallelism=1, timeout=300, overlap=false, token=)


Done.
Duration: 606.91 s

# the rollout took 2 * 300 seconds, but was left in an ACTIVE state even though the job failed on each host:
$ bin/helios -z http://localhost:5801 deployment-group-status my-group
running in helios project, using /Users/mattbrown/code/helios/helios-tools/target/helios-tools-0.9.0-SNAPSHOT-shaded.jar
Name: my-group
Job Id: failing-job:1:9044bc3
Status: ACTIVE
Host selectors:
  foo = bar

HOST              UP-TO-DATE    JOB                      STATE
host2.            X             failing-job:1:9044bc3    FAILED
mattbrown-air.    X             failing-job:1:9044bc3    FAILED
```

With this change, rolling out a "bad" job to an individual host still takes the entire `timeout` duration - but a failure of a host does not impact the rest of the rollout or the final status of the group.

**Comment by mattnworb** [comment]
@davidxia 

>Add ignore-failures to user output as well.

done in 746da8a

**Comment by mattnworb** [comment]
> As a user, it'd be nice not to have to wait for health checks if I don't care if they fail anyways.

Yeah, I agree, but I am not sure the best way to find the balance between the way this implementation (follow through the normal steps and respect the parallelism setting, but never set the state to FAILED) versus what #1103 did (remove AWAIT_RUNNING and just roll through every undeploy/deploy practically at the same time).

I think this might be the better choice so that we can a) be sure the status is never failed if the user told us to `--ignore-failures` and b) still respect parallelism/overlap/etc settings so that the rolling-update has the same general plan as always, except that the failure is ignored.

**Comment by mattnworb** [comment]
@davidxia I've also updated the field's help message:

```
When specified, the rolling-update will ignore *all* failures and will proceed
to deploying the job to all hosts in the deployment group. The rolling-update
will go through the normal rollout plan (respecting the --par and --overlap
settings), and will wait for the job to reach RUNNING on each host as normal;
however, any failure that would otherwise cause the rolling-update to abort and
set the deployment group's status to FAILED is *ignored*. Be *VERY* careful
about using this option, as it has the potential to completely take down your
service by rolling out a broken job to all of the hosts in your group.
```

PR #1108, opened 2017-04-11
remove version check and use of docker tag -f

### LINKED ISSUE #1106: helios-use latest fails with "unknown shorthand flag"
```
> helios-use latest
Switching to latest
latest: Pulling from spotify/helios-solo
Digest: sha256:89bdcc8bcb2485ccee0481382e8ece449a04aec12e0e0cf878ee0dd05f5a92a3
Status: Image is up to date for spotify/helios-solo:latest
unknown shorthand flag: 'f' in -f
See 'docker tag --help'.
```

```
> docker -v
Docker version 17.03.1-ce, build c6d412e
```

mattnworb:
The new docker version numbering scheme (`17.03.1-ce` etc) breaks the assumption in this script that it is safe to use `tag -f` if the second number in the version string is < 12.

To fix this, remove the version check altogether. Fixes #1106.

### DIFF
--- a/solo/helios-use
+++ b/solo/helios-use
@@ -30,14 +30,7 @@ if [ ! -z "$1" ]; then
   echo "Switching to $requested_version"
   docker pull "$REPO:$requested_version"
 
-  client_version=$(docker version --format '{{.Client.Version}}')
-  cv_arr=( ${client_version//./ } )  # replace periods and split into array
-  # Docker 1.12+ doesn't have the -f switch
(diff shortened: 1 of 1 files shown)


mattnworb:
@spotify/helios-team

davidxia:
Is the force implied now?

mattnworb:
i think so https://docs.docker.com/engine/deprecated/#f-flag-on-docker-tag

PR #1114, opened 2017-04-20
Allow adding labels to containers

carlpett:
Hi!
This PR adds the ability to set labels on the containers started by Helios. There is probably a bunch of work remaining, I haven't been able to successfully start the integration tests on my machine yet, and the building and starting of helios-solo doesn't seem to work on Windows with boot2docker? (I can understand if this is not a configuration that you test...)

But while I get that ironed out, it'd be great if you have any feedback on places I have missed to add this, or more tests that need to be written?

### DIFF
--- a/helios-tools/src/main/java/com/spotify/helios/cli/command/JobCreateCommand.java
+++ b/helios-tools/src/main/java/com/spotify/helios/cli/command/JobCreateCommand.java
@@ -123,6 +123,7 @@ public class JobCreateCommand extends ControlCommand {
   private final Argument metadataArg;
   private final Argument addCapabilityArg;
   private final Argument dropCapabilityArg;
+  private final Argument labelsArg;
   private final Supplier<Map<String, String>> envVarSupplier;
 
   public JobCreateCommand(final Subparser parser) {
@@ -270,6 +271,12 @@ public class JobCreateCommand extends ControlCommand {
         .setDefault(new ArrayList<String>())
         .help("The Linux capabilities this Helios job drops from its Docker container. "
               + "Defaults to nothing.");
+
+    labelsArg = parser.addArgument("--labels")
+            .action(append())
+            .setDefault(new ArrayList<String>())
+            .help("Labels (key-value pairs) to apply onto the container. "
+                    + "Defaults to nothing.");
 
     this.envVarSupplier = envVarSupplier;
   }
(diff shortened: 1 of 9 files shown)


mattnworb:
Hey @carlpett , thanks for the PR! Can you describe a bit more about the motivation behind this change?

carlpett:
Hi!
We want to be able to set labels on the containers we deploy, and we also want to be able to set them after building the container (ie, not with a `LABEL` statement in the Dockerfile). We use this as metadata in our monitoring (cadvisor and similar expose the labels along with metrics).

mattnworb:
Thanks for the explanation. So it would be sufficient for you if Helios set these labels when creating the containers but didn't do anything else with the labels, such as letting you query for jobs that match certain labels, expose this in CLI commands, etc etc ?

carlpett:
Yup, we're pretty much only interested in the creation part, the inspection parts aren't really necessary for us.

helios-tools/src/main/java/com/spotify/helios/cli/command/JobCreateCommand.java
```diff
@@ -541,6 +548,13 @@ int run(final Namespace options, final HeliosClient client, final PrintStream ou
       builder.setDropCapabilities(dropCaps);
     }
 
+    final Map<String, String> labels = Maps.newHashMap();
```
carlpett: Indeed. Fixed!

carlpett:
Hm, did my change actually cause that, or do the tests sometime time out?

mattnworb:
@carlpett no, the tests are a bit flaky - I restarted the CircleCI build and it looks green now.

For the actual contents of the PR, it looks pretty good to me and also like a pretty straightforward change. One aspect of it that I am concerned about is that we have had bad experiences in the past when adding new fields to the Job class that causes the JobId.hash field to change such that an older client/CLI submitting a new Job calculates the hash based on different parameters than the up-to-date server does, which causes the CreateJobResource to respond with an error.

See for instance https://github.com/spotify/helios/pull/1075 and https://github.com/spotify/helios/pull/1076 - but since the JobIdHashIsStableTest test looks like it passes for your change, maybe that is not a huge concern. I'll have to look over the relevant code there again to be sure though.

mattnworb:
One test that would be good to add would be something in TaskConfigTest to verify that the labels are carried through to the ContainerConfig instance as expected. I wouldn't worry too much about running helios-solo or the integration tests (although sorry to hear they don't work well on Windows, although I am not surprised either - we don't have anyone who develops on Windows machines here AFAIK).

carlpett:
Good idea. Looked to me like none of the existing test cases really matched this, so I added a new test that validates that the ContainerConfig matches the job sent to the TaskConfig builder. Currently only checks the new labels field, though.

I think that this change does not break the JobIdHashIsStableTest since it assigns a default empty map, which should be omitted from the json according to the documentation?

mattnworb:
> I think that this change does not break the JobIdHashIsStableTest since it assigns a default empty map, which should be omitted from the json according to the documentation?

Yep that is exactly correct. For future reference (as I'll probably forget some day), [only non-empty fields are used in the `Json.sha1digest(object)` method](https://github.com/spotify/helios/blob/e4374120393cdb8c7f65382ef30a922cc66a074b/helios-client/src/main/java/com/spotify/helios/common/Json.java#L329) that `Job.Builder` calls to compute the hash of a job.

mattnworb:
@carlpett thanks again for the contribution - this should be available in `0.9.123`.

PR #1115, opened 2017-04-24
ensure all output from HostListCommand is sorted

mattnworb:
In the default cause of a command like `helios hosts`, the output
currently is not sorted at all by the command - the iteration order of
the Map returned by `client.hostStatuses()` controls the output of
hosts/rows printed to stdout.

In every other case (using `--json` or `-q`) the output *is* sorted.

This change makes it so that the Map used from `client.hostStatuses()`
is always sorted, and slightly modifies the test setup so that we can
verify that the HostListCommand is sorting data if it is returned in an
unsorted way.


mattnworb:
@spotify/helios-team

PR #1120, opened 2017-05-24
integrate with Google ContainerRegistry support in docker-client

mattnworb:
Updates docker-client to 8.6.0 which has support for configuring the
DockerClient instance to authenticate against Google Container Registry.

(Note: this requires https://github.com/spotify/docker-client/pull/762 to be merged and released in docker-client first, before this PR can be merged)

The credentials to use in authenticating for pulls from GCR are
configured with a new flag `--docker-gcp-account-credentials`. The GCR
authentication layer is only enabled when this flag is set. This
argument value is expected to be a file containing the JSON key for the
service or user account to pull images with.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/agent/AgentConfig.java
+++ b/helios-services/src/main/java/com/spotify/helios/agent/AgentConfig.java
@@ -20,6 +20,7 @@
 
 package com.spotify.helios.agent;
 
+import com.google.auth.oauth2.GoogleCredentials;
 import com.spotify.docker.client.DockerHost;
 import com.spotify.helios.servicescommon.CommonConfiguration;
 import com.spotify.helios.servicescommon.FastForwardConfig;
@@ -67,6 +68,9 @@ public class AgentConfig extends CommonConfiguration<AgentConfig> {
   private List<String> extraHosts;
   private boolean jobHistoryDisabled;
   private int connectionPoolSize;
+
+  /** Credentials to use with Google Container Registry. */
+  private GoogleCredentials googleCredentials;
 
   public boolean isInhibitMetrics() {
     return inhibitMetrics;
@@ -367,6 +371,15 @@ public class AgentConfig extends CommonConfiguration<AgentConfig> {
 
   public AgentConfig setConnectionPoolSize(final int connectionPoolSize) {
     this.connectionPoolSize = connectionPoolSize;
+    return this;
+  }
+
+  public GoogleCredentials getGoogleCredentials() {
+    return googleCredentials;
+  }
+
+  public AgentConfig setGoogleCredentials(final GoogleCredentials googleCredentials) {
+    this.googleCredentials = googleCredentials;
     return this;
   }
 }
--- a/pom.xml
+++ b/pom.xml
@@ -137,7 +137,7 @@
       <dependency>
         <groupId>com.spotify</groupId>
         <artifactId>docker-client</artifactId>
-        <version>8.4.0</version>
+        <version>8.6.0</version>
         <classifier>shaded</classifier>
       </dependency>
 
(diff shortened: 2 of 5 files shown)


pom.xml
```diff
@@ -137,7 +137,7 @@
       <dependency>
         <groupId>com.spotify</groupId>
         <artifactId>docker-client</artifactId>
-        <version>8.4.0</version>
+        <version>8.6.0-SNAPSHOT</version>
```
mattnworb: will need to change to `8.6.0` once the new stuff in docker-client is released

mattnworb:
I made sure that this logs what file is being read at startup (and the identity in it) for troubleshooting:

```
10:50:22.887 helios[7111]: INFO  [AgentMain STARTING] AgentService: loading Google Cloud credentials for Docker from /Users/mattbrown/Downloads/xxx.json
10:50:23.036 helios[7111]: INFO  [AgentMain STARTING] ContainerRegistryAuthSupplier: loaded credentials for service account with clientEmail=mbrown-test-helios-gcr@xxxx.iam.gserviceaccount.com
```

helios-services/src/main/java/com/spotify/helios/agent/AgentConfig.java
```diff
@@ -369,4 +371,13 @@ public AgentConfig setConnectionPoolSize(final int connectionPoolSize) {
     this.connectionPoolSize = connectionPoolSize;
     return this;
   }
+
+  public File getGoogleCloudAccountCredentials() {
```
gimaker: nitpick: I think it would make more sense to have this be of type `GoogleCredentials` and load the credential file elsewhere (e.g. in `AgentParser`).

That would make it easier to add a flag like `--docker-gcp-use-application-default-credentials` in the future if we want to do that.

mattnworb: Good point, although this is kind of difficult with the structure of ContainerRegistryAuthSupplier that is currently in docker-client. It would need to be refactored too, to allow passing the GoogleCredentials in. I think this would be a better design, but I'll probably hold off on doing so until there is a need for Helios to use the ADC.

mattnworb:
I think the only test we could do here is test that AgentService constructs a DockerClient with this registryAuthSupplier instance when `agentConfig.getGoogleCloudAccountCredentials()` is set, but the auth info is not exposed by `DefaultDockerClient` today for a test to assert against, so we'd have to add all that 😞

helios-services/src/main/java/com/spotify/helios/agent/AgentConfig.java
gimaker: Derp.

That seems less than ideal. For example, if I have a build agent in GCP that both pulls and pushes images to GCR using something like `dockerfile-maven-plguin`. Then I can't use the SA that is associated with the VM? Instead I must create a SA key and add glue to make sure it's loaded and used?

mattnworb: no: https://github.com/spotify/docker-client/commit/b44708fc5bf78801c07849ea75410c1979264844

I just meant that I didn't think to add a factory like `ContainerRegistryAuthSupplier.forCredentials()`, all it has now is `forApplicationDefaultCredentials()` and `forStream(InputStream credentialsStream)`

davidxia:
```
Require upper bound dependencies error for ch.qos.logback:logback-classic:1.1.11 paths to dependency are:
+-com.spotify:helios-services:0.9.0-SNAPSHOT
  +-ch.qos.logback:logback-classic:1.1.11
and
+-com.spotify:helios-services:0.9.0-SNAPSHOT
  +-com.spotify:helios-tools:0.9.0-SNAPSHOT
    +-ch.qos.logback:logback-classic:1.1.11 (managed) <-- ch.qos.logback:logback-classic:1.1.2
and
+-com.spotify:helios-services:0.9.0-SNAPSHOT
  +-net.kencochrane.raven:raven-logback:4.1.1
    +-ch.qos.logback:logback-classic:1.1.11 (managed) <-- ch.qos.logback:logback-classic:1.0.13
and
+-com.spotify:helios-services:0.9.0-SNAPSHOT
  +-com.spotify:docker-client:8.6.0-SNAPSHOT
    +-ch.qos.logback:logback-classic:1.1.11 (managed) <-- ch.qos.logback:logback-classic:1.2.1
and
+-com.spotify:helios-services:0.9.0-SNAPSHOT
  +-com.spotify:logging:2.0.17
    +-ch.qos.logback:logback-classic:1.1.11 (managed) <-- ch.qos.logback:logback-classic:1.0.13
and
+-com.spotify:helios-services:0.9.0-SNAPSHOT
  +-io.dropwizard:dropwizard-core:0.7.1
    +-io.dropwizard:dropwizard-jackson:0.7.1
      +-ch.qos.logback:logback-classic:1.1.11 (managed) <-- ch.qos.logback:logback-classic:1.1.2
and
+-com.spotify:helios-services:0.9.0-SNAPSHOT
  +-io.dropwizard:dropwizard-core:0.7.1
    +-io.dropwizard:dropwizard-logging:0.7.1
      +-ch.qos.logback:logback-classic:1.1.11 (managed) <-- ch.qos.logback:logback-classic:1.1.2
```

mattnworb:
@gimaker I decided to include the refactoring you suggested, please take another look.

Now that docker-client 8.6.0 is released, I think this PR is all-good for review.

mattnworb:
upgrading jackson seems to cause [`AddExtraHostContainerDecoratorTest.simpleTest()`](https://github.com/spotify/helios/blob/013d1b614e0a7c15bfe11418b34a93af5306b6a2/helios-services/src/test/java/com/spotify/helios/agent/AddExtraHostContainerDecoratorTest.java#L37) to fail 😖

helios-services/src/main/java/com/spotify/helios/agent/AgentConfig.java
```diff
@@ -20,9 +20,11 @@
 
 package com.spotify.helios.agent;
 
+import com.google.auth.oauth2.GoogleCredentials;
 import com.spotify.docker.client.DockerHost;
 import com.spotify.helios.servicescommon.CommonConfiguration;
 import com.spotify.helios.servicescommon.FastForwardConfig;
+import java.io.File;
```
gimaker: unused import? :)

PR #1121, opened 2017-05-24
remove extra repositories from pom.xml

mattnworb:
I don't think any of these are needed - all of the libraries used are in
Maven Central.


mattnworb:
I guess they are needed after all

PR #1122, opened 2017-05-25
Add docs on rolling-update --overlap

davidxia:


davidxia:
cc @caipre @mavenraven

davidxia:
Yea it's a small link. I'll do that in another PR.

PR #1126, opened 2017-06-30
Add support to helios for creating tmpfs mounts

gimaker:
Allows users to specify tmpfs partitions in their jobs that are then
created and mounted when the container is run. For example:

```
"ramdisks": {
  # "<mount-point-in-container>": "<mount-options">
  "/run": "rw,noexec,nosuid,size=65536k",
}
```

In the above example `/run` in the container would be mounted to a
tmpfs partition.

This is, for example, useful to put `/tmp` on a tmpfs. One reason
why you might want to do this for Java apps is because the HotSpot JVM
writes a hsperfdata file in `/tmp` synchrously during GC -- which
can lead to issues where GC takes a long time because it's blocked on IO.

### DIFF
--- a/helios-system-tests/src/main/java/com/spotify/helios/system/RamdiskTest.java
+++ b/helios-system-tests/src/main/java/com/spotify/helios/system/RamdiskTest.java
@@ -1,0 +1,138 @@
+/*-
+ * -\-\-
+ * Helios System Tests
+ * --
+ * Copyright (C) 2017 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.helios.system;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.spotify.helios.common.descriptors.Goal.START;
+import static com.spotify.helios.common.descriptors.HostStatus.Status.UP;
+import static com.spotify.helios.common.descriptors.TaskStatus.State.RUNNING;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.Arrays.asList;
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.junit.Assert.assertEquals;
+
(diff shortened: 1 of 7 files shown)


gimaker:
@davidxia @mattnworb @mavenraven

gimaker:
Will add a system test as well

helios-system-tests/src/main/java/com/spotify/helios/system/RamdiskTest.java
```diff
@@ -0,0 +1,138 @@
...
+import com.spotify.helios.common.protocol.CreateJobResponse;
+import com.spotify.helios.common.protocol.JobDeployResponse;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.Socket;
+import java.util.concurrent.Callable;
+import org.junit.Before;
+import org.junit.Test;
+
+public class RamdiskTest extends SystemTestBase {
```
gimaker: This is all copy-pasta from `VolumeTest`

Ideas for better ways to test this appreciated

gimaker:
Looks like tmpfs (a la `docker --tmpfs=...`) isn't supported in CircleCI. https://discuss.circleci.com/t/docker-tmpfs-support/10416/2

helios-system-tests/src/main/java/com/spotify/helios/system/RamdiskTest.java
```diff
@@ -0,0 +1,138 @@
...
+    // Wait for the job to run
+    final TaskStatus taskStatus = awaitJobState(
+        client, testHost(), jobId, RUNNING, LONG_WAIT_SECONDS, SECONDS);
+    assertJobEquals(job, taskStatus.getJob());
+
+    final Integer dfPort = taskStatus.getPorts().get("df").getExternalPort();
+
+    assert dfPort != null;
+
+    // Read "foo" from /volume/bar
```
davidxia: Outdated comment?

mattnworb:
This functionality is very nice to have 👍

gimaker:
> What happens if someone specifies a volume under both ramdisks and volumes, which has precedence? If it would yield a docker error, it might be nice if the JobValidator could validate that a volume isn't specified in both

In practice the `tmpfs` mount takes precedence - if this is intended, stable behaviour or I don't know. I submit a PR to check for conflicts in `JobValidator`.

> Is it too late to change the name of the field in the job? :grinning: since the docker HostConfig message calls it tmpfs I think it would be nice if the same name was used in the Helios job too.

Technically, no (we are not using it yet). If you really want to change it please do so quickly so we can start using it.

I picked `ramdisks` since that seemed more consistent with `volumes` - `tmpfses` just don't have a nice ring to it ;) Also helios isn't consistent with the docker-engine API today -- `volumes` in helios is transformed into a mix of `volumes` and `binds` (in `HostConfig`) in Docker.

> btw is there ever a use case where /tmp should be anything but a tmpfs?

:man_shrugging: I'm sure there is. E.g. if you have very limited amount of RAM you might not want to waste it on `/tmp`, for example.

PR #1129, opened 2017-07-13
Handle docker-client issue with exec healthcheck

mishako:
Exec healthcheck doesn't function properly in circumstances that lead to
https://github.com/spotify/docker-client/issues/513. However the code that triggers the exception is
not crucial, so the healthcheck doesn't need to be interrupted.

See https://github.com/spotify/docker-client/issues/513

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/agent/HealthCheckerFactory.java
+++ b/helios-services/src/main/java/com/spotify/helios/agent/HealthCheckerFactory.java
@@ -102,6 +102,8 @@ public final class HealthCheckerFactory {
         String output = "";
         try (LogStream stream = docker.execStart(execId)) {
           output = stream.readFully();
+        } catch (RuntimeException e) {
+          handleExecStartConnectionReset(e);
         }
 
         final int exitCode = docker.execInspect(execId).exitCode();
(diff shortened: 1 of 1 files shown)


helios-services/src/main/java/com/spotify/helios/agent/HealthCheckerFactory.java
```diff
@@ -144,6 +146,19 @@ private static boolean compatibleDockerVersion(final DockerClient docker) {
       final int minor = Integer.parseInt(Iterables.get(split, 1, "0"));
       return major == 1 && minor >= 18;
     }
+
+    /**
+     * Gracefully handle a nonfatal issue: https://github.com/spotify/docker-client/issues/513
+     */
+    private static void handleExecStartConnectionReset(RuntimeException exception) {
+      Throwable rootCause = Throwables.getRootCause(exception.getCause());
```
mattnworb: just curious, why `getRootCause(exception.getCause())` instead of `getRootCause(exception)`?

mishako:
Fixed the `getCause` thing! Also flipped message comparison to avoid possible null pointer exception. Not sure why the build fails.

mattnworb:
build seems to fail due to some flaky system tests

PR #1131, opened 2017-07-21
HeliosSoloDeployment: support Google Container Registry credentials

mattnworb:
HeliosSoloDeployment: support Google Container Registry credentials

If the configuration profile used by helios-solo contains an entry for `google-container-registry.credentials`, then HeliosSoloDeployment will launch the helios-solo container with additional configuration:

- add an environment variable for `HELIOS_AGENT_OPTS` which contains the flag `--docker-gcp-account-credentials=<file>`
- add a volume bind for the directory containing the credentials file so that helios agent can find the credentials pointed to above

This will allow helios-agent inside the helios-solo container to be able to fetch images from Google Container Registry for any tests that need to do so.

If the value is not set in the configuration file for helios-testing/helios-solo (`helios.conf` or `helios-base.conf`), or to override the value, the system property `-Dhelios.testing.profiles.local.google-container-registry.credentials` can be set to the file path.

If the configuration entry points to a file that doesn't exist then it is ignored.

This allows the user to configure helios-testing/HeliosSoloDeployment to be flexible about what credentials are used inside the helios-solo container, by making it just a matter of configuration to point to a file containing the credentials.

### DIFF
--- a/helios-testing/src/main/java/com/spotify/helios/testing/HeliosConfig.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/HeliosConfig.java
@@ -35,19 +35,18 @@ public class HeliosConfig {
   public static final String APP_CONFIG_FILE = "helios.conf";
 
   /**
-   * @param component The name of the component we're loading config for
-   *                  (i.e. Temp jobs, Helios solo)
    * @return The root configuration loaded from the helios configuration files.
    */
-  static Config loadConfig(final String component) {
-    final ConfigResolveOptions resolveOptions =
-        ConfigResolveOptions.defaults().setAllowUnresolved(true);
+  static Config loadConfig() {
+    final ConfigResolveOptions resolveOptions = ConfigResolveOptions
+        .defaults()
+        .setAllowUnresolved(true);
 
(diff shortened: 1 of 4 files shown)


mattnworb:
@davidxia @caipre @mavenraven

mattnworb:
Added something so that this works for human users too, who have credentials stored in the `gcloud` CLI.

helios-testing/src/main/java/com/spotify/helios/testing/HeliosConfig.java
```diff
@@ -35,19 +35,18 @@
   public static final String APP_CONFIG_FILE = "helios.conf";
 
   /**
-   * @param component The name of the component we're loading config for
-   *                  (i.e. Temp jobs, Helios solo)
    * @return The root configuration loaded from the helios configuration files.
    */
-  static Config loadConfig(final String component) {
```
davidxia: I wonder why we ever had this param.

mattnworb: it used to be used in debug logging as a marker

PR #1132, opened 2017-08-01
Run code reformatting tool and optimize imports

caipre:
This uses a modified configuration of [Spotify checkstyle][0]. All submodules
passed `mvn checkstyle:check` locally. Most of these are whitespace changes, so
it may be useful to add `?w=1` when viewing the diff on GitHub.

[0]: https://github.com/caipre/spotify-checkstyle-config/commit/1b3411965272cdb42ffc092c0579727bf337b514


davidxia:
Why not use the upstream checkstyle or try to get those modifications merged into upstream?

caipre:
@davidxia This still uses upstream checkstyle. The changes in my fork are mostly directives instructing IntelliJ what to do when formatting; checkstyle is still the measure for correctness.

I can submit a PR there as well.

PR #1133, opened 2017-08-01
Initial support for authorization with access tokens

caipre:
This PR adds an `accessTokenCommand` key to the `$HOME/.helios/config` file
that may be executed to provide an access token for use with authentication and
authorization. Commands that act mutatively (eg, `helios deploy`) have been
marked as needing authorization; this is not yet enforced.

Needs more testing, but submitting this initial PR to check the approach.

cc @mattnworb @davidxia @mavenraven


caipre:
Closing to try a different approach.

PR #1135, opened 2017-08-07
client: Support authentication using access tokens

caipre:
This adds support for acquiring an access token within the HeliosClient
Builder via Google Application Default Credentials. When available, the
client will prefer to send an access token in requests as the means for
authentication. As we don't intend for Google credentials to be a hard
requirement, the client will fall back onto the existing authentication
mechanism of self-signed certs and mutual TLS if tokens are unavailable.

An alternative approach to https://github.com/spotify/helios/pull/1133. This approach foregoes pretense of making this a generic access token and enables use of the token by default. I can add a flag to disable use of the token if we feel that's obligated.

cc @mattnworb @mavenraven @davidxia

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/client/AuthenticatingHttpConnector.java
+++ b/helios-client/src/main/java/com/spotify/helios/client/AuthenticatingHttpConnector.java
@@ -22,7 +22,9 @@ package com.spotify.helios.client;
 
 import static java.net.HttpURLConnection.HTTP_FORBIDDEN;
 import static java.net.HttpURLConnection.HTTP_UNAUTHORIZED;
+import static java.util.Collections.singletonList;
 
+import com.google.auth.oauth2.AccessToken;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Optional;
 import com.google.common.collect.ImmutableList;
@@ -45,6 +47,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Queue;
 import javax.security.auth.x500.X500Principal;
+import javax.ws.rs.HEAD;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -57,6 +60,7 @@ public class AuthenticatingHttpConnector implements HttpConnector {
   private static final Logger log = LoggerFactory.getLogger(AuthenticatingHttpConnector.class);
 
   private final String user;
+  private final Optional<AccessToken> accessTokenOpt;
   private final Optional<AgentProxy> agentProxy;
   private final Optional<CertKeyPaths> clientCertificatePath;
   private final List<Identity> identities;
@@ -65,22 +69,25 @@ public class AuthenticatingHttpConnector implements HttpConnector {
   private final DefaultHttpConnector delegate;
 
   public AuthenticatingHttpConnector(final String user,
+                                     final Optional<AccessToken> accessTokenOpt,
                                      final Optional<AgentProxy> agentProxyOpt,
                                      final Optional<CertKeyPaths> clientCertificatePath,
                                      final EndpointIterator endpointIterator,
                                      final DefaultHttpConnector delegate) {
-    this(user, agentProxyOpt, clientCertificatePath, endpointIterator,
+    this(user, accessTokenOpt, agentProxyOpt, clientCertificatePath, endpointIterator,
         delegate, getSshIdentities(agentProxyOpt));
   }
 
   @VisibleForTesting
   AuthenticatingHttpConnector(final String user,
+                              final Optional<AccessToken> accessTokenOpt,
                               final Optional<AgentProxy> agentProxyOpt,
                               final Optional<CertKeyPaths> clientCertificatePath,
                               final EndpointIterator endpointIterator,
                               final DefaultHttpConnector delegate,
                               final List<Identity> identities) {
     this.user = user;
+    this.accessTokenOpt = accessTokenOpt;
     this.agentProxy = agentProxyOpt;
     this.clientCertificatePath = clientCertificatePath;
     this.endpointIterator = endpointIterator;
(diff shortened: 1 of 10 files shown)


helios-client/src/main/java/com/spotify/helios/client/AuthenticatingHttpConnector.java
```diff
@@ -105,8 +111,11 @@ public HttpURLConnection connect(final URI uri, final String method, final byte[
     try {
       log.debug("connecting to {}", ipUri);
 
-      if (clientCertificatePath.isPresent()) {
-        // prioritize using the certificate file if set
+      if (accessTokenOpt.isPresent()) {
+        // prioritize using bearer token
+        return connectWithBearerToken(ipUri, method, entity, headers);
+      } else if (clientCertificatePath.isPresent()) {
```
mattnworb: Why not send the header along with the certificate if both are present, rather than making this an either-or? Seems like they could be separate concerns as one happens at the SSL layer and the other at the HTTP request layer.

caipre: Yeah, they're not mutually exclusive at all. I'm not sure there's much value in doing both though. So long as the master is https, the client cert is only used for authentication, and examining the token seems like a better method.

mattnworb: I think one benefit of sending both would be that we could merge and release this ahead of any ability in the backend/server-side to handle these tokens, to decouple the two.

caipre:
> Not sure if you considered this already, but another way to implement the changes in AuthenticatingHttpConnector could be to make a new HttpConnector implementation that adds the header, since we have a pattern in the existing classes of each HttpConnector wrapping another delegate to add more logic on top of it. This might simplify the implementation/testing a bit.

I'd not really considered this. This implementation already feels pretty simple to me; do you feel strongly about this?

> Also - do you think there would be a case for someone wanting to use credentials in a file on disk that isn't registered in gcloud, i.e. something like --google-application-credentials FILE? I could see a need for that where there is a need to run some helios CLI commands on a host without wanting to setup gcloud. Or do you think that could be solved with the existing flag/logic and just setting GOOGLE_APPLICATION_CREDENTIALS?

Yeah, using that env var would work, though potentially we want to use a CLI flag for the same reasons as we created `DOCKER_GOOGLE_CREDENTIALS`. I can look into adding support for that.

mattnworb:
> I'd not really considered this. This implementation already feels pretty simple to me; do you feel strongly about this?

No 😄 

> Yeah, using that env var would work, though potentially we want to use a CLI flag for the same reasons as we created DOCKER_GOOGLE_CREDENTIALS. I can look into adding support for that.

on the other hand though, since someone invoking `helios <command>` clearly wants to just run that command, I think it is not too unexpected that `GOOGLE_APPLICATION_CREDENTIALS=/foo helios <command>` would use those credentials - there is nothing else being done in the command that could possibly be using those credentials for an unintended use. The motivation for supporting DOCKER_GOOGLE_CREDENTIALS was that the scope for an environment variable there is the entire Maven run/process, and other code in there (such as unit or integration tests) might be accessing Google resources with different expectations about what credentials are used.

helios-client/src/main/java/com/spotify/helios/client/AuthenticatingHttpConnector.java
caipre: Fair point. I'll make the change.

helios-client/src/test/java/com/spotify/helios/client/AuthenticatingHttpConnectorTest.java
```diff
@@ -139,6 +140,20 @@ protected boolean matchesSafely(final URI item) {
     };
   }
 
+  private CustomTypeSafeMatcher<Map<String, List<String>>> hasKeys(final List<String> keys) {
```
mattnworb: for maximum flexibility, this could take a `Matcher<Iterable<String>>` (or some other collection type) which would allow you to do `argThat(hasKeys(contains("Authorization"))` at the call site, to re-use the existing matcher. If hamcrest had a matcher that let you match on a `map.keys()` then you wouldn't need this at all even

helios-tools/src/main/java/com/spotify/helios/cli/CliParser.java
```diff
@@ -297,10 +299,14 @@ public boolean getJson() {
           .setDefault(System.getProperty("user.name"))
           .help("username");
 
-      googleCredentialsArg = addArgument("--google-application-default-credentials")
+      useGoogleCredentialsArg = addArgument("--use-google-credentials")
           .type(Boolean.class)
           .setDefault(true)
-          .help(SUPPRESS);
+          .help("enable Google Credentials to supply an access token");
```
mattnworb: would be helpful to mention what happens if `--google-credentials FILE` isn't passed

caipre:
> Btw, if --use-google-credentials defaults to true, is there some way for the user to specify they want to disable that behavior?

Yes, with `--use-google-credentials=false`. See https://github.com/spotify/helios/pull/1135/files#diff-c30060fe97ea34d4b7c37fd603cdc1c6R210

helios-tools/src/main/java/com/spotify/helios/cli/CliParser.java
caipre: Done.

helios-client/src/test/java/com/spotify/helios/client/AuthenticatingHttpConnectorTest.java
caipre: I'm going to skip this one; happy to learn more about it, but feels unnecessary for the moment.

caipre:
🎉

PR #1138, opened 2017-08-08
upgrade to docker-client 8.8.4

mattnworb:
This version of docker-client:shaded adds relocation of guava classes


mattnworb:
won't ever work because of https://github.com/spotify/docker-client/pull/862

PR #1144, opened 2017-08-18
upgrade to jackson 2.9.0, change Json property inclusion

### LINKED ISSUE #1143: "Id hash mismatch" if helios-client/helios-testing uses Jackson 2.9.x while helios-master/helios-solo uses Jackson 2.8.x
The logic in Job.Builder that creates the "hash" component of the JobId [does the following](https://github.com/spotify/helios/blob/29106a7e5f290a37152ef0e28b21758ed5d0848f/helios-client/src/main/java/com/spotify/helios/common/Json.java#L344-L352):

- serialize the non-metadata parameters of the Job into a JSON string
- deserialize this string back into a Map<String, Object>
- serialize this Map into JSON bytes using `objectWriter.writeValueAsBytes()`. The objectWriter instance has been configured with `setSerializationInclusion(JsonInclude.Include.NON_EMPTY)`.

When the Create Job API method is called, the helios-master re-calculates the hash for the submitted Job on the server-side and verifies that it is equal to the hash submitted by the client. If the two values do not match, a CreateJobResponse is sent back with a status of `INVALID_JOB_DEFINITION` and an errors list containing `Id hash mismatch: hash1 != hash2`.

The same logic is performed by helios-master when run standalone and also when run inside of helios-solo (as a docker image) in integration tests via helios-testing.

A Job or TemporaryJob (with helios-testing) with a `registration` section like

```java
temporaryJobs.job()
  .port("http", 8080)
  .registration("foobar", "tcp", "http")
```

is represented in JSON like `"registration": {"foobar/tcp":{"ports":{"http":{}}}}`.

With Jackson 2.8.x, which is currently used in helios (-client, -master, -testing, -solo), this portion of the JSON is preserved in each step of the JobId hash calculation mentioned below.

With Jackson 2.9.x, the last step of the JobID hash calculation (where the Map is serialized into JSON bytes) omits the entire `"registration"` field, likely because it determines that the inner object of `{"http":{}}` is empty, which then (theoretically) propagates outward. See https://github.com/FasterXML/jackson-databind/issues/1444 and https://github.com/FasterXML/jackson-databind/issues/1736.

Since the JobID hash calculation relies on Jackson when calculating the hash, and creating a Job requires the client and the server to calcualte the same JobID hash, there is a need for both the client and the server to both be using a Jackson version <= 2.8.x or both be using >= 2.9.x. The same applies for tests that use helios-testing/TemporaryJobs (the client) and helios-solo (the server).

If a helios-testing user upgrades their Maven project to use Jackson 2.9.x, then their tests that attempt to deploy TemporaryJobs will fail with 

> CreateJobResponse{status=INVALID_JOB_DEFINITION, errors=[Id hash mismatch: ... ]

because the helios-master inside the helios-solo docker image is using Jackson 2.8.8.

If we upgrade helios to use Jackson 2.9.x, then any users with existing helios-testing tests where the project uses Jackson version <= 2.8.x will download `spotify/helios-solo:latest` the next time they run their tests and begin to get `CreateJobResponse{status=INVALID_JOB_DEFINITION...}` since the `latest` docker image is pulled on each run. Additionally, when a helios-master is deployed using Jackson 2.9.x, any existing helios-cli users of previous versions (which Jackson 2.8.x bundled inside) will be submitting Jobs with an ID hash that the server calculates a different value for.

Arguably the Jackson behavior in 2.8.x with this type of JSON and `setSerializationInclusion(JsonInclude.Include.NON_EMPTY)` (changed in https://github.com/FasterXML/jackson-databind/issues/1444) was incorrect all along, but Helios's insistence that the client and the server calculate the same hash (while using third party libraries which could change serialization behavior with any version upgrade) has created a situation where we need a helios-testing user to be aware and locked into the Jackson version used in the master branch of Helios, and where we need helios-cli/helios-client users to be using a version of Jackson that behaves the same way as the version used inside helios-master.

I'm not fully sure yet but https://github.com/FasterXML/jackson-databind/issues/1444 makes it sound like it might be possible to use `setDefaultPropertyInclusion()` (introduced in 2.9.x I think?) to get the same serialization behavior with these empty objects as we had in 2.8.x.

The larger issue here is that the JobId hash - which is of dubious value - relies on a third-party library for which a different version is frequently used in projects that use helios-testing.

mattnworb:
~~~merging this will break the tests of any helios-testing users, who will end up downloading a new docker image of `spotify/helios-solo:latest` on their next test run. All test users will have to upgrade their helios-testing/helios-client dependency in sync with the merge of this PR.~~~

Fixes #1143.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/Json.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/Json.java
@@ -51,7 +51,10 @@ public class Json {
       .configure(FAIL_ON_UNKNOWN_PROPERTIES, false);
 
   private static final ObjectWriter NORMALIZING_OBJECT_WRITER = new ObjectMapper()
-      .setSerializationInclusion(JsonInclude.Include.NON_EMPTY)
+      .setDefaultPropertyInclusion(
+          //parameters are valueIncl=NON_EMPTY, contentIncl=ALWAYS
+          JsonInclude.Value.construct(JsonInclude.Include.NON_EMPTY, JsonInclude.Include.ALWAYS)
+      )
       .configure(SORT_PROPERTIES_ALPHABETICALLY, true)
       .configure(ORDER_MAP_ENTRIES_BY_KEYS, true)
       .configure(WRITE_DATES_AS_TIMESTAMPS, false)
--- a/pom.xml
+++ b/pom.xml
@@ -157,32 +157,32 @@
             <dependency>
                 <groupId>com.fasterxml.jackson.dataformat</groupId>
                 <artifactId>jackson-dataformat-yaml</artifactId>
-                <version>2.8.8</version>
+                <version>2.9.0</version>
             </dependency>
             <dependency>
                 <groupId>com.fasterxml.jackson.datatype</groupId>
                 <artifactId>jackson-datatype-guava</artifactId>
-                <version>2.8.8</version>
+                <version>2.9.0</version>
             </dependency>
             <dependency>
                 <groupId>com.fasterxml.jackson.core</groupId>
                 <artifactId>jackson-core</artifactId>
-                <version>2.8.8</version>
+                <version>2.9.0</version>
             </dependency>
             <dependency>
                 <groupId>com.fasterxml.jackson.core</groupId>
                 <artifactId>jackson-annotations</artifactId>
-                <version>2.8.8</version>
+                <version>2.9.0</version>
             </dependency>
             <dependency>
                 <groupId>com.fasterxml.jackson.core</groupId>
                 <artifactId>jackson-databind</artifactId>
-                <version>2.8.8</version>
+                <version>2.9.0</version>
             </dependency>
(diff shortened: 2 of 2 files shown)


mattnworb:
> merging this will break the tests of any helios-testing users, who will end up downloading a new docker image of spotify/helios-solo:latest on their next test run. All test users will have to upgrade their helios-testing/helios-client dependency in sync with the merge of this PR.

Thinking about this some more, this is not accurate. Since this change makes helios use jackson 2.9 and maintains the prior behavior we used to have with 2.8, it should be safe to merge at any time, as helios-testing users using jackson 2.8 should not see any change at all. I've edited this out of my original PR message

caipre:
Makes sense to me. 👍

mrm1ck:
https://github.com/spotify/helios/pull/1144.patch
https://github.com/spotify/helios/pull/1144.diff
This looks better

PR #1146, opened 2017-08-24
remove overly noisy log line

mattnworb:


mattnworb:
I rather just remove it, since we never run the masters with debug logging (too much noise), and if troubleshooting something locally in the future someone can just re-add it

PR #1151, opened 2017-09-08
Add shmSize to resources configuration

mspangdal:
Not sure if I missed something but I cant find a way to set this for our containers. So here is a proposal for adding it

Resources conf could be extended to support CLI too but that can be done in a seperate PR


mspangdal:
The one test that failed seems like its not related to the change, perhaps rerunning tests will make it 💚 ?

>java.util.concurrent.TimeoutException: The callable com.spotify.helios.system.SystemTestBase$14 defined in com.spotify.helios.system.SystemTestBase.awaitTaskState(SystemTestBase.java:951) failed to return anything after 400000 ms

PR #1156, opened 2017-09-28
Fix DG event topic name

davidxia:


helios-services/src/main/java/com/spotify/helios/master/MasterService.java
```diff
@@ -187,14 +188,12 @@ public MasterService(final MasterConfig config,
       }
     }
 
-    final String deploymentGroupEventTopic = TaskStatusEvent.TASK_STATUS_EVENT_TOPIC;
-
     final List<EventSender> eventSenders =
-        EventSenderFactory.build(environment, config, metricsRegistry, deploymentGroupEventTopic);
+        EventSenderFactory.build(environment, config, metricsRegistry, DG_EVENT_TOPIC);
```
mattnworb: did you mean to change this value too, used for the healthchecks IIRC?

caipre: If I'm [reading this correctly][0], it shouldn't matter. As you say, the topic is used to test the connection and authorization. Both the `HeliosTaskStatusEvents` and `HeliosDeploymentGroupEvents` topics are meant to exist, it's just that we haven't been publishing messages to the latter.

[0]: https://github.com/spotify/helios/blob/master/helios-services/src/main/java/com/spotify/helios/servicescommon/EventSenderFactory.java#L71-L74

PR #1157, opened 2017-09-29
Add log message for rolling-update failures

caipre:
Presently, helios-master will emit Kafka and Google Cloud PubSub events
when a rolling-update fails (in addition to storing the failure and
error message in ZooKeeper). This change adds a log message for the
failure as well, which should be useful for systems without a defined
event handler or for quick-and-dirty integrations that don't need the
larger feature set of something like Kafka or Google Cloud PubSub.

@davidxia @mattnworb @mavenraven

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdateOpFactory.java
+++ b/helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdateOpFactory.java
@@ -195,6 +195,8 @@ public class RollingUpdateOpFactory {
         deploymentGroup, task, errMsg, errorCode, metadata);
     events.add(taskEv);
     events.add(eventFactory.rollingUpdateFailed(deploymentGroup, taskEv));
+    log.info("rolling-update step on deployment-group name={} failed; message={}",
+        deploymentGroup.getName(), errMsg);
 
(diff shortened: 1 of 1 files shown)


helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdateOpFactory.java
```diff
@@ -195,6 +195,8 @@ public RollingUpdateOp error(final String msg, final String host,
         deploymentGroup, task, errMsg, errorCode, metadata);
     events.add(taskEv);
     events.add(eventFactory.rollingUpdateFailed(deploymentGroup, taskEv));
+    log.info("rolling-update step on deployment-group name={} failed; message={}",
```
mattnworb: might be nice if this could mention the job too - something to identify _which_ rolling-update on the group it was - but seems like that might not be available in this method

caipre: At least for the cases I've looked at, `errMsg` includes that information.

https://github.com/spotify/helios/blob/master/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java#L846

PR #1160, opened 2017-10-11
Fix bug: use scoped GoogleCredentials

davidxia:


davidxia:
I tried to write a test for this, but couldn't think of a straight forward way.

helios-client/src/main/java/com/spotify/helios/client/GoogleCredentialsAccessTokenProvider.java
```diff
@@ -74,12 +74,11 @@ static AccessToken getAccessToken(List<String> scopes) throws IOException {
       return null;
     }
 
-    if (!scopes.isEmpty()) {
-      credentials.createScoped(scopes);
-    }
-    credentials.refresh();
+    final GoogleCredentials newCredentials =
```
caipre: Minor: `credentials` isn't `final` above, so you could just replace it inline.

```
if (!scopes.isEmpty()) {
    credentials = credentials.createScoped(scopes);
}
```

mattnworb: +1 to this

PR #1161, opened 2017-10-12
Adds docs for how parallelism works.

mavenraven:
@davidxia @mattnworb @caipre

### DIFF
--- a/docs/parallelism.md
+++ b/docs/parallelism.md
@@ -1,0 +1,59 @@
+# Paralellism
+
+`--parallelism` is a flag available on `rolling-update`.  This flag is better thought of as "chunking" a deployment, instead
+of true paralellism.
+
+Lets say we have hosts **A**, **B**, and **C**, each in the running state:
+
+| A | B | C |
+|---|---|---|
+| R | R | R |
+
+We issue a `rolling-update` with `--parallelism` set to 2. The first thing that happens is that **A** and **B** undeploy their currently running job, entering the undeployed state. Note that undeploying is an asynchronous operation for the Helios master as well as the Helios agents on hosts **A** and **B**. The host **C** is unaffected:
+
+| A | B | C |
+|---|---|---|
+| U | U | R |
+
+The next thing that happens is that **A** and **B** are put into the deploy state [in that order]. Again, this operation is asynchronous. Host **C** is still  unaffected:
+
+| A | B | C |
+|---|---|---|
+| D | D | R |
+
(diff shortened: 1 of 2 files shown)


docs/parallelism.md
```diff
@@ -0,0 +1,59 @@
+# Paralellism
```
mattnworb: nit: this is misspelt

docs/parallelism.md
```diff
@@ -0,0 +1,59 @@
...
+
+| A | B | C |
+|---|---|---|
+| W | W | R |
+
+Now that **A** is running again, the rolling update will continue with waiting for **B** to complete deploying successfully. Since **B** has been deploying asynchrously while the rolling update waited for **A** to finish, the time to wait is typically very short. This is the mechanism that makes `--parallelism` decrease the time it takes for a deployment to finish.
+
+| A | B | C |
+|---|---|---|
+| R | W | R |
```
mattnworb: maybe this diagram would show more data if you labeled it as "R2" or "R1" for which version is running?

docs/parallelism.md
```diff
@@ -0,0 +1,59 @@
+# Paralellism
+
+`--parallelism` is a flag available on `rolling-update`.  This flag is better thought of as "chunking" a deployment, instead
+of true paralellism.
+
+Lets say we have hosts **A**, **B**, and **C**, each in the running state:
```
davidxia: nit: Let's since you mean let us.

docs/parallelism.md
```diff
@@ -0,0 +1,59 @@
+# Paralellism
+
+`--parallelism` is a flag available on `rolling-update`.  This flag is better thought of as "chunking" a deployment, instead
```
davidxia: don't need the comma at the end.

davidxia: I prefer the word partition instead of chunk :) https://en.wikipedia.org/wiki/Partition_of_a_set

bernhaker:
Good nob

caipre:
LGTM.

docs/parallelism.md
mavenraven: Great idea, thanks!

docs/parallelism.md
mavenraven: thanks, fixed

docs/parallelism.md
mavenraven: fixed

docs/parallelism.md
mavenraven: 👍

PR #1162, opened 2017-10-14
bump jackson dep to 2.9.2

danielnorberg:
Jackson 2.9.2 has a fix for an issue that affects automatter users, of which several are also helios-testing users.

https://github.com/FasterXML/jackson-databind/issues/1756

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -157,32 +157,32 @@
             <dependency>
                 <groupId>com.fasterxml.jackson.dataformat</groupId>
                 <artifactId>jackson-dataformat-yaml</artifactId>
-                <version>2.9.0</version>
+                <version>${jackson.version}</version>
             </dependency>
             <dependency>
                 <groupId>com.fasterxml.jackson.datatype</groupId>
                 <artifactId>jackson-datatype-guava</artifactId>
-                <version>2.9.0</version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -157,32 +157,32 @@
             <dependency>
                 <groupId>com.fasterxml.jackson.dataformat</groupId>
                 <artifactId>jackson-dataformat-yaml</artifactId>
-                <version>2.9.0</version>
```
davidxia: I think we kept these hardcoded versions because @dflemstr said variable versions don't play well with maven tools that upgrade dependency versions. I don't have strong feelings though. And looking at the [versions-maven-plugin](http://www.mojohaus.org/versions-maven-plugin/examples/update-properties.html), I think it can handle variable versions.

@dflemstr can you explain again why you didn't like this? I'm just curious.

mattnworb: I think this causes problems if something depending on helios as a dependency also specifies a `jackson.version` property. But then again, we should be consistent in this pom and either use properties for repeated versions like this or no use it at all.

davidxia: Looks like we don't really use properties for repeated versions.

davidxia:
We're now on jackson 2.9.6.

PR #1164, opened 2017-10-16
Add note about using hosts for label discovery in deployment groups

robkinyon:
This adds a quick note to use the `helios hosts` command to figure out which labels (and values) apply to a given host.

### DIFF
--- a/docs/deployment_groups.md
+++ b/docs/deployment_groups.md
@@ -31,6 +31,14 @@ Creating a deployment group with the same name and labels as an existing one wil
 "not modified" response. Creating a deployment group with the same name but different labels as an
 existing one will return an error stating that the deployment group already exists.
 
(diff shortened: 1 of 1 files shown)


docs/deployment_groups.md
```diff
@@ -31,6 +31,14 @@ Creating a deployment group with the same name and labels as an existing one wil
 "not modified" response. Creating a deployment group with the same name but different labels as an
 existing one will return an error stating that the deployment group already exists.
 
+To figure out which labels are available and what sort of hosts each label key
+and value would return, you can use
```
davidxia: Can you change to just "you can use `helios hosts`"`? We'll assume the user by this point knows how the `hosts` subcommand works.

davidxia:
@robkinyon How about https://github.com/spotify/helios/pull/1189 instead?

PR #1165, opened 2017-10-16
Add rollout options to job config

davidxia:


davidxia:
@mavenraven 

cc @danielhan

caipre:
@davidxia Are you planning to clean up the commits?

davidxia:
@caipre Fixed the commit message. Also updated `inspect` output.

davidxia:
I'm OK with this approach if we update docs that say job configs with null rollout options will have those options defaulted to some values by the master at rolling-update time not on job creation time.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -624,19 +624,28 @@ public void rollingUpdate(final DeploymentGroup deploymentGroup,
...
     checkNotNull(deploymentGroup, "deploymentGroup");
 
-    log.info("preparing to initiate rolling-update on deployment-group: name={}, jobId={}",
-        deploymentGroup.getName(), jobId);
+    final Job job = getJob(jobId);
+    if (job == null) {
+      throw new JobDoesNotExistException(jobId);
+    }
+
+    final RolloutOptions mergedRolloutOptions =
```
davidxia: As I mentioned before, the downside I see so far with this approach is that jobs can be created with null values for rollout options. They'll exist on disk with null values for these options. When you inspect them, they'll output null values. Then when you start a rolling-update, the master will fill in the defaults. This is different behavior from the other options in the job config. The state they are in when created and on disk is the final configuration at deployment time.

helios-tools/src/main/java/com/spotify/helios/cli/command/JobCreateCommand.java
```diff
@@ -277,6 +279,11 @@ protected JobCreateCommand(final Subparser parser, Supplier<Map<String, String>>
         .help("Labels (key-value pairs) to apply onto the container. "
               + "Defaults to nothing.");
 
+    rolloutOptionsArg = parser.addArgument("--rollout-options")
```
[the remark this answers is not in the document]
davidxia: for consistency with the other args

helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
```diff
@@ -255,6 +268,7 @@ public Job(
     this.labels = Optional.fromNullable(labels).or(EMPTY_LABELS);
     this.secondsToWaitBeforeKill = secondsToWaitBeforeKill;
     this.ramdisks = firstNonNull(ramdisks, EMPTY_RAMDISKS);
+    this.rolloutOptions = Optional.fromNullable(rolloutOptions).orNull();
```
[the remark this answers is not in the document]
davidxia: Do you mean instead of `Optional.fromNullable(rolloutOptions).orNull();` we do `Optional.fromNullable(rolloutOptions).or(EMPTY_ROLLOUT_OPTIONS);?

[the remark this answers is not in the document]
mavenraven: yep, exactly!

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
caipre: On the other hand, if we fill in the null values on job create, the config file used to create the job will differ from the config stored in zk. Though I suppose that's strictly true even when we keep the null values, since our JSON reader allows missing fields while the writer writes all fields.

Field in config / cli|Behavior|Field in Zookeeper / `helios inspect`
-|-|-
Present|Fill with `null`|As given in config / cli
Present|Fill with default|As given in config / cli
Absent|Fill with `null`|Strictly differs but logically equivalent
Absent|Fill with default|Differs from config and may be unexpected

Neither alternative to the handing of absent fields is clearly preferable in my view. I'd say that filling with defaults gives the most information, and can be clear enough if we document the defaults and their application during job creation.

helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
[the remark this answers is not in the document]
mattnworb: I think the use of Optional here is totally unnecessary - could just be `this.rolloutOptions = rolloutOptions`.

helios-services/src/test/java/com/spotify/helios/master/DeploymentGroupTest.java
```diff
@@ -139,7 +139,7 @@ public void testUpdateDeploymentGroupHosts() throws Exception {
         .setName(GROUP_NAME)
         .setHostSelectors(ImmutableList.of(HostSelector.parse("role=melmac")))
         .setJobId(job.getId())
-        .setRolloutOptions(RolloutOptions.newBuilder().build())
+        .setRolloutOptions(RolloutOptions.DEFAULT_ROLLOUT_OPTIONS)
```
caipre: We should add a test where the `RolloutOptions` are partially specified.

helios-client/src/main/java/com/spotify/helios/common/descriptors/RolloutOptions.java
```diff
@@ -79,31 +92,48 @@ public Builder toBuilder() {
...
+        .setOverlap(overlap)
         .setToken(token)
         .setIgnoreFailures(ignoreFailures);
   }
 
-  public long getTimeout() {
+
+  /**
+   * Return a new RolloutOptions instance by merging this instance with another one.
+   * The other instance is assumed to have no null-valued fields.
```
mattnworb: what happens if this assumption is violated?

helios-client/src/main/java/com/spotify/helios/common/descriptors/RolloutOptions.java
```diff
@@ -165,39 +175,31 @@ public String toString() {
...
-    private boolean migrate;
-    private boolean overlap;
+    private Long timeout;
+    private Integer parallelism;
+    private Boolean migrate;
+    private Boolean overlap;
     private String token;
-    private boolean ignoreFailures;
-
-    public Builder() {
```
caipre: I changed the behavior of this `Builder` to not set defaults for fields since I wanted the `rollingUpdateCommand` to be able to send a partially filled out set of options. An alternative approach might be to preserve the "builder gives defaults" pattern here and change [this line](https://github.com/spotify/helios/pull/1165/files#diff-09ebe80c4b7d83ee0a1323aed194d6baR167) to use the raw constructor.

helios-tools/src/main/java/com/spotify/helios/cli/command/RollingUpdateCommand.java
```diff
@@ -114,28 +112,27 @@ public Long get() {
               + "this will NOT abort the rolling update, it will just cause this command to exit.");
 
     migrateArg = parser.addArgument("--migrate")
-        .setDefault(false)
         .action(storeTrue())
+        .setDefault((Object) null)
```
mattnworb: isn't this the default already?

davidxia: `storeTrue()` implicitly sets the default to false

helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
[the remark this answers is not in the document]
davidxia: > I think the use of Optional here is totally unnecessary - could just be this.rolloutOptions = rolloutOptions.

Good point. Why do we have all these `Optional.fromNullable(X).orNull()` anyways?

helios-client/src/main/java/com/spotify/helios/common/descriptors/RolloutOptions.java
davidxia: `firstNonNull()` will throw NPE if both are null. So I think this is OK.

helios-client/src/main/java/com/spotify/helios/common/descriptors/RolloutOptions.java
davidxia: @caipre Your link didn't go to a line. Are you referring to [this line](https://github.com/spotify/helios/pull/1165/files#diff-48145829c6492a8c9e879873420922afR106)?

helios-tools/src/main/java/com/spotify/helios/cli/command/RollingUpdateCommand.java
davidxia: It's ugly. Let me know if anyone has a better way to not default to false.

mattnworb: ah that is pretty weird. I guess the only other option really is to write your own ArgumentAction that doesn't do that, but that feels silly.

helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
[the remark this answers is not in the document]
mattnworb: No idea - perhaps the idea was to use the same pattern as `Optional.fromNullable(x).or(foo)`, but even that is better expressed nowadays as `firstNonNull(x, foo)`.

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
davidxia: > I'd say that filling with defaults gives the most information, and can be clear enough if we document the defaults and their application during job creation.

agreed

helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
```diff
@@ -64,6 +64,14 @@
...
  *   "expires" : null,
  *   "gracePeriod": 60,
+ *   "rolloutOptions": {
+ *     "migrate": false,
+ *     "parallelism": 2,
+ *     "timeout": 1000,
+ *     "overlap": true,
+ *     "token": "insecure-access-token",
+ *     "ignoreFailures": false
+ *   },
```
gimaker: I don't think `migrate` and `token` make sense to allow in a job.

davidxia: `token` is already in the job. `migrate` doesn't make sense, but it seems more work and potentially more confusing to exclude some RolloutOption fields than to include everything.

helios-services/src/test/java/com/spotify/helios/master/DeploymentGroupTest.java
davidxia: If I add a unit test for the fallback logic in ZKMM, do we still need this?

helios-client/src/main/java/com/spotify/helios/common/descriptors/RolloutOptions.java
caipre: Above this line a bit: https://github.com/spotify/helios/pull/1165/files#diff-09ebe80c4b7d83ee0a1323aed194d6baR172

helios-services/src/test/java/com/spotify/helios/master/DeploymentGroupTest.java
caipre: Are you willing to assert that every codepath that constructs a `DeploymentGroup` does so with a `RolloutOptions` that has all fields filled in?

helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
mattnworb: Just to toss the idea out there, you could introduce a new type ("JobRolloutOptions", for instance) with just the relevant fields, and then have the merging code/function be capable of producing a RolloutOptions from this type and the existing values/defaults.

helios-services/src/test/java/com/spotify/helios/master/DeploymentGroupTest.java
davidxia: Do you think enforcing this in the `DeploymentGroup` constructor is cleaner than adding a bunch of null checks everywhere we do `DeploymentGroup.getRolloutOptions().getXXX()`?

davidxia:
Last two commits address 1-3 [here](https://github.com/spotify/helios/pull/1165#pullrequestreview-70228459).

mattnworb:
@caipre just curious, do you have another approach in mind?

docs/user_manual.md
```diff
@@ -226,6 +234,13 @@ If is specified, Helios will unregister from service discovery and wait the
 specified number of seconds before undeploying. Optional, defaults to `0` for
 no grace period.
 
+#### rolloutOptions
+If specified, Helios will use these options when performing a `rolling-update`.
+The order of precedence for these options is as follows.
+1. Values passed in via `rolling-update`'s switches
+2. Values here
```
mattnworb: could this be more specific, and say "Values specified in the job object"? "here" was confusing to me at first

caipre: This is how I worded it in the internal document:

> When a deployment is issued to a helios master, the RolloutOptions will be constructed from the request parameters and job configuration with the following precedence:
>
> 1. Parameters included in the API request (eg, flags passed on the command line)
> 2. Parameters defined in the job configuration (of the job being rolled out)
> 3. The default value for the parameter as defined in the RolloutOptions class

davidxia: Thanks I'll use that.

mattnworb:
thanks for addressing my feedback

PR #1166, opened 2017-10-16
Replace deprecated WHITESPACE with CharMatcher.whitespace()

davidxia:


davidxia:
The tests were indeed flaky. merging

PR #1172, opened 2017-10-26
log more details when rolling-update times out

mattnworb:
Log the elapsed time of a deploy when it has timed out, to help in
debugging problems where rollouts appear to timeout too fast.

Ideally the elapsed time could be included in the error message reported
back to the user, but this seemed like too large of a code change, as
the elapsed time is internal to this `isRolloutTimedOut` function.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
+++ b/helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
@@ -957,18 +957,27 @@ public class ZooKeeperMasterModel implements MasterModel {
 
   private boolean isRolloutTimedOut(final ZooKeeperClient client,
                                     final DeploymentGroup deploymentGroup) {
+    final String groupName = deploymentGroup.getName();
+    final long groupTimeoutSetting = deploymentGroup.getRolloutOptions().getTimeout();
+    final long secondsSinceDeploy;
     try {
-      final String statusPath = Paths.statusDeploymentGroupTasks(deploymentGroup.getName());
-      final long secondsSinceDeploy = MILLISECONDS.toSeconds(
+      final String statusPath = Paths.statusDeploymentGroupTasks(groupName);
+      secondsSinceDeploy = MILLISECONDS.toSeconds(
           System.currentTimeMillis() - client.getNode(statusPath).getStat().getMtime());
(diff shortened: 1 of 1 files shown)


helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -957,18 +957,27 @@ private RollingUpdateOp rollingUpdateAwaitRunning(final ZooKeeperClient client,
...
       // statusPath doesn't exist or some other ZK issue. probably this deployment group
       // was removed.
-      log.warn("error determining deployment group modification time: {} - {}",
-          deploymentGroup.getName(), e);
+      log.warn("error determining deployment group modification time: {}", groupName, e);
       return false;
     }
+
+    if (secondsSinceDeploy > groupTimeoutSetting) {
+      log.info("rolling-update on deployment-group name={} has timed out after "
```
mattnworb: `info` to match the log message that is logged when a rollout fails due to timeout here https://github.com/spotify/helios/blob/f02446657f06c4fa8f7aad9a23bf87d429fb80db/helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdateOpFactory.java#L198-L199

mattnworb:
I restructured the method slightly so that the `>` check is not inside a try-catch for a zookeeper exception.

@davidxia @caipre

helios-services/src/main/java/com/spotify/helios/master/ZooKeeperMasterModel.java
```diff
@@ -957,18 +957,27 @@ private RollingUpdateOp rollingUpdateAwaitRunning(final ZooKeeperClient client,
 
   private boolean isRolloutTimedOut(final ZooKeeperClient client,
                                     final DeploymentGroup deploymentGroup) {
+    final String groupName = deploymentGroup.getName();
+    final long groupTimeoutSetting = deploymentGroup.getRolloutOptions().getTimeout();
```
davidxia: I guess if NPE is here, we'll see it in logs?

mattnworb: yes. this unboxing already existed, but as far as I can tell this value should never actually be null (I think it would have been preferable if the internal representation could be some type without nullable fields).

PR #1175, opened 2017-10-30
Clarify rollout options docs

davidxia:
Provide example and reorder job config to be alphabetical.

### DIFF
--- a/docs/user_manual.md
+++ b/docs/user_manual.md
@@ -147,14 +147,6 @@ example that uses all the available configuration keys with an explanation of ea
   },
   "expires" : "2014-06-01T12:00:00Z",
   "gracePeriod": 60,
-  "rolloutOptions": {
-    "migrate": false,
-    "parallelism": 2,
-    "timeout": 1000,
-    "overlap": true,
-    "token": "insecure-access-token",
-    "ignoreFailures": false
-  },
   "healthCheck" : {
     "type" : "http",
     "path" : "/healthcheck",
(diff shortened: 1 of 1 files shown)


docs/user_manual.md
```diff
@@ -332,6 +325,43 @@ These settings correspond to the ones in the [Docker docs][2].
 
 What is allowed here will vary based upon the discovery service plugin used.
 
+#### rolloutOptions
+When you run `helios rolling-update`, the rolling-update options will be constructed from the
+values you provide on the CLI and the job configuration with the following precedence:
+
+* Parameters included in the API request (eg, flags passed on the command line)
+* Parameters defined in the job configuration (of the job being rolled out)
```
mattnworb: how about adding something like "see above for an example of what this looks like in JSON"?

davidxia: ah forgot that part.

docs/user_manual.md
```diff
@@ -332,6 +325,43 @@ These settings correspond to the ones in the [Docker docs][2].
...
+
+```
+"rolloutOptions": {
+  "parallelism": 2,
+  "overlap": false,
+  "ignoreFailures": false
+}
+```
+
+Will ultimately have these options at `rolling-update` time. Since timeout, migrate, and token
```
mattnworb: For clarity, I'd move the `Since timeout ...` part below the json snippet, otherwise "Will ultimately have these options at rolling-update time." looks like an incomplete sentence

davidxia:
cc @matslina

PR #1176, opened 2017-10-31
Inconsistent values in the example job config

jclopes:
If only the missing fields are updated the other fields should be kept the same as in the example just above.


davidxia:
Closing as I haven't heard back. Please reopen if you think the example is wrong.

PR #1181, opened 2017-11-07
Fix NPEs

davidxia:


helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdateOpFactory.java
```diff
@@ -160,7 +161,7 @@ public RollingUpdateOp yield() {
 
   private boolean isIgnoreFailures() {
     return deploymentGroup.getRolloutOptions() != null
-           && deploymentGroup.getRolloutOptions().getIgnoreFailures();
+           && TRUE.equals(deploymentGroup.getRolloutOptions().getIgnoreFailures());
```
mavenraven: @davidxia this will still throw if getRolloutOptions() returns null, right?

mavenraven:
👍

helios-services/src/main/java/com/spotify/helios/rollingupdate/RollingUpdateOpFactory.java
davidxia: see line above

caipre:
👍

PR #1183, opened 2017-11-08
Revert "Fix NPEs"

mavenraven:
This reverts commit 704dfd597e9d96821a0787d0a7d5db738e9b239e.

@danielnorberg @mattnworb @davidxia @caipre @danielhan

PR #1186, opened 2017-11-08
Adds back reverted commits and fixes NPE in rolling-update cmd

davidxia:


mattnworb:
between this PR, and #1181 #1183 #1184 I am not seeing any explanation of what caused the exceptions originally, so it is really hard to review this at all.

What caused the NullPointerExceptions in 1181?
Why do tests not catch that? Can tests be added that catch that?
Why revert it, and why now revert the revert without additional changes?

davidxia:
> What caused the NullPointerExceptions in 1181?

https://github.com/spotify/helios/pull/1181/files didn't cause any NPEs. In fact, it fixes a bunch of server-side NPEs.

> Why do tests not catch that? Can tests be added that catch that?

Are you talking about catching the NPE fixed by #1181?

> Why revert it, and why now revert the revert without additional changes?

Not sure about the first. For the second question, there are additional changes. I added a new commit.

helios-tools/src/main/java/com/spotify/helios/cli/command/RollingUpdateCommand.java
```diff
@@ -184,27 +186,36 @@ protected int runWithJobId(final Namespace options, final HeliosClient client,
       return 1;
     }
 
+    final RolloutOptions optionsFromJob = job.getRolloutOptions();
```
davidxia: NPE starts here. This variable can be null.

helios-tools/src/main/java/com/spotify/helios/cli/command/RollingUpdateCommand.java
```diff
@@ -21,16 +21,17 @@
 package com.spotify.helios.cli.command;
 
 import static com.google.common.base.Preconditions.checkArgument;
-import static com.spotify.helios.common.descriptors.Job.EMPTY_TOKEN;
 import static java.lang.String.format;
 import static net.sourceforge.argparse4j.impl.Arguments.storeTrue;
 
+import com.google.api.client.repackaged.com.google.common.base.Objects;
```
mattnworb: looks like an accidental import

davidxia: thanks, fixed

caipre:
👍

PR #1187, opened 2017-11-10
Include new docker-ce as possible dependency

martina-if:
The new way to install docker is installing docker community edition.

@davidxia @mattnworb


davidxia:
Sorry, the docker-ce deb package name change is old. We just never updated it here.

davidxia:
@mattnworb OK if we merge this?

PR #1190, opened 2017-11-13
Upgrade docker-client dep from 8.9.0 to 8.9.2

davidxia:
depends on https://github.com/spotify/docker-client/pull/922

### DIFF
--- a/helios-client/pom.xml
+++ b/helios-client/pom.xml
@@ -74,7 +74,6 @@
         <dependency>
             <groupId>com.spotify</groupId>
             <artifactId>ssh-agent-proxy</artifactId>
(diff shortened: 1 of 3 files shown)


davidxia:
Tests now pass.

davidxia:
In general my motivation for upgrades like this is to not fall too far behind upstream. I don't like having to upgrade in cases like this where we don't need any of the new features and upgrading presents a little bit of risk. But my other fear of not tracking upstream is the scenario where we need to fix a bug in docker-client, then if we're really far behind it's scarier to upgrade. Not sure if there's a solution better than incrementally upgrading.

davidxia:
I'll merge next Monday.

PR #1193, opened 2017-11-21
Add setGoogleAccessToken to HeliosClient.Builder

davidxia:
in cases where client cannot use env var HELIOS_GOOGLE_CREDENTIALS.


davidxia:
ah good point. We'll punt and see if anyone ever needs to pass arbitrary values for the Authorization header.

davidxia:
cc @martina-if

martina-if:
@davidxia @mattnworb  I will actually use a Bearer token that I get from an HTTP header. But if I can just create this `AccessToken` object with it I think it should be fine.

davidxia:
Just FYI, the AccessToken needs to already have the required scopes set.

martina-if:
I see in the source code that I can, just passing the string in the constructor. Thanks!!

PR #1195, opened 2017-12-08
fix HeliosSoloDeployment library for Docker For Mac

danielhan:
InetAddress.getLocalHost() returns the loopback device. This does not allow an application that is running inside a container from being able to communicate with the host, since hitting 127.0.0.1 will make it talk to itself. An ipv4 address of the host needs to be given to the container.

@davidxia @mavenraven @mattnworb

### DIFF
--- a/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
@@ -60,10 +60,10 @@ import java.net.Inet4Address;
 import java.net.InetAddress;
 import java.net.NetworkInterface;
 import java.net.SocketException;
+import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -194,12 +194,17 @@ public class HeliosSoloDeployment implements HeliosDeployment {
       if (isDockerForMac(dockerInfo)) {
         log.info("determineHeliosHost: local environment appears to be Docker for Mac");
         try {
-          List<InetAddress> addrs = new ArrayList<InetAddress>();
-          for (NetworkInterface nic : Collections.list(NetworkInterface.getNetworkInterfaces())) {
-            addrs.addAll(Collections.list(nic.getInetAddresses()));
+          InetAddress localhost = InetAddress.getLocalHost();
+          if (localhost.isLoopbackAddress()) {
+            List<InetAddress> addrs = new ArrayList();
+            for (NetworkInterface nic : Collections.list(NetworkInterface.getNetworkInterfaces())) {
+              addrs.addAll(Collections.list(nic.getInetAddresses()));
+            }
+            return findFirstInet4Addr(addrs);
+          } else {
+            return localhost.getHostAddress();
           }
-          return findFirstInet4Addr(addrs);
-        } catch (SocketException e) {
+        } catch (SocketException | UnknownHostException e) {
           throw new HeliosDeploymentException("Cannot resolve local hostname", e);
         }
       }
--- a/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
+++ b/helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
@@ -56,9 +56,14 @@ import com.spotify.helios.common.protocol.JobUndeployResponse;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigValue;
 import java.io.File;
+import java.net.Inet4Address;
 import java.net.InetAddress;
-import java.net.UnknownHostException;
+import java.net.NetworkInterface;
+import java.net.SocketException;
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


helios-testing/src/test/java/com/spotify/helios/testing/HeliosSoloDeploymentTest.java
```diff
@@ -408,4 +412,32 @@ public void testLogService() throws Exception {
...
+    InterfaceAddress ifaceAddr1 = mock(InterfaceAddress.class);
+    InterfaceAddress ifaceAddr2 = mock(InterfaceAddress.class);
+    InterfaceAddress ifaceAddr3 = mock(InterfaceAddress.class);
+    InetAddress inetAddr0 = mock(InetAddress.class);
+    InetAddress inetAddr1 = mock(InetAddress.class);
+    InetAddress inetAddr2 = mock(InetAddress.class);
+    InetAddress inetAddr3 = mock(InetAddress.class);
+    when(inetAddr0.toString()).thenReturn("/fe80:0:0:0:0:0:0:1%lo0");
+    when(inetAddr1.toString()).thenReturn("/0:0:0:0:0:0:0:1");
+    when(inetAddr2.toString()).thenReturn("/127u0.0.1");
```
mattnworb: typo in `127u`?

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -157,6 +161,31 @@
...
+    }
+    return findFirstInet4Addr(addrs);
+  }
+
+  public static String findFirstInet4Addr(Collection<InterfaceAddress> addrs)
+      throws HeliosDeploymentException {
+    for(InterfaceAddress addr : addrs) {
+      String addrStr = addr.getAddress().toString();
+      if(addrStr.matches("^/\\d+\\.\\d+\\.\\d+\\.\\d+$") &&
+          !addrStr.startsWith("/127")) {
```
mattnworb: I would suggest using `interfaceAddress.getAddress().isLoopbackAddress()` instead of testing against `/127`

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -157,6 +161,31 @@
...
+      addrs.addAll(nic.getInterfaceAddresses());
+    }
+    return findFirstInet4Addr(addrs);
+  }
+
+  public static String findFirstInet4Addr(Collection<InterfaceAddress> addrs)
+      throws HeliosDeploymentException {
+    for(InterfaceAddress addr : addrs) {
+      String addrStr = addr.getAddress().toString();
+      if(addrStr.matches("^/\\d+\\.\\d+\\.\\d+\\.\\d+$") &&
```
mattnworb: could test `addr.getAddress() instanceof Inet4Address` instead of doing a regex

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -157,6 +161,31 @@
...
+    for (NetworkInterface nic : Collections.list(nics)) {
+      addrs.addAll(nic.getInterfaceAddresses());
+    }
+    return findFirstInet4Addr(addrs);
+  }
+
+  public static String findFirstInet4Addr(Collection<InterfaceAddress> addrs)
+      throws HeliosDeploymentException {
+    for(InterfaceAddress addr : addrs) {
+      String addrStr = addr.getAddress().toString();
```
mattnworb: also I think you can just do `InetAddress.getHostAddress()` to get the ip address as a string, rather than `toString`, so you won't have to bother with the `/` as the host-ip separator.

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -157,6 +161,22 @@
...
 
+  /**
+   * Return first found non-loopback ip from enumeration of network devices.
+   * InetAddress.getLocalHost() will not work, as it returns the loopback device
+   */
+  public static String findFirstInet4Addr(Enumeration<NetworkInterface> nics)
+      throws HeliosDeploymentException {
+    for(NetworkInterface nic : Collections.list(nics)) {
+      for(InetAddress addr : Collections.list(nic.getInetAddresses())) {
+        if(addr instanceof Inet4Address && !addr.isLoopbackAddress()) {
```
danielhan: @mattnworb I made the changes you suggested.  One thing to note is that Inet4Address is not mockable, being a final class, nor does it have a public constructor.  As this prevents the possibility of cleanly unit testing, I combined the logic into a single method.

mattnworb: This might be helpful for creating stub inetaddress instances in your test: https://github.com/google/guava/blob/08983fb25b9f915d2ca51393697f34c9a3854bbd/guava/src/com/google/common/net/InetAddresses.java#L135

davidxia:
@danielhan What're the steps to repro the issue that this fixes? I was curious and wanted to run this code on both DFM and docker toolbox to see they both work now.

danielhan:
@davidxia i would agree that this is fairly complex.  to reproduce the issue, you have to set up a helios integration test environment with at least two containers, and one needs to talk to the other.  what is supposed to happen is that each container is supposed to know the host's ip address, in order to route requests to other containers.  however, on docker for mac, the containers get the 127.0.0.1 as the host's ip.

davidxia:
Is there a simple mvn verify cmd I can just run on a repo that’ll do this? I’m assuming this PR was motivated by a user raising an issue when running tests on their code?

helios-testing/src/main/java/com/spotify/helios/testing/HeliosSoloDeployment.java
```diff
@@ -173,10 +192,19 @@ private String determineHeliosHost(final Info dockerInfo) throws HeliosDeploymen
 
     if (dockerHostAddressIsLocalhost()) {
       if (isDockerForMac(dockerInfo)) {
+        log.info("determineHeliosHost: local environment appears to be Docker for Mac");
         try {
-          log.info("determineHeliosHost: local environment appears to be Docker for Mac");
-          return InetAddress.getLocalHost().getHostAddress();
-        } catch (UnknownHostException e) {
+          InetAddress localhost = InetAddress.getLocalHost();
+          if (localhost.isLoopbackAddress()) {
```
danielhan: @davidxia @mattnworb updated logic so that we only guess the IP when we find loopback

danielhan:
@mattnworb @davidxia is this worth doing?  folks on Docker for Mac can address the issue by mapping their Mac's name to an externally routable IP address.  There is no clean way I can think of to programmatically discover an IP address that is guaranteed to work in all scenarios.

davidxia:
Given that this seems to be an edge case and the difficulties you've had trying to do it in code, I'm fine updating the [helios FAQ](https://ghe.spotify.net/helios/helios/blob/master/docs/faq.md) with a workaround.

danielhan:
@dxia https://github.com/spotify/helios/blob/master/docs/integration_tests_with_helios_solo.md#docker-for-mac-address-for-temporaryjob-containers-is-returned-as-127001 exists.  i did not see this

davidxia:
:( Do you want to close this PR and just point affected users to that part of the docs?

PR #1198, opened 2017-12-14
client: Reduce oauth scope to readonly

caipre:
@davidxia @mattnworb

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/client/GoogleCredentialsAccessTokenProvider.java
+++ b/helios-client/src/main/java/com/spotify/helios/client/GoogleCredentialsAccessTokenProvider.java
@@ -91,6 +91,6 @@ class GoogleCredentialsAccessTokenProvider {
   static AccessToken getAccessToken() throws IOException {
     // Google Service Account Credentials require an access scope before calling `refresh()`;
     // see https://cloud.google.com/compute/docs/access/service-accounts#accesscopesiam.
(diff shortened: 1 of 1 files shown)


helios-client/src/main/java/com/spotify/helios/client/GoogleCredentialsAccessTokenProvider.java
```diff
@@ -91,6 +91,6 @@ static AccessToken getAccessToken(final GoogleCredentials credentials,
   static AccessToken getAccessToken() throws IOException {
     // Google Service Account Credentials require an access scope before calling `refresh()`;
     // see https://cloud.google.com/compute/docs/access/service-accounts#accesscopesiam.
-    return getAccessToken(singletonList("https://www.googleapis.com/auth/cloud-platform"));
+    return getAccessToken(singletonList("https://www.googleapis.com/auth/cloud-platform.read-only"));
```
davidxia: Shouldn't we use `cloudplatformprojects.readonly` based on https://ghe.spotify.net/system-z/system-z-frontend/pull/2222#issuecomment-1864246?

davidxia: Hm, I'm actually not sure what makes sense from an open source POV.

caipre: Probably this should be configurable, but read-only is a good first step.

PR #1199, opened 2017-12-14
Add equals(), hashCode(), and toString() to ServiceRegistration

davidxia:


davidxia:
Thanks. My intellij was set on the wrong option.

davidxia:
updated

davidxia:
For context (should've mentioned this sooner), this is needed for me to use ServiceRegistration as a key in a map.

PR #1201, opened 2017-12-22
Revert "Upgrade docker-client dep from 8.9.0 to 8.9.2"

mattnworb:
Reverts spotify/helios#1190

We've seen problems with jobs that use `exec` healthchecks since this change, see https://github.com/spotify/docker-client/pull/916#issuecomment-352896897


caipre:
The motivation seems to've just been to keep in sync with `docker-client`, so I'm fine with this revert until we sort out the issues. 👍

PR #1202, opened 2018-01-04
agent: Report when container exits in healthchecks

caipre:
The `TaskMonitor` has logic to throttle a container if the container
restarts too frequently. Prior to this change, however, the throttling
logic is bypassed when a container exits during healthchecks, since in
that case we throw an exception. The `TaskRunner` should report the
container's exit, even if the container never reached "running".

@mattnworb @davidxia


caipre:
Note that though the exception is caught, `listener.failed` does not force a re-evaluation of the throttling logic. https://github.com/spotify/helios/blob/master/helios-services/src/main/java/com/spotify/helios/agent/TaskMonitor.java#L99

PR #1203, opened 2018-01-15
client: Make Google token scopes configurable

davidxia:
* Add userinfo.email as a default scope to Google token.


helios-client/src/main/java/com/spotify/helios/client/HeliosClient.java
```diff
@@ -609,6 +610,9 @@ public static Builder newBuilder() {
     private boolean sslHostnameVerification = true;
     private boolean googleCredentialsEnabled = true;
     private AccessToken googleAccessToken;
+    private List<String> googleAccessTokenScopes = ImmutableList.of(
```
caipre: What about moving these to `GoogleCredentialsAccessTokenProvider.DEFAULT_SCOPES`?

caipre:
👍

PR #1206, opened 2018-01-26
Mark project as bug-fix only, no new features

davidxia:


caipre:
This works for me. I think we'll need to have a discussion with the community to see if a transfer of ownership makes sense, or if we want to provide a roadmap toward archiving the project.

PR #1207, opened 2018-02-14
Generate new Google access token at request time

davidxia:
instead of HeliosClient creation time. This will prevent users
who create a HeliosClient once and reuse it from accidentally
using expired tokens.


caipre:
This fetches a new token for every request. Did you look into how difficult it might be to grab a refresh token?

davidxia:
I just looked a bit. Would using a refresh token be better in some way? You still need to make a request that goes over the network. I just didn't see any advantages to doing that.

davidxia:
Any other thoughts on this?

mattnworb:
Looking at the source code I appear to be wrong about the GoogleCredentials class - I don't see a method that you can call which internally do the "check if expires soon, if so refresh, else return already present token" dance for you. This seems to only be in methods like [`Map<String, List<String>> getRequestMetadata(URI uri)`](https://github.com/google/google-auth-library-java/blob/v0.7.1/oauth2_http/java/com/google/auth/oauth2/OAuth2Credentials.java#L127-L139). It would be nice if we could use one of the methods already in that library to avoid reimplementing that.

davidxia:
@mattnworb See new commit. Is this what you meant?

davidxia:
I thought our approach was to generate a new token right before each request was made so we would never need to check and refresh?

mattnworb:
`GoogleCredentials.getAccessToken()` looks like it will return `null` if you haven't called `refresh` at least once. I don't think we should fetch a token from Google on every single HTTP request in HeliosClient as that adds overhead, and typically many methods in HeliosClient are called at the same time to fulfill some higher-level operation.

Also now that most of the work is being done in the `Supplier<Optional<AccessToken>>` being injected into the `AuthenticatingHttpConnector` I think it would be good to add a test of that code as it seems like we can't verify the behavior just from looking at the code.

davidxia:
Opened [issue here](https://github.com/google/google-auth-library-java/issues/147) asking if the oauth2 lib can expose a method to check expiration and refresh token if expired.

In the meantime, I think we can use `getRequestMetadata(URI uri)` with a dummy URI. The implementation that's relevant doesn't seem to use uri for anything.

davidxia:
See new commit. `bin/helios -d guc3 hosts heliosagent -vvv` shows the same access token is used for both HTTP requests.

PR #1208, opened 2018-02-21
Removes uses of deprecated 'HostAndPort.getHostText' method.

andrewapollonsky-spotify:
Instead use HostandPort.getHost()

### DIFF
--- a/docs/testing_framework.md
+++ b/docs/testing_framework.md
@@ -106,7 +106,7 @@ public class ServiceIT {
     // allocated port of the job
     HostAndPort port = temporaryJob.address("wiggum");
     String pingUri = String.format("http://%s:%d/ping",
-                                   port.getHostText(),
+                                   port.getHost(),
                                    port.getPort());
 
     HttpResponse response = client.execute(new HttpGet(pingUri));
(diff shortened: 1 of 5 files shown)


mattnworb:
Currently Helios declares that it uses Guava 20. In practice, people that depend on helios libraries might be able to use newer versions than this, provided that those newer versions haven't removed methods that Helios depends on, like `HostAndPort.getHostText()` mentioned in the PR summary. They might also be able to use _older_ versions of the library, provided that that version has all the classes and methods on those classes that Helios needs to call.

`HostAndPort.getHostText()` was [removed in Guava 22](https://github.com/google/guava/commit/870046dfd34b9d2d2c944015ffce5c7e7e3406ba#diff-96c26bb4dc9802f742174dd86eb894f1).

`HostAndPort.getHost()` was [introduced in Guava 20](https://github.com/google/guava/commit/b0babb69b05ed4d15cce74635ae96cf8ba78c85f).

So effectively I think this change will make the requirement on Guava be >= 20 (which it may have already been, if we are using other methods introduced specifically in 20). Given that the pom already states 20, that sounds good to me.

👍

PR #1210, opened 2018-03-19
refactor GoogleCredentialsAccessTokenSupplier

mattnworb:
These PR contains two refactorings:

1. throw exception if file at HELIOS_GOOGLE_CREDENTIALS cannot be read

    This changes the behavior when the HELIOS_GOOGLE_CREDENTIALS environment variable is set to a file that does not exist or cannot be read from falling to Application Default Credentials to instead throw an exception.

    If the environment variable contains invalid configuration, it seems better to throw an exception rather than silently use a different set of (possibly incorrect) credentials.

    We have seen some confusing behavior internally arising from the fact that this class would load an unexpected set of credentials when the file at the environment variable could not be loaded.

2. introduce a AuthorizationHeaderSupplier interface

    This is a refactoring of the GoogleCredentialsAccessTokenSupplier class and its usage in AuthenticationHttpConnector and HeliosClient.Builder to make those classes less tied to the Google Cloud OAuth2 token concepts.

    The goal here is to make it much simpler to add non-Google implementations of this interface in the future without having to have other layers of Helios (such as AuthenticatingHttpConnector) aware of those implementation details.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/client/AuthenticatingHttpConnector.java
+++ b/helios-client/src/main/java/com/spotify/helios/client/AuthenticatingHttpConnector.java
@@ -24,10 +24,8 @@ import static java.net.HttpURLConnection.HTTP_FORBIDDEN;
 import static java.net.HttpURLConnection.HTTP_UNAUTHORIZED;
 import static java.util.Collections.singletonList;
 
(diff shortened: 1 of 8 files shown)


Alexanderjone:
#1210

PR #1211, opened 2018-03-29
Fix bug that erases metadata from job JSON file

davidxia:
This fixes a bug that causes metadata passed in for job creation
from the JSON file is erased by the absence of the `--metadata`
switch in the CLI command.


caipre:
👍🏻

PR #1218, opened 2018-06-08
Prevent creation of jobs with external ports and grace period

davidxia:
Presently, the default `gracePeriod` is 600 seconds (10 minutes)
and the default `rolloutTimeout` is 300 seconds (5 minutes).
These defaults interact poorly when static ports are in use:
the old container holds the port for 10 minutes, blocking the new
container from starting and causing the entire deployment to fail.
This behavior shouldn't be allowed.

Behavior with this PR:

```
cat helios-job.json

{
  "image": "example.net/foo/bar",
  "gracePeriod": 500,
  "ports": {
    "http": {
      "internalPort": 8080,
      "protocol": "tcp"
    }
  },
  "registration": {
    "my-cool-service/http": {
      "ports": {
        "http": {}
      }
    }
  },
  "volumes": {
    "/some/path:ro": "/another/path"
  }
}

helios -z http://localhost:5801 create dxia-test:1 -f helios-job.json

Creating job: {"addCapabilities":[],"command":[],"created":null,"creatingUser":null,"dropCapabilities":[],"env":{},"expires":null,"gracePeriod":500,"healthCheck":null,"hostname":null,"id":"dxia-test:1","image":"example.net/foo/bar","labels":{},"metadata":{},"networkMode":null,"ports":{"http":{"externalPort":8080,"internalPort":8080,"ip":null,"protocol":"tcp"}},"ramdisks":{},"registration":{"my-cool-service/http":{"ports":{"http":{}}}},"registrationDomain":"","resources":null,"rolloutOptions":null,"secondsToWaitBeforeKill":null,"securityOpt":[],"token":"","volumes":{"/some/path:ro":"/another/path"}}

Failed: CreateJobResponse{status=INVALID_JOB_DEFINITION, errors=[Job has a grace period and static/external ports. Please choose one or the other.], id='dxia-test:1:eb7f2180f0e935671b7eeed9d841584a3836e932'}
```

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
@@ -23,6 +23,7 @@ package com.spotify.helios.common;
 import static com.google.common.base.Strings.isNullOrEmpty;
 import static java.lang.String.format;
 import static java.util.Collections.emptySet;
+import static java.util.Collections.singleton;
 import static java.util.regex.Pattern.compile;
 
 import com.google.common.base.Joiner;
@@ -208,7 +209,18 @@ public class JobValidator {
       }
     }
 
+    errors.addAll(validateGracePeriodAndExternalPorts(job));
+
     return errors;
+  }
+
+  private Set<String> validateGracePeriodAndExternalPorts(final Job job) {
+    final Integer gracePeriod = job.getGracePeriod();
+    if (gracePeriod != null && gracePeriod > 0 && job.hasExternalPorts()) {
+      return singleton("This configuration will prevent new containers from deploying during the "
+                       + "gracePeriod because of port conflicts.");
+    }
+    return emptySet();
   }
 
   /**
--- a/helios-client/src/test/java/com/spotify/helios/common/JobValidatorTest.java
+++ b/helios-client/src/test/java/com/spotify/helios/common/JobValidatorTest.java
@@ -21,6 +21,7 @@
 package com.spotify.helios.common;
 
 import static com.google.common.collect.Sets.newHashSet;
+import static com.google.common.io.Resources.getResource;
 import static com.spotify.helios.common.descriptors.Job.DEFAULT_NETWORK_MODE;
 import static com.spotify.helios.common.descriptors.Job.EMPTY_CAPS;
 import static com.spotify.helios.common.descriptors.Job.EMPTY_COMMAND;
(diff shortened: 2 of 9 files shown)


davidxia:
cc @myott @vbhavsar

helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
```diff
@@ -208,9 +209,20 @@ public JobValidator(final boolean shouldValidateJobHash,
...
+    errors.addAll(validateGracePeriodAndExternalPorts(job));
+
     return errors;
   }
 
+  private Set<String> validateGracePeriodAndExternalPorts(final Job job) {
+    final Integer gracePeriod = job.getGracePeriod();
+    if (gracePeriod != null && gracePeriod > 0 && job.hasExternalPorts()) {
+      return singleton(
+          "Job has a grace period and static/external ports. Please choose one or the other.");
```
caipre: An explanation of why this is a problem may be useful (or a link to docs explaining it). Something like "This configuration will prevent new containers from deploying during the gracePeriod because of port conflicts."

mattnworb: How about adding a brief description of the reason only one of the two can be chosen? Maybe just as simple as "Please choose one or the other, as the two conflict with each other".

helios-client/src/test/java/com/spotify/helios/common/JobValidatorTest.java
```diff
@@ -596,4 +601,30 @@ public void testJobFromJsonWithInvalidRegistration() throws Exception {
...
+
+    final Job externalPorts = Json.read(
+        stringFromResource("job-with-external-ports.json"), Job.class);
+    assertThat(validator.validate(externalPorts), Matchers.<String>empty());
+
+    final Job externalPortsAndGracePeriod = Json.read(
+        stringFromResource("job-with-external-ports-and-grace-period.json"), Job.class);
+    assertThat(validator.validate(externalPortsAndGracePeriod),
+        hasItem("Job has a grace period and static/external ports. "
+                + "Please choose one or the other."));
```
mattnworb: is there a test for grace period and no static ports?

davidxia: Thanks. Added.

helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
davidxia: Thanks. Updated with "This configuration will prevent new containers from deploying during the gracePeriod because of port conflicts"

PR #1221, opened 2018-07-03
migrating from circleci 1.0 to 2.0

lavanyachennupati:


.circleci/config.yml
```diff
@@ -0,0 +1,58 @@
...
+    # 'See docs on artifact collection here https://circleci.com/docs/2.0/artifacts/'
+    - run: mkdir -p $CIRCLE_ARTIFACTS $CIRCLE_TEST_REPORTS
+
+    - run:
+        name: Install dependencies
+        working_directory: ~/spotify/helios
+        command: |
+         apt-get -y -qq update
+         apt-get -y -qq install jq
+         update-alternatives --set java /usr/lib/jvm/jdk1.8.0/bin/java; sudo update-alternatives --set javac /usr/lib/jvm/jdk1.8.0/bin/javac; echo -e "export JAVA_HOME=/usr/lib/jvm/jdk1.8.0" >> $BASH_ENV
```
mattnworb: could be worthwhile to extract `jdk1.8.0` or `1.8.0` to a parameter, see an example of how that is done in https://github.com/spotify/scio/blob/master/.circleci/config.yml for `scala_212_version`

.circleci/config.yml
```diff
@@ -0,0 +1,58 @@
...
+         if [[ $(python --version 2>&1) = *"2.7.10"* ]] || pyenv versions --bare | grep -x -q '2.7.10'; then pyenv global 2.7.10;else pyenv install --skip-existing  2.7.10 && pyenv global 2.7.10 && pyenv rehash && pip install virtualenv && pip install nose && pip install pep8 && pyenv rehash;fi
```
mattnworb: could you a comment here for future readers about why we need to install / set python 2.7.10, and why virtualenv/nose/pep8 is needed?

myott:
This change makes the CircleCI 2 build work like 1. Docker is already running on the node, we just need to configure (see pre_machine), then restart docker.

lavanyachennupati:
@davidxia. circleCi2.0 gives us the freedom to choose an executor type. 
https://circleci.com/docs/2.0/executor-types/ . We were running into issues with connecting to docker from within the tests using the docker executor. Also mounting volumes was challenging. Using machine executor type solves most of the issues with connecting to docker from the tests.

.circleci/config.yml
lavanyachennupati: done!

PR #1222, opened 2018-07-10
Relax volume validation requirements

caipre:
This commit adds support for mounting named volumes by relaxing the
validation performed by the `JobValidator`. This allows jobs to use
options for volumes managed by a particular driver, such as:

```
docker volume create --driver local \
   --opt type=btrfs --opt device=/dev/sda2 \
   myvolume
```

The above volume can then be mounted into a container using the
following helios job configuration (full configuration omitted):

```
{
  ...
  "volumes": {
    "/mnt/myvol": "myvolume"
  }
}
```

@mattnworb @davidxia

### DIFF
--- a/docs/user_manual.md
+++ b/docs/user_manual.md
@@ -428,7 +428,7 @@ see [volumes][] in the Docker reference for full details. The container path can
 be suffixed with "rw" or "ro" to create a read-write or read-only volume mount,
 respectively.
 
-Format: `[container-path]:[rw|ro]:[host-path|volume-name]`.
+Format: `"[container-path]:[rw|ro]": "[host-path|volume-name]"`.
 
 Examples:
 
--- a/docs/user_manual.md
+++ b/docs/user_manual.md
@@ -421,13 +421,22 @@ container.
 #### volumes
 Container volumes. Optional. 
 
-Specify either a single path to create a data volume, or a source path and a
-container path to mount a file or directory from the host. 
-  
-The container path can be suffixed with "rw" or "ro" to create a read-write or
-read-only volume, respectively.
+Specify a single path to create a data volume, a container path and a source
+path to mount a directory from the host, or a container path and a name to mount
+a named volume. The named volume will be created by docker if it does not exist;
+see [volumes][] in the Docker reference for full details. The container path can
+be suffixed with "rw" or "ro" to create a read-write or read-only volume mount,
+respectively.
 
-Format: `[container-path]:[rw|ro]:[host-path]`.
+Format: `[container-path]:[rw|ro]:[host-path|volume-name]`.
+
+Examples:
+
+  * Mount a new data volume (name assigned by docker) to `/foo` in the container: `"volumes": {"/foo": ""}`
+  * Mount `/src` from the host to `/dst` in the container: `"volumes": {"/dst": "/src"}`
+  * Mount a volume named `my-vol` to `/dst` in the container: `"volumes": {"/dst": "my-vol"}`
+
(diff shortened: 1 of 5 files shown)


docs/user_manual.md
```diff
@@ -421,13 +421,22 @@ container.
...
-The container path can be suffixed with "rw" or "ro" to create a read-write or
-read-only volume, respectively.
+Specify a single path to create a data volume, a container path and a source
+path to mount a directory from the host, or a container path and a name to mount
+a named volume. The named volume will be created by docker if it does not exist;
+see [volumes][] in the Docker reference for full details. The container path can
+be suffixed with "rw" or "ro" to create a read-write or read-only volume mount,
+respectively.
+
+Format: `[container-path]:[rw|ro]:[host-path|volume-name]`.
```
mattnworb: The format here doesn't match what is mentioned in the bullet points below, since the latter is representing it as a JSON object. Could you make this line match the below (i.e. with double quotes)?

helios-client/src/main/java/com/spotify/helios/common/JobValidator.java
```diff
@@ -161,13 +161,10 @@ public JobValidator(final boolean shouldValidateJobHash,
         errors.add("Volume path is not absolute: " + path);
         continue;
       }
-      if (!isNullOrEmpty(source) && !source.startsWith("/")) {
-        errors.add("Volume source is not absolute: " + source);
-        continue;
-      }
```
mattnworb: Should we still do some validation of `String source = entry.getValue()` ? For instance I'd expect that if it contained a `/`, it would need to be an absolute path. I'd imagine that for the other types of volumes a `/` would never be allowed (although trying to reimplement docker's validation logic would be overkill)

helios-client/src/test/java/com/spotify/helios/common/JobValidatorTest.java
```diff
@@ -169,6 +169,7 @@ public void testValidHostnamesPass() {
   @Test
   public void testValidVolumesPass() {
     final Job j = Job.newBuilder().setName("foo").setVersion("1").setImage("foobar").build();
+    assertThat(validator.validate(j.toBuilder().addVolume("/foo", "bar").build()), is(empty()));
```
mattnworb: could you add a test of `.addVolume("/foo", "")` too? (if it is not covered elsewhere)

helios-services/src/main/java/com/spotify/helios/agent/TaskConfig.java
```diff
@@ -112,7 +112,7 @@ public ContainerConfig containerConfig(final ImageInfo imageInfo,
     builder.hostname(job.getHostname());
     builder.env(containerEnvStrings());
     builder.exposedPorts(containerExposedPorts());
-    builder.volumes(volumes());
+    builder.volumes(volumes().keySet());
```
mattnworb: why is this changed, wouldn't it lose the source part of the volume specification?

caipre:
> Do you know what would happen if a Helios job had a volume spec like "/foo/bar":"named-volume" and that volume did not exist?

When I tried this locally Docker created the volume and mounted it as expected.

helios-services/src/main/java/com/spotify/helios/agent/TaskConfig.java
caipre: The method accepting a `Map<String, Map>` [was deprecated](https://github.com/spotify/docker-client/blob/353417a693174296b4cba538e8c7ef5ae177b3e9/src/main/java/com/spotify/docker/client/messages/ContainerConfig.java#L278-L286).

PR #1227, opened 2018-08-15
Upgrade google-auth-library-java from 0.7.1 to 0.10.0

davidxia:
so we can use `Oauth2Credentials.refreshAccessToken()` instead of `getRequestMetadata()`.

See https://github.com/google/google-auth-library-java/issues/147.


davidxia:
cc @spotify/helios-team

Issue #1177, opened 2017-11-01
OldJobReaper should not garbage collects jobs in use by empty DeploymentGroups


mattnworb:
Currently the [OldJobReaper](https://github.com/spotify/helios/blob/08698a663cb166bf044e789cf9bde261b71a2b01/helios-services/src/main/java/com/spotify/helios/master/reaper/OldJobReaper.java) will remove any helios job that is not deployed, meaning that there is no host that the master is aware of that should be running the job.

Internally at Spotify, usually when we deploy a new revision of a service/container to Helios, we create the Job, a DeploymentGroup, and do a `rolling-update job group` on _all_ helios clusters that we have, even if there are no hosts currently in that deployment group. The intention here is to make it possible for the service/container owners to be able to easily add hosts/capacity in a site that currently has no hosts of that group/selector and have Helios automatically deploy the same job that is running in other sites to the new hosts in this new site.

The OldJobReaper causes problems with this intention though, as sometimes the job that has been associated with the empty DeploymentGroup has been deleted. This means that when the service/container owner provisions new hosts for the first time in a site that currently has no hosts in that group, their job is sometimes not automatically deployed. This leads to inconsistent experience for these service/container owners, as sometimes when they add capacity to a new site the Helios job is automatically deployed, and sometimes it is not.

I would propose that ideally OldJobReaper should add a condition when deciding to reap a job to check that the job is not currently associated with any DeploymentGroup entities, preventing the inconsistent behavior mentioned above.

One possible way to implement this would be to have the OldJobReaper, at the start of an iteration, fetch all existing DeploymentGroups, and then when iterating over each job to decide whether to reap it check to see if any of the DeploymentGroups have that job in their `deploymentGroup.jobId` field. Fetching the set of DeploymentGroups at startup would avoid the need for a `O(n^2)` algorithm like `for job in model.getJobs(), for group in model.getDeploymentGroups: ...`.

caipre:
I think this is probably fine, and a worthwhile improvement for our users. It's sensible to not remove objects when other objects depend on them.

I don't think we have an `OldDeploymentGroupReaper` though, so the next natural question is whether we'd want to introduce something like that. My hunch is that the number and on-disk cost of these is very small, to the point where we don't really need to worry about it.

mattnworb:
[...]

PR #1231, opened 2018-08-28
Do not reap jobs that are part of a DeploymentGroup

myott:


myott:
@spotify/docker-admins

myott:
Fixes https://github.com/spotify/helios/issues/1177

helios-services/src/test/java/com/spotify/helios/master/reaper/OldJobReaperTest.java
```diff
@@ -126,55 +135,176 @@ public JobStatus getJobStatus() {
...
+    final Datapoint datapoint = new Datapoint(
+        name.getMethodName(),
+        emptyMap(),
+        events(ImmutableList.of(HOURS.toMillis(2), HOURS.toMillis(3))));
+
+    testReap(datapoint, masterModel, true);
+  }
+
+  private void testReap(
+      final Datapoint datapoint,
```
mattnworb: with this refactoring of the test, the `Datapoint` class seems a bit less useful, I wonder if it would just be simpler to pass the data in Datapoint's fields into this method as parameters?

helios-services/src/main/java/com/spotify/helios/master/reaper/OldJobReaper.java
```diff
@@ -79,6 +82,8 @@
   private final long retentionMillis;
   private final Clock clock;
 
+  private Set<JobId> jobsInDeploymentGroups;
```
mattnworb: unfortunate that the structure of the superclass means you have to store this data in a class field rather than a local somewhere, but I suppose not using the structure imposed on you by the superclass would be more annoying

helios-services/src/test/java/com/spotify/helios/master/reaper/OldJobReaperTest.java
myott: 👍 absolutely

PR #1232, opened 2018-08-29
Add support for container runtime

vbhavsar:
This PR enables [setting runtime](https://docs.docker.com/engine/api/v1.37/#operation/ContainerCreate) when creating a container. This allows using GPUs as container runtimes, for example.

### DIFF
--- a/docs/user_manual.md
+++ b/docs/user_manual.md
@@ -200,6 +200,7 @@ example that uses all the available configuration keys with an explanation of ea
   "secondsToWaitBeforeKill": 120,
   "securityOpt" : [ "label:user:USER", "apparmor:PROFILE" ],
   "token": "insecure-access-token",
+  "runtime": "nvidia",
   "volumes" : {
     "/destination/path/in/container.yaml:ro" : "/source/path/in/host.yaml"
   }
@@ -400,6 +401,13 @@ Will ultimately have these options at `rolling-update` time.
 
 Since timeout, migrate, and token weren't specified either via CLI or job config, they get their
 values from the defaults in [`RolloutOptions`][rollout-options-code].
+
+#### runtime
+Optional. Container runtime to use. When specifying a runtime, you will need to ensure that 
+the docker daemon knows about the specified runtime. For example, to enable nvidia runtime,
+you will need to configure the host to install runtime.
+
+[Puppet example](https://ghe.spotify.net/puppet/spotify-puppet/blob/d2c2d8da25299ec591a7fd67d70037b5cfbb8713/modules/nvidia_docker2/manifests/init.pp).
 
 #### secondsToWaitBeforeKill
 Optional. When a job is being stopped or undeployed, the helios-agent will ask
--- a/docs/user_manual.md
+++ b/docs/user_manual.md
@@ -407,7 +407,7 @@ Optional. Container runtime to use. When specifying a runtime, you will need to 
 the docker daemon knows about the specified runtime. For example, to enable nvidia runtime,
 you will need to configure the host to install runtime.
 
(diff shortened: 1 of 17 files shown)


docs/user_manual.md
```diff
@@ -401,6 +402,13 @@ Will ultimately have these options at `rolling-update` time.
 Since timeout, migrate, and token weren't specified either via CLI or job config, they get their
 values from the defaults in [`RolloutOptions`][rollout-options-code].
 
+#### runtime
+Optional. Container runtime to use. When specifying a runtime, you will need to ensure that 
+the docker daemon knows about the specified runtime. For example, to enable nvidia runtime,
+you will need to configure the host to install runtime.
+
+[Puppet example](https://ghe.spotify.net/puppet/spotify-puppet/blob/d2c2d8da25299ec591a7fd67d70037b5cfbb8713/modules/nvidia_docker2/manifests/init.pp).
```
mattnworb: Since this is on github.com and open source, shouldn't link to our internal GHE :)

helios-services/src/test/java/com/spotify/helios/agent/TaskConfigTest.java
```diff
@@ -52,6 +52,7 @@
   private static final Set<String> CAP_ADDS = ImmutableSet.of("cap1", "cap2");
   private static final Set<String> CAP_DROPS = ImmutableSet.of("cap3", "cap4");
   private static final Map<String, String> LABELS = ImmutableMap.of("label", "value");
+  private static final String RUNTIME = "nvidia";
```
mattnworb: I'd skip the constant since it is used in only one method

helios-client/src/main/java/com/spotify/helios/common/descriptors/Job.java
```diff
@@ -272,6 +277,7 @@ public Job(
     this.secondsToWaitBeforeKill = secondsToWaitBeforeKill;
     this.ramdisks = firstNonNull(ramdisks, EMPTY_RAMDISKS);
     this.rolloutOptions = rolloutOptions;
+    this.runtime = Optional.fromNullable(runtime).orNull();
```
myott: Appears that you can do `this.runtime = runtime` here.

vbhavsar: Hmm, you're right. I wonder why other fields follow this pattern.

mattnworb: It is a weird pattern that I don't think should be continued. This code used to do a lot of `Optional.fromNullable(foo).orElse(blah).get()`, but at some point that was changed to use `MoreObjects.firstNonNull(a, b)` as there is no real reason to create an `Optional` instance that is then discarded immediately.

vbhavsar: Will address this in a separate PR.

docs/user_manual.md
```diff
@@ -401,6 +402,13 @@ Will ultimately have these options at `rolling-update` time.
 Since timeout, migrate, and token weren't specified either via CLI or job config, they get their
 values from the defaults in [`RolloutOptions`][rollout-options-code].
 
+#### runtime
+Optional. Container runtime to use. When specifying a runtime, you will need to ensure that 
+the docker daemon knows about the specified runtime. For example, to enable nvidia runtime,
+you will need to configure the host to install runtime.
```
davidxia: @vbhavsar, is "runtime" word at the end of this sentence correct? Should it be "nvidia" or "nvidia driver" instead?

vbhavsar: It is a bit confusing to read. I will change the wording.

PR #1234, opened 2018-09-21
Add runtime config support to CLI switch and inspect cmd

davidxia:
Allow CLI to override Job config file's runtime property via `--runtime` switch.
Every other Job property has this behavior.

Show runtime value in `inspect` command's output.


davidxia:
cc @spotify/docker-admins

caipre:
👍

PR #1240, opened 2018-10-06
Fix errors when building with JDK11

davidxia:
* Upgrade various plugins whose old versions don't work on JDK11.
* Upgrade com.spotify:foss-root pom from 5 to 9.
* Fix various checkstyle and Javadoc errors resulting from upgrades of above.


mattnworb:
I suggest combining this with #1242, so that the CI can tell us if these are all the changes that are needed or not.

Does Circle support a build matrix like Travis?

davidxia:
It does. I've moved this commit to #1242.

PR #1244, opened 2018-10-11
Fix errors when building with Java 11; test compiling on Circle

davidxia:
* Upgrade various plugins whose old versions don't work on JDK11.
* Upgrade com.spotify:foss-root pom from 5 to 9.
* Fix various checkstyle and Javadoc errors resulting from upgrades of
  above.

Add the Dockerfile for the image in which we compile helios on Java 11.
Circle CI 2.0 doesn't provide Java 11 with its machine executor. So we must
use the Docker executor with a custom environment in the form of a
container pre-installed with Java 11 and other build tools.

* run tests on Circle CI on Java 11

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,9 +18,6 @@ jobs:
     machine: true
     
     steps:
-    # Machine Setup
-    #   If you break your build into multiple jobs with workflows, you will probably want to do the parts of this that are relevant in each
-    # The following `checkout` command checks out your code to your working directory. In 1.0 this is done implicitly. In 2.0 you can choose where in the course of a job your code should be checked out.
     - checkout
 
     # 'See docs on artifact collection here https://circleci.com/docs/2.0/artifacts/'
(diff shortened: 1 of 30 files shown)


davidxia:
cc @spotify/docker-admins

.circleci/config.yml
```diff
@@ -64,3 +61,30 @@ jobs:
...
+      MAVEN_OPTS: -Xmx128m
+
+    steps:
+    - checkout
+
+    - run: ./circle.sh pre_machine
+
+    - run:
+        command: ./circle.sh dependencies
+        no_output_timeout: 20m
```
davidxia: I had to increase the timeout. A docker executor seems to have less resources and therefore is slower. Without this increased timeout, downloading deps often timed out and failed the build.

circle.sh
```diff
@@ -2,23 +2,17 @@
 
 case "$1" in
   pre_machine)
-    # ensure correct level of parallelism
```
davidxia: This check doesn't seem to be valuable anymore. It'll be simpler to remove it since my java11 build is parallelism 1.

.circleci/config.yml
mattnworb: could you add this as a comment into the config for future readers?

davidxia:
yea

.circleci/config.yml
davidxia: https://github.com/spotify/helios/pull/1248

PR #1245, opened 2018-10-11
Fix illegal reflective access operation in DefaultHttpConnector

davidxia:
related to #1158


mattnworb:
are you sure the test errors weren't flukes?

I tried to re-run the build but it fails since the branch is deleted

PR #1247, opened 2018-10-12
circle: remove unnecessary node anchor and comment in config

davidxia:


davidxia:
cc @spotify/docker-admins

PR #1250, opened 2018-10-14
update docker-client to java 9 compatible version

AKkilla69:
[Order Number.txt](https://github.com/spotify/helios/files/2476320/Order.Number.txt)


davidxia:
So weird. Why the link to the order file that looks like it came from Amazon for a Milwaukee M18 18-Volt Lithium-Ion Cordless Combo Kit?

PR #1251, opened 2018-10-17
Add prefix to client.jobs() call

caipre:
With no arguments, the client.jobs() call causes helios-master to
respond with a list all known jobs (unpaginated). If the master has
enough jobs, such a request is likely to timeout or be dropped due to
response size, which bubbles up as a failure in the application using
helios-testing.

This change adds the temporary jobs' prefix to the call so that only
jobs matching the prefix are returned. This should reduce (but does not
eliminate) the impact of this issue.

@mattnworb @davidxia


caipre:
Tests passing now. Wish they were easier to run locally.

caipre:
> We could also consider removing removeOldJobs() altogether

`s/removeOldJobs()/helios-testing/` ;)

PR #1252, opened 2018-10-17
Upgrade docker-client dep from 8.13.0 to 8.14.1

davidxia:


davidxia:
cc @spotify/docker-admins

PR #1254, opened 2018-10-18
add error-prone for static analysis

mattnworb:
https://errorprone.info/

also fixed one warning it found:
- https://errorprone.info/bugpattern/PreconditionsCheckNotNullPrimitive

--- 

also in this change is dropping a compiler target of 1.7 for a few of the modules (as error-prone is 8+ only).

helios-client, -testing and -tools are currently built with a 1.7
target. In 0bf718b these modules were excluded from being upgraded to
a 1.8 target to not exclude people who still used 1.7 or depended on it
(with the CLI, or Hadoop-using teams that were stuck on 1.7).

I'm not 100% certain it is worth doing that, but it was needed for error-prone.

### DIFF
--- a/helios-services/src/main/java/com/spotify/helios/agent/TaskRunner.java
+++ b/helios-services/src/main/java/com/spotify/helios/agent/TaskRunner.java
@@ -77,7 +77,7 @@ class TaskRunner extends InterruptingExecutionThreadService {
     this.listener = checkNotNull(builder.listener, "listener");
     this.existingContainerId = builder.existingContainerId;
     this.registrar = checkNotNull(builder.registrar, "registrar");
-    this.secondsToWaitBeforeKill = checkNotNull(builder.secondsToWaitBeforeKill, "waitBeforeKill");
+    this.secondsToWaitBeforeKill = builder.secondsToWaitBeforeKill;
     this.healthChecker = Optional.fromNullable(builder.healthChecker);
     this.serviceRegistrationHandle = Optional.absent();
     this.containerId = Optional.absent();
--- a/helios-tools/src/main/java/com/spotify/helios/cli/command/JobHistoryCommand.java
+++ b/helios-tools/src/main/java/com/spotify/helios/cli/command/JobHistoryCommand.java
@@ -90,7 +90,7 @@ public class JobHistoryCommand extends ControlCommand {
 
     for (final TaskStatusEvent event : events) {
       final String host = checkNotNull(event.getHost());
-      final long timestamp = checkNotNull(event.getTimestamp());
+      final long timestamp = event.getTimestamp();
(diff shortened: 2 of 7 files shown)


mattnworb:
@spotify/docker-admins 

this is pretty low priority, but I was adding errorprone to other projects and wanted to add it here as well. Turns out our code quality was already pretty good, so not much to change.

helios-services/src/main/java/com/spotify/helios/agent/TaskRunner.java
```diff
@@ -77,7 +77,7 @@ private TaskRunner(final Builder builder) {
     this.listener = checkNotNull(builder.listener, "listener");
     this.existingContainerId = builder.existingContainerId;
     this.registrar = checkNotNull(builder.registrar, "registrar");
-    this.secondsToWaitBeforeKill = checkNotNull(builder.secondsToWaitBeforeKill, "waitBeforeKill");
+    this.secondsToWaitBeforeKill = builder.secondsToWaitBeforeKill;
```
mattnworb: `builder.secondsToWaitBeforeKill` is a primitive type and therefore `checkNotNull` was pointless

helios-tools/src/main/java/com/spotify/helios/cli/command/JobHistoryCommand.java
```diff
@@ -90,7 +90,7 @@ int run(final Namespace options, final HeliosClient client, final PrintStream ou
 
     for (final TaskStatusEvent event : events) {
       final String host = checkNotNull(event.getHost());
-      final long timestamp = checkNotNull(event.getTimestamp());
+      final long timestamp = event.getTimestamp();
```
mattnworb: same reasoning as above, `getTimestamp()` is a primitive value and not an object

pom.xml
```diff
@@ -715,6 +715,31 @@
                 </plugins>
             </build>
         </profile>
+
+        <!-- using github.com/google/error-prone-javac is required when running on JDK 8 -->
+        <profile>
+            <id>jdk8</id>
+            <activation>
+                <jdk>1.8</jdk>
```
davidxia: This means this profile is activated if the jdk is 1.8? What happens if the jdk is another version?

mattnworb: the configuration in this profile with `-J-Xbootclasspath/p:` is only needed on Java 8. errorprone doesn't support < 8. on > 9 this method of overriding the system javac is not allowed.

more details in https://github.com/google/error-prone/blob/gh-pages/docs/installation.md#maven, sorry I forgot to add this link initially.

mattnworb: to be more specific, I think all this is needed because "Using the error-prone-javac is required when running on JDK 8"

PR #1255, opened 2018-11-14
set SUPPRESS_GCLOUD_CREDS_WARNING environment variable

mattnworb:
to stop showing warnings about using user credentials, which as a CLI we intend.


caipre:
👍

PR #1258, opened 2018-12-05
Add option to set timeout to Temporary Jobs created from Environment

davidals:


caipre:
@davidals Thanks for the PR. Mind adding some documentation about this new configuration option?

https://github.com/spotify/helios/blob/master/docs/testing_framework.md#configuration-by-file

docs/testing_framework.md
```diff
@@ -251,6 +251,8 @@ accomplish this. Within a profile, the following keys having meaning:
 * `hostPickingStrategyKey` - string - a string used to seed the random number
   generator.  It can be basically anything, but it's best for it to not
   be empty.
+* `heliosMasterTimeoutSeconds` - int - an integer with the timeout in seconds for the HeliosClient
```
mattnworb: I had to look this up in the code myself, but HeliosClient.Builder is using the timeout you pass it as both the `connect` and `read` timeout on the underlying HttpClient. It might be worth clarifying that here - it is not the timeout related to the deployment itself, but rather the HTTP messages sent to Helios' API.

davidals:
Renamed it to `heliosClientHttpTimeoutSeconds`, I think just `httpTimeout` could be misleading. Also kept the `in Seconds` information as well since the other timeouts are in Millis. But the HeliosClient.Builder only takes integers.

PR #1265, opened 2019-01-10
Add a unit file to manage helios-agent service

vbhavsar:
This sets up helios-agent to be configured as a daemon and managed via systemd.

@mattnworb


src/deb/helios-agent/helios-agent-wrapper
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
mattnworb: I think it would be good to add some comments here about why we had to add this wrapper script (since the `EnvironmentFile` etc directives in the unit file don't support running arbitrary shell scripts)

src/deb/helios-agent/helios-agent.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Helios agent daemon
```
mattnworb: I think we also want to add a `After=network.target` to this section to signal that this service should be started after the network is available

PR #1269, opened 2019-04-05
Adding global flag `--strict-start` to wildcard-based commands.

hstefan:
This forces job names to be matched from the string start. By default,
some commands will match on any jobs that contain the input name as a
substring, this option forces the match to only happen if the string
starts with the input name. Affects the subcommands: remove, inspect,
rolling-update, undeploy, deploy and stop.

The motivation behind this is production usages returning
"JOB_AMBIGUOUS_REFERENCE" in cases where a job's name is a substring of
another (eg: foo-bar and bar-foo-bar, as the first a substring of the
second).


helios-tools/src/main/java/com/spotify/helios/cli/CliParser.java
```diff
@@ -95,6 +95,7 @@
   private final List<Target> targets;
   private final String username;
   private boolean json;
+  private boolean strictStart;
```
mattnworb: could you add `strictStart` to `WildcardJobCommand` instead of as a global flag for all commands in CliParser? It looks to me like it is only used and needed in `WildcardJobCommand` instances.

PR #1275, opened 2019-09-03
excluding docker client from maven plugin

lavanyachennupati:
@smstone

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -13,3 +13,4 @@ coverage:
     patch:
       default:
         enabled: false
+        threshold: 10
--- a/helios-services/pom.xml
+++ b/helios-services/pom.xml
@@ -37,6 +37,9 @@
                         <artifactId>docker-maven-plugin</artifactId>
                         <version>1.2.0</version>
                         <configuration>
+                            <ignoredDependencies>
+                                <ignoredDependency>com.spotify.docker.client</ignoredDependency>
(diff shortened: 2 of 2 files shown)


lavanyachennupati:
@smstone @mattnworb @davidxia 
Running into https://github.com/spotify/dockerfile-maven/issues/236.

This is fixed in  8.14.1 of `docker-client`. We upgraded the version of `docker-client` to this version. 
But still are running into the above issue. Perhaps this could be because the  current version (8.14.1) of `docker-maven-plugin` 8.14.1 depends on an older version of `docker-client`. 

#1274 upgrades to using `dockerfile-maven-plugin` 1.4.12 which pulls in the correct `docker-client` version. But there are more changes needed for the `dockerfile-maven-plugin` to work, like requiring a docker file. (docker-maven-plugin currently used generates it for us).

Trying to see if the changes in this pr will help us solve the issue we are seeing.

mattnworb:
I'm not familiar with `ignoredDependencies` and I don't see it in the plugin source.

Does this work if you run it locally?

How about we release a new version of docker-maven-plugin with the updated docker-client?

davidxia:
> How about we release a new version of docker-maven-plugin with the updated docker-client?

Good idea. I made a PR here https://github.com/spotify/docker-maven-plugin/pull/430

PR #1278, opened 2019-09-13
client: periodically refresh helios master SRV records

davidxia:
The helios client currently resolves helios master SRV records
once at startup and keeps them for the duration of its lifetime.
If you have a long-lived client, it won't discover new masters.

This change introduces a new class `PeriodicResolver`.
This class delegates to `Resolver` and simply periodically
refreshes the SRV records it returns.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/PeriodicResolver.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/PeriodicResolver.java
@@ -1,0 +1,64 @@
+/*-
+ * -\-\-
+ * Helios Client
+ * --
+ * Copyright (C) 2016 - 2019 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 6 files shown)


helios-client/src/main/java/com/spotify/helios/common/Resolver.java
```diff
@@ -44,11 +48,11 @@ private static String env(final String name, final String defaultValue) {
     return Optional.fromNullable(getenv(name)).or(defaultValue);
   }
 
-  public static Supplier<List<URI>> supplier(final String srvName, final String domain) {
+  Supplier<List<URI>> supplier(final String srvName, final String domain) {
     return supplier(srvName, domain, DEFAULT_RESOLVER);
   }
 
-  static Supplier<List<URI>> supplier(final String srvName, final String domain,
+  Supplier<List<URI>> supplier(final String srvName, final String domain,
```
mattnworb: is there a need to keep this as returning `Supplier` after this PR? I see in `PeriodicResolver` you always call `.get()` immediately, and it seems like the only other callers of this method are in the unit test. How about make this just return a `List<URI>`?

helios-client/src/main/java/com/spotify/helios/common/PeriodicResolver.java
```diff
@@ -0,0 +1,64 @@
...
+public class PeriodicResolver implements Supplier<List<URI>> {
+
+  private List<URI> endpoints;
+
+  private PeriodicResolver(final String srvName,
+                           final String domain,
+                           final Resolver resolver,
+                           final ScheduledExecutorService executorService) {
+    endpoints = resolver.supplier(srvName, domain).get();
+    executorService.scheduleWithFixedDelay(() ->
```
mattnworb: a minor thing, but I like to launch the threads in the static method rather than the constructor, which I think makes unit testing a bit easier, if you ever need to test/control things like "return a new value after the thread fires again"

PR #1292, opened 2020-04-02
validate duplicate ports with protocols

firstway:
We deploy a service which uses SAME port but with different protocols (TCP and UDP), 
such as 
```
"gossip-tcp": {
   "internalPort": 11102,
   "externalPort": 11102,
   "protocol": "tcp"
},
"gossip-udp": {
  "internalPort": 11102,
   "externalPort": 11102,
   "protocol": "udp"
}

```
Current helios will reject this conf as it only considers ports when does  config validation.
this pr tried to address this issue.


firstway:
@mattnworb  any thoughts?

PR #1297, opened 2020-06-15
Change ZooKeeperMasterMode.staticPorts() to return Set instead of List

davidxia:
so later on helios-master won't try to delete the same ZK path multiple times
in a single transaction when undeploying a job that uses the same port
with different protocols.


firstway:
👍

PR #1301, opened 2020-08-06
Support jackson 2.11.0+

jwiklund:
Jackson 2.11.0 changes the default _tzSerializedWithColon value from false to true.
This causes a "Id hash mismatch" when trying to deploy a job.

withColonInTimeZone has been available since 2.9.1 so this will break on older version.
If this is an issue we would probably have to resort to reflection.

### DIFF
--- a/helios-client/src/main/java/com/spotify/helios/common/Json.java
+++ b/helios-client/src/main/java/com/spotify/helios/common/Json.java
@@ -36,6 +36,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectReader;
 import com.fasterxml.jackson.databind.ObjectWriter;
 import com.fasterxml.jackson.databind.type.TypeFactory;
+import com.fasterxml.jackson.databind.util.StdDateFormat;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


davidxia:
Just curious what you need newer jackson versions for. Is it to fix this vuln? https://github.com/spotify/helios/network/alert/pom.xml/com.fasterxml.jackson.core:jackson-databind/open

jwiklund:
@davidxia We have a mixed repository with both apollo and scio (to test the combination) and the newest scio 0.9.2 requires jackson 2.11.0.

davidxia:
Are you able to do some shading for just your service? Helios is in maintenance mode and rolling this out will affect the entire fleet and take a while.

jwiklund:
I have copied this Json file into our repository which seems to fix the problem for us. I guess we can keep this open if someone else encounters the same problem to avoid pushing this out.

davidxia:
oh nice, which JSON file are you referring to?

jwiklund:
I mean the Json.java file in this repository, I applied this PR's change and have it in our repository and it loads before the Json class from helios-client.jar :)

jwiklund:
Since it's only used for testing it's kind of okay :)

davidxia:
Whoa, what dark magic is this?

jwiklund:
Well pre modules everything is a flat namespace so java will load the first file that matches the name, I'm assuming if we turn on module support this way of hacking stuff will stop working but it's convenient when needed (and very annoying when it's done by misstake and you have to figure out why that class you can clearly see have a function does not).

